PR #17, opened 2020-06-28
Add bonding curve contract

### LINKED ISSUE #5: Build bonding curve with cw20-base
Has a function that takes native tokens to cw20 token at a variable exchange rate. Import basic cw20 functionality from base, just implement the bonding curve logic

ethanfrey:
This is obsolete. Better to throw away and start again. Only leaving this here as the README is likely worth saving

Closes #5 

So far, this is just defining all the proper math so we can evaluate if the implementation is correct.


ethanfrey:
Abandoned idea.

Now being done as #185

PR #24, opened 2020-07-24
Implement subkeys as a cw1 contract

ethanfrey:
This idea was thrown around Summer 2019.

Basically an account that can authorize arbitrary external accounts to spend some limited amount of tokens from it. And in theory perform other actions (like staking).

This implements subkey controls over the native `Bank::SendMsg` action as part of a cw1-proxy contract.

### DIFF
--- a/contracts/cw1-subkeys/src/balance.rs
+++ b/contracts/cw1-subkeys/src/balance.rs
@@ -222,9 +222,9 @@ mod test {
         let no_btc = (balance.clone() - coin(555, "BTC")).unwrap();
         assert_eq!(no_btc, Balance(vec![coin(12345, "ETH")]));
 
-        // subtract more than we have
+        // subtract more than we have is tolerated
         let underflow = balance.clone() - coin(666, "BTC");
-        assert!(underflow.is_err());
+        assert!(underflow.is_ok());
 
         // subtract non-existent denom
         let missing = balance.clone() - coin(1, "ATOM");
--- a/contracts/cw1-subkeys/src/balance.rs
+++ b/contracts/cw1-subkeys/src/balance.rs
@@ -110,17 +110,29 @@ impl ops::Add<Balance> for Balance {
     }
 }
 
+impl ops::SubAssign<Coin> for Balance {
+    fn sub_assign(&mut self, other: Coin) {
+        // Warning: fails silently if no tokens
+        if let Some((i, c)) = self.find(&other.denom) {
+            if c.amount <= other.amount {
+                self.0.remove(i);
+            } else {
+                self.0[i].amount = (c.amount - other.amount).unwrap();
+            }
+        }
+    }
+}
+
 impl ops::Sub<Coin> for Balance {
     type Output = StdResult<Self>;
 
     fn sub(mut self, other: Coin) -> StdResult<Self> {
         match self.find(&other.denom) {
             Some((i, c)) => {
-                let remainder = (c.amount - other.amount)?;
-                if remainder.u128() == 0 {
+                if c.amount <= other.amount {
                     self.0.remove(i);
                 } else {
-                    self.0[i].amount = remainder;
+                    self.0[i].amount = (c.amount - other.amount).unwrap();
                 }
             }
             // error if no tokens
--- a/contracts/cw1-subkeys/src/balance.rs
+++ b/contracts/cw1-subkeys/src/balance.rs
@@ -129,8 +129,7 @@ impl ops::Sub<Coin> for Balance {
 #[cfg(test)]
 mod test {
     use super::*;
-    use cosmwasm_std::{from_slice, to_vec};
-    use std::convert::TryInto;
+    use cosmwasm_std::coin;
 
     #[test]
     fn balance_has_works() {
--- a/contracts/cw1-subkeys/src/balance.rs
+++ b/contracts/cw1-subkeys/src/balance.rs
@@ -67,6 +67,22 @@ impl Balance {
     pub fn is_empty(&self) -> bool {
         self.0.is_empty()
     }
+
+    /// similar to `Balance.sub`, but doesn't fail when minuend less than subtrahend
+    pub fn sub_saturating(mut self, other: Coin) -> StdResult<Self> {
+        match self.find(&other.denom) {
+            Some((i, c)) => {
+                if c.amount <= other.amount {
+                    self.0.remove(i);
+                } else {
+                    self.0[i].amount = (self.0[i].amount - other.amount)?;
+                }
+            }
+            // error if no tokens
+            None => return Err(StdError::underflow(0, other.amount.u128())),
+        };
+        Ok(self)
+    }
 }
 
 impl ops::AddAssign<Coin> for Balance {
@@ -213,6 +229,31 @@ mod test {
         // subtract more than we have
         let underflow = balance.clone() - coin(666, "BTC");
         assert!(underflow.is_err());
+
+        // subtract non-existent denom
+        let missing = balance.clone() - coin(1, "ATOM");
+        assert!(missing.is_err());
+    }
+
+    #[test]
+    fn balance_subtract_saturating_works() {
+        let balance = Balance(vec![coin(555, "BTC"), coin(12345, "ETH")]);
+
+        // subtract less than we have
+        let less_eth = balance.clone().sub_saturating(coin(2345, "ETH")).unwrap();
+        assert_eq!(
+            less_eth,
+            Balance(vec![coin(555, "BTC"), coin(10000, "ETH")])
+        );
+
+        // subtract all of one coin (and remove with 0 amount)
+        let no_btc = balance.clone().sub_saturating(coin(555, "BTC")).unwrap();
+        assert_eq!(no_btc, Balance(vec![coin(12345, "ETH")]));
+
+        // subtract more than we have
+        let saturating = balance.clone().sub_saturating(coin(666, "BTC"));
+        assert!(saturating.is_ok());
+        assert_eq!(saturating.unwrap(), Balance(vec![coin(12345, "ETH")]));
 
         // subtract non-existent denom
         let missing = balance.clone() - coin(1, "ATOM");
--- a/contracts/cw1-subkeys/src/balance.rs
+++ b/contracts/cw1-subkeys/src/balance.rs
@@ -1,16 +1,15 @@
 use schemars::JsonSchema;
-use serde::{de, ser, Deserialize, Deserializer, Serialize};
-use std::convert::TryFrom;
-use std::{fmt, ops};
+use serde::{Deserialize, Serialize};
+use std::ops;
 
-use cosmwasm_std::{underflow, StdError, Coin};
+use cosmwasm_std::{Coin, StdError, StdResult};
 
-// Wallet wraps Vec<Coin> and provides some nice helpers. It mutates the Vec and can be
+// Balance wraps Vec<Coin> and provides some nice helpers. It mutates the Vec and can be
 // unwrapped when done.
 #[derive(Serialize, Deserialize, Clone, Default, Debug, PartialEq, JsonSchema)]
-pub struct Wallet(pub Vec<Coin>);
+pub struct Balance(pub Vec<Coin>);
 
-impl Wallet {
+impl Balance {
     pub fn into_vec(self) -> Vec<Coin> {
         self.0
     }
@@ -66,7 +65,7 @@ impl Wallet {
     }
 }
 
-impl ops::AddAssign<Coin> for Wallet {
+impl ops::AddAssign<Coin> for Balance {
     fn add_assign(&mut self, other: Coin) {
         match self.find(&other.denom) {
             Some((i, c)) => {
@@ -81,7 +80,7 @@ impl ops::AddAssign<Coin> for Wallet {
     }
 }
 
-impl ops::Add<Coin> for Wallet {
+impl ops::Add<Coin> for Balance {
     type Output = Self;
 
     fn add(mut self, other: Coin) -> Self {
@@ -90,24 +89,24 @@ impl ops::Add<Coin> for Wallet {
     }
 }
 
-impl ops::AddAssign<Wallet> for Wallet {
-    fn add_assign(&mut self, other: Wallet) {
+impl ops::AddAssign<Balance> for Balance {
+    fn add_assign(&mut self, other: Balance) {
         for coin in other.0.into_iter() {
             self.add_assign(coin);
         }
     }
 }
 
-impl ops::Add<Wallet> for Wallet {
+impl ops::Add<Balance> for Balance {
     type Output = Self;
 
-    fn add(mut self, other: Wallet) -> Self {
+    fn add(mut self, other: Balance) -> Self {
         self += other;
         self
     }
 }
 
-impl ops::Sub<Coin> for Wallet {
+impl ops::Sub<Coin> for Balance {
     type Output = StdResult<Self>;
 
     fn sub(mut self, other: Coin) -> StdResult<Self> {
@@ -121,7 +120,7 @@ impl ops::Sub<Coin> for Wallet {
                 }
             }
             // error if no tokens
-            None => return StdError::underflow(0, other.amount.u128()),
+            None => return Err(StdError::underflow(0, other.amount.u128())),
         };
         Ok(self)
     }
@@ -134,33 +133,36 @@ mod test {
     use std::convert::TryInto;
 
     #[test]
-    fn wallet_has_works() {
-        let wallet = Wallet(vec![coin(555, "BTC"), coin(12345, "ETH")]);
+    fn balance_has_works() {
+        let balance = Balance(vec![coin(555, "BTC"), coin(12345, "ETH")]);
 
         // less than same type
-        assert!(wallet.has(&coin(777, "ETH")));
+        assert!(balance.has(&coin(777, "ETH")));
         // equal to same type
-        assert!(wallet.has(&coin(555, "BTC")));
+        assert!(balance.has(&coin(555, "BTC")));
 
         // too high
-        assert!(!wallet.has(&coin(12346, "ETH")));
+        assert!(!balance.has(&coin(12346, "ETH")));
         // wrong type
-        assert!(!wallet.has(&coin(456, "ETC")));
+        assert!(!balance.has(&coin(456, "ETC")));
     }
 
     #[test]
-    fn wallet_add_works() {
-        let wallet = Wallet(vec![coin(555, "BTC"), coin(12345, "ETH")]);
+    fn balance_add_works() {
+        let balance = Balance(vec![coin(555, "BTC"), coin(12345, "ETH")]);
 
         // add an existing coin
-        let more_eth = wallet.clone() + coin(54321, "ETH");
-        assert_eq!(more_eth, Wallet(vec![coin(555, "BTC"), coin(66666, "ETH")]));
+        let more_eth = balance.clone() + coin(54321, "ETH");
+        assert_eq!(
+            more_eth,
+            Balance(vec![coin(555, "BTC"), coin(66666, "ETH")])
+        );
 
         // add an new coin
-        let add_atom = wallet.clone() + coin(777, "ATOM");
+        let add_atom = balance.clone() + coin(777, "ATOM");
         assert_eq!(
             add_atom,
-            Wallet(vec![
+            Balance(vec![
                 coin(777, "ATOM"),
                 coin(555, "BTC"),
                 coin(12345, "ETH"),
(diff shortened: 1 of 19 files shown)


ethanfrey:
@maurolacy I sketched out my ideas on some sort of reuse of code from cw1_whitelist to add the idea of subkeys (which is allowances for native tokens, via `BankMsg::Send`). Hopefully the code comments and the README make sense. 

Can you make a PR off of this to complete the work? I left some `panics` in the contract code that needs implementing, and comments for the test cases that would be good to have. It should be rather straight-forward (if a non-trivial amount of code).

Once you make one PR implementing it, if you have better ideas on how to promote code reuse (a cleaner way that I did), please make a second PR building on it that makes those changes - feel free to reorganize code in other contracts and packages when doing this)

maurolacy:
> @maurolacy I sketched out my ideas on some sort of reuse of code from cw1_whitelist to add the idea of subkeys (which is allowances for native tokens, via `BankMsg::Send`). Hopefully the code comments and the README make sense.
> 
> Can you make a PR off of this to complete the work? I left some `panics` in the contract code that needs implementing, and comments for the test cases that would be good to have. It should be rather straight-forward (if a non-trivial amount of code).
> 
> Once you make one PR implementing it, if you have better ideas on how to promote code reuse (a cleaner way that I did), please make a second PR building on it that makes those changes - feel free to reorganize code in other contracts and packages when doing this)

OK, I'll take a look and ask you if I have any doubts.

contracts/cw1-subkeys/src/balance.rs
```diff
@@ -106,17 +110,29 @@ impl ops::Add<Balance> for Balance {
     }
 }
 
+impl ops::SubAssign<Coin> for Balance {
```
ethanfrey, line 113: Yeah, this has to panic (no result possible), so I didn't implement. Panics inside contracts (when run in vm) have no useful message. And we should give a sensible error message for "not enough funds".

I would just use the sub method and maybe add docs why

contracts/cw1-subkeys/src/balance.rs
```diff
@@ -106,17 +110,29 @@ impl ops::Add<Balance> for Balance {
...
     fn sub(mut self, other: Coin) -> StdResult<Self> {
         match self.find(&other.denom) {
             Some((i, c)) => {
-                let remainder = (c.amount - other.amount)?;
-                if remainder.u128() == 0 {
+                if c.amount <= other.amount {
                     self.0.remove(i);
                 } else {
-                    self.0[i].amount = remainder;
+                    self.0[i].amount = (c.amount - other.amount).unwrap();
```
ethanfrey, line 135: Again, please avoid unwrap, makes deployed contracts near impossible to debug

contracts/cw1-subkeys/src/balance.rs
```diff
@@ -206,9 +222,9 @@ mod test {
         let no_btc = (balance.clone() - coin(555, "BTC")).unwrap();
         assert_eq!(no_btc, Balance(vec![coin(12345, "ETH")]));
 
-        // subtract more than we have
+        // subtract more than we have is tolerated
```
ethanfrey, line 225: This is most often what we do not want. Look how Uint128 handles underflow, that is the model we do want

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -82,14 +83,38 @@ pub fn handle_increase_allowance<S: Storage, A: Api, Q: Querier, T>(
...
+    let spender_raw = &deps.api.canonical_address(&spender)?;
+    let owner_raw = &deps.api.canonical_address(&env.message.sender)?;
+
+    if !cfg.is_admin(&owner_raw) {
+        return Err(StdError::unauthorized());
+    }
+    if spender_raw == owner_raw {
+        return Err(StdError::generic_err("Cannot set allowance to own account"));
+    }
+
```
ethanfrey, line 96: Good logic here

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -82,14 +83,38 @@ pub fn handle_increase_allowance<S: Storage, A: Api, Q: Querier, T>(
...
+            allowance.expires = exp;
+        }
+        allowance.balance.add_assign(amount.clone());
+        Ok(allowance)
+    })?;
+
+    let res = HandleResponse {
+        messages: vec![],
+        log: vec![
+            log("action", "increase_allowance"),
```
ethanfrey, line 109: And nice logs

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -102,18 +127,47 @@ pub fn handle_decrease_allowance<S: Storage, A: Api, Q: Querier, T>(
...
+    if spender_raw == owner_raw {
+        return Err(StdError::generic_err("Cannot set allowance to own account"));
+    }
+
+    let allowance = allowances(&mut deps.storage).update(spender_raw.as_slice(), |allow| {
+        if allow.is_none() {
+            // Fail fast
+            return Err(StdError::generic_err("No allowance for this account"));
+        }
+        let mut allowance = allow.unwrap();
```
ethanfrey, line 146: Logic is correct but I try to avoid unwrap at all in the code to make it clear there is no panic (yes, I know you just checked). You can use Option.ok_or_else(..generic err)? To do the same thing without unwrap (yes i know just syntax but I do grep for unwrap)

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -102,18 +127,47 @@ pub fn handle_decrease_allowance<S: Storage, A: Api, Q: Querier, T>(
...
+
+        if let Some(exp) = expires {
+            allowance.expires = exp;
+        }
+
+        allowance.balance = allowance.balance.sub(amount.clone())?; // Fails if no tokens
```
ethanfrey, line 152: This must error if amount is bigger than balance. And I think y you disabled that check above. (Allow underflow comment)

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -102,18 +127,47 @@ pub fn handle_decrease_allowance<S: Storage, A: Api, Q: Querier, T>(
...
+        Ok(allowance)
+    })?;
+    if allowance.balance.is_empty() {
+        allowances(&mut deps.storage).remove(spender_raw.as_slice());
```
ethanfrey, line 156: Nice way to handle this without a second read.

You could also avoid update and just load, modify, and decide to save or remove.

contracts/cw1-subkeys/src/state.rs
```diff
@@ -9,19 +9,19 @@ use crate::balance::Balance;
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema, Default)]
 pub struct Allowance {
-    balance: Balance,
-    expires: Expiration,
+    pub balance: Balance,
+    pub expires: Expiration,
 }
```
ethanfrey, line 14: Good call, these should all be public members

contracts/cw1-subkeys/src/balance.rs
maurolacy: I see. In fact I pushed this commit just to have a discussion about it. I'll revert it.

contracts/cw1-subkeys/src/balance.rs
maurolacy: I decided to go ahead with these commits because of your comment in the original TODO:
_underflow is okay, removes that denom
eg. if allowance was [5 ETH, 2 BTC] and I decrease by 4 BTC, new allowance is [5 ETH]._

If you're okay with failing in case of underflow, I'm OK too. In fact I think it's a safer option, because underflowing could introduce accounting issues on the client side. Plus the logs will have wrong or incomplete information.

contracts/cw1-subkeys/src/contract.rs
maurolacy: Thanks. Copied it from erc20-base, I think.

contracts/cw1-subkeys/src/contract.rs
maurolacy: OK, will change it.

contracts/cw1-subkeys/src/contract.rs
maurolacy: I'll revert back those, and we'll be safe.

contracts/cw1-subkeys/src/state.rs
maurolacy: I did it just for simplicity (avoiding getters / setters). I could implement those if you want, but I think this is clear, and good enough.

contracts/cw1-subkeys/src/contract.rs
maurolacy: Yes. I'll change it if you want, but I like update: it prevents you from forgetting to save. I also like its functional style, sort of.

contracts/cw1-subkeys/src/contract.rs
maurolacy: I knew it had to exist, but wasn't aware of `ok_or_else()`, thanks.

contracts/cw1-subkeys/src/balance.rs
ethanfrey: On decrease allowance, underfloor is okay (treat errors a 0). On execute (if we spend the money) the allowance is a hard cap, we cannot go below 0.

You could add a "saturating-sub" method to balance that does the logic ignoring underflow to use in decrease allowance. We have two different use cases for sub so it makes sense for two functions

contracts/cw1-subkeys/src/contract.rs
ethanfrey: It's okay either way. I do like update myself and I think you use it quite well here, especially using the return value. I think you could optimize, but feel this is more legible, so feel free to keep it.

contracts/cw1-subkeys/src/state.rs
ethanfrey: It was a mistake by me. Any serializable object in cosmwasm code has all public members. They are just "dumb" data structures

contracts/cw1-subkeys/src/balance.rs
maurolacy: OK, will do.

contracts/cw1-subkeys/src/state.rs
maurolacy: I like the approach.

ethanfrey:
Please ping me when you implement execute and I will give a final review. The allowances look fine

maurolacy:
OK, I've done a first implementation of `execute`. I have a number of doubts, some of which are reflected in the code.

Please take a look. Once this is right, I'll write the unit tests.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -57,19 +57,46 @@ where
...
+            match msg {
+                CosmosMsg::Bank(BankMsg::Send {
+                    from_address: _,
+                    to_address: _,
+                    amount,
+                }) => {
+                    // FIXME?: Can from_address be different from env.message.sender? (guess so)
+                    // FIXME?: Can from_address and to_address be the same? (i.e. fancy sender)
+                    // Decrease allowance
+                    for coin in amount {
```
ethanfrey, line 82: This loop looks good and is the only check/update we need here

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -145,7 +172,7 @@ where
         if let Some(exp) = expires {
             allowance.expires = exp;
         }
-        allowance.balance = allowance.balance.sub(amount.clone())?; // Fails if no tokens
+        allowance.balance = allowance.balance.sub_saturating(amount.clone())?; // Fails if no tokens
```
ethanfrey, line 175: The comment "fails if no tokens". Is that out of date now?

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -57,19 +57,46 @@ where
...
+        let mut allowance =
+            allow.ok_or_else(|| StdError::generic_err("No allowance for this account"))?;
+        for msg in &msgs {
+            match msg {
+                CosmosMsg::Bank(BankMsg::Send {
+                    from_address: _,
+                    to_address: _,
+                    amount,
+                }) => {
+                    // FIXME?: Can from_address be different from env.message.sender? (guess so)
```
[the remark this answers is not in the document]
maurolacy: I see. In fact that's what I was thinking this morning: that there's no way for these messages to get through, unless the sender address is valid / correct.

So, in short it's the responsibility of the sender to set the correct `from_address`. And that has to be the contract address; which makes total sense.

contracts/cw1-subkeys/src/contract.rs
maurolacy: Right, thanks.

maurolacy:
> Looks good. Just missing tests, but besides the minor cleanup I mentioned, it looks correct.

I'm glad I figured it out. I must say in the name of truth that it was mostly based on logic and deductions, as I still lack detailed knowledge of the platform and the SDK.

contracts/cw1-subkeys/src/contract.rs
maurolacy: Correction: this still fails if no tokens. It silently ignores underflows, but doesn't ignore a missing balance entry for that denomination. I think this is an important distinction: an admin could make a spelling mistake when trying to remove an entry, and if this fails silently, he will never be notified of the error.

Nevertheless, if you think this is not desirable / important, I'll change it to always failing silently.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -57,19 +57,46 @@ where
...
-        // TODO
-        // for each message, check if the message is a BankMsg::Send and the if the subkey has sufficient allowance.
-        // if so, reduce the allowance and resend this message
-        //
-        // Note, you may want to use the cosmwasm_std::Context object to build the response
-        panic!("unimplemented")
+        let mut allowances = allowances(&mut deps.storage);
+        let allow = allowances.may_load(owner_raw.as_slice())?;
+        let mut allowance =
+            allow.ok_or_else(|| StdError::generic_err("No allowance for this account"))?;
```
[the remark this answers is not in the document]
maurolacy: Changed, thanks.

contracts/cw1-subkeys/src/contract.rs
maurolacy: Related: Shouldn't `denom` be case insensitive, for robustness and simplicity?

ethanfrey: Very good distinction to make (underflow vs. no denom) and I agree with it. Nice thinking.

The comment could be a little more verbose and capture that.

ethanfrey: As to case insensitive... it is case sensitive in the sdk. we could normalize both stored values and the ones we match to always be lower case before comparing. I guess that would be nice. But I'm not sure if there are issues there. Can you make an issue to add that, and we can add that later, once this is merged. And then we can maybe get some comments on the idea. I am not completely sure about that one.

ethanfrey:
> I'm glad I figured it out. I must say in the name of truth that it was mostly based on logic and deductions, as I still lack detailed knowledge of the platform and the SDK.

Please add an issue to https://github.com/CosmWasm/docs2 for all info you would like to have readily accessible when first wirting a contract. We are planning/starting an overhaul of the docs site and input from people just starting with CosmWasm is invaluable.

You an make a new issue or add some thoughts here: https://github.com/CosmWasm/docs2/issues/5

contracts/cw1-subkeys/src/contract.rs
maurolacy: Me neither... I'll create the issue.

maurolacy:
> Please add an issue to https://github.com/CosmWasm/docs2 for all info you would like to have readily accessible when first wirting a contract. We are planning/starting an overhaul of the docs site and input from people just starting with CosmWasm is invaluable.
> 
> You an make a new issue or add some thoughts here: [CosmWasm/docs2#5](https://github.com/CosmWasm/docs2/issues/5)

For me, the most "difficult" part in some cases was to figure out what NOT to do (i. e. what was not necessary). To be fair, this is probably related to my way of approaching things, as I'm no fan of reading lot of docs when starting with something new, and prefer to gain a big picture gradually. Starting from the details, so to speak.

Regargind the complexities, by example, the fact that for relaying the messages it's enough to set the `messages` member in the `HandleResponse` struct was far from obvious to me. I got it by analysing the already existing functionality in `execute` for the admin case, and deducting the behaviour of the SDK.

I must say that the examples are very good, and were very useful in general. `cw1-whitelist` and `cw20-base` in particular, of which `cw1-subkeys` is just an elaboration, or iteration. The API itself is well thought of, and makes a lot of sense, which is very helpful.

I'll probably elaborate on this in the issue later, but I would say that some short documents, even diagrams, with the aim to gain a big picture quickly, particularly of the interactions and responsibilities of the different parts and frameworks, would be very useful for future would-be contract writers.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -216,14 +216,38 @@ pub fn query_allowance<S: Storage, A: Api, Q: Querier>(
...
+            admins: admins.clone(),
+            mutable: false,
+        };
+        init(deps, env.clone(), init_msg).unwrap();
 
-    // you probably want some `setup_test_case` function that inits a contract with 2 admins
-    // and 2 subkeys with some allowances. these keys can be constants strings here,
-    // used like `HumanAddr::from(admin1)` and then all tests can just run against that
+        // Add subkeys with initial allowances
+        for (spender, expiration) in spenders.iter().zip(expirations) {
```
ethanfrey, line 240: interesting zip. works well actually.

I have used `&[(HumanAddr, Expiration)]` in similar cases, which made more sense to me, but also may be harder to construct (if you have a list of admins anyway)

ethanfrey: Actually reading https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.zip
It seems that the test case below with 4 spenders and 2 expirations, only the first two spenders will be registered as `zip` ends when the first of the two iterators ends.

That would not be my expected behavior from this function signature, and should be modified or documented with rustdoc at the function level

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -239,11 +263,115 @@ mod tests {
...
+            expires: Some(expires_height.clone()),
+        };
+        handle(&mut deps, env.clone(), msg).unwrap();
+
+        // Verify
+        let allowance = query_allowance(&deps, spender2.clone()).unwrap();
+        assert_eq!(
+            allowance,
+            Allowance {
+                balance: Balance(vec![coin(amount1 * 2, &allow1.denom), allow2.clone()]),
```
ethanfrey, line 334: for the sake of completeness, could you add a different denom here, either allow2, or maybe even allow3?

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -239,11 +263,115 @@ mod tests {
...
+        let allowance = query_allowance(&deps, spender3.clone()).unwrap();
+        assert_eq!(
+            allowance,
+            Allowance {
+                balance: Balance(vec![allow1.clone()]),
+                expires: expires_never.clone()
+            }
+        );
+
+        // Add to spender4 (new account) (expires = Some)
```
ethanfrey, line 357: Nicely covering all the use cases.

It actually seems the `zip` above did exactly what you wanted it to do. 
but `setup_test_case` needs better rustdoc, so this is expected behavior

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -375,11 +375,124 @@ mod tests {
...
+        };
+        handle(&mut deps, env.clone(), msg).unwrap();
+
+        // Verify
+        let allowance = query_allowance(&deps, spender1.clone()).unwrap();
+        assert_eq!(
+            allowance,
+            Allowance {
+                balance: Balance(vec![
+                    coin(amount1 / 2 + (amount1 & 1), denom1),
```
ethanfrey, line 468: I would usually use `+ (amount1 % 2)` rather than `+ (amount 1 & 1)` but they produce the same result.
I guess you are more used to embedded systems than I am.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -375,11 +375,124 @@ mod tests {
...
+                    allow2.clone()
+                ]),
+                expires: expires_height.clone()
+            }
+        );
+
+        // Subtract from spender2 (existing) account (brings denom to 0, no other denoms left => should delete Allowance)
```
ethanfrey, line 475: Very nice coverage of all the test cases

ethanfrey:
Oops, I realized I had some change requests above, after I merged.
Can you make a small PR with any minor adjustments (mainly the rustdoc for setup_test_case), but anything else you feel like changing as well.

maurolacy:
Sure, I was planning to finish those two remaining tests tomorrow morning anyway.

Sorry to commit / push partial stuff. I'm just used to working that way, but I can change it if it's a burden.

contracts/cw1-subkeys/src/contract.rs
maurolacy: > It seems that the test case below with 4 spenders and 2 expirations, only the first two spenders will be registered as `zip` ends when the first of the two iterators ends.

Yes, but I've put only two spenders in the array. Maybe renaming it to `initial_spenders` is a good idea.

contracts/cw1-subkeys/src/contract.rs
maurolacy: Sure, will do it in the continuation PR.

contracts/cw1-subkeys/src/contract.rs
maurolacy: I think it's related to the  comment above. I'll clarify the code with better naming / comments.

ethanfrey:
No, it is great. Please keep pushing.

Two issues.. it is late and I was trigger happy on a merge. And two, I opened this PR, otherwise I would have waited for you to take it out of draft mode before merging.

Generally, open a draft PR when starting something and mark "ready for review" when you feel it is good to merge if approved.

PR #28, opened 2020-08-08
Subkeys 2

maurolacy:
Improved test cases after previous PR comments.

Added the two remaining test cases.

### DIFF
--- a/contracts/cw1-subkeys/src/contract.rs
+++ b/contracts/cw1-subkeys/src/contract.rs
@@ -219,6 +219,7 @@ mod tests {
     use crate::balance::Balance;
     use cosmwasm_std::testing::{mock_dependencies, mock_env};
     use cosmwasm_std::{coin, coins};
+    use cw1_whitelist::msg::ConfigResponse;
 
     // this will set up the init for other tests
     fn setup_test_case<S: Storage, A: Api, Q: Querier>(
@@ -232,7 +233,7 @@ mod tests {
         // Init a contract with admins
         let init_msg = InitMsg {
             admins: admins.clone(),
-            mutable: false,
+            mutable: true,
         };
         init(deps, env.clone(), init_msg).unwrap();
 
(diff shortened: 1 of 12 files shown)


maurolacy:
Still need to test `execute()`. Guess I'll need an integration test for it... will be working on it asap.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -251,14 +252,142 @@ mod tests {
...
+        setup_test_case(
+            &mut deps,
+            &env,
+            &admins,
+            &initial_spenders,
+            &initial_allowances,
+            &initial_expirations,
+        );
+
+        // Check allowances work for accounts with balances
```
ethanfrey, line 285: Nice tests

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -251,14 +252,142 @@ mod tests {
...
+            config,
+            ConfigResponse {
+                admins: initial_admins.clone(),
+                mutable: true,
+            }
+        );
+
+        // Add a third (new) admin
+        let msg = HandleMsg::UpdateAdmins {
+            admins: [&config.admins[..], &[admin3.clone()]].concat(),
```
ethanfrey, line 332: Nice use of the spread and concat operators. But somehow, I would prefer to see this more explicit.

```rust
let new_admins = vec![owner.clone(), admin2.clone(), admin3.clone()]
let msg = HandleMsg::UpdateAdmins { admins: new_admins.clone() };
// ...
assert_eq!(config, ConfigResponse { admins: new_admins, mutable: true });
```

This is the exact same functionality you have written but easier to read. And key in test cases is legibility (performance is not an issue)

ethanfrey:
Going through this code again, not the PR, I notice there is no test for execute, that the allowances work as they are supposed to.

maurolacy:
Yes. I'll write the `execute()` tests tomorrow.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -646,4 +646,72 @@ mod tests {
...
+        assert_eq!(res.log, vec![log("action", "execute"), log("owner", spender1)]);
+
+        // And then cannot (not enough funds anymore)
+        let res = handle(&mut deps, env, handle_msg.clone());
+        match res.unwrap_err() {
+            StdError::Underflow { .. } => {},
+            e => panic!("unexpected error: {}", e),
+        }
+
+        // Owner / admins can do anything (at the contract level)
```
ethanfrey, line 713: Good test case.

Can you add another where the spender (with allowance) tries to send a message besides `BankMsg::Send`. Anything else to show it fails.

Just one more test and this is :+1:

maurolacy: Will do.

PR #29, opened 2020-08-11
Add deployment job to CI

webmaster128:


webmaster128:
Blocked by https://github.com/CosmWasm/cosmwasm-plus/issues/30

ethanfrey:
Also blocked by https://github.com/CosmWasm/rust-optimizer/issues/18

ethanfrey:
I closed #30 and https://github.com/CosmWasm/rust-optimizer/pull/19 is ready for review (to fix the build errors)

.circleci/config.yml
```diff
@@ -1,15 +1,23 @@
 version: 2
 workflows:
   version: 2
-  test:
+  # test:
```
ethanfrey, line 4: Why are these all commented out?
I assume just as WIP while testing the other script, right?

.circleci/config.yml
```diff
@@ -1,15 +1,23 @@
...
-      - contract_cw1_whitelist
-      - contract_cw20_base
-      - contract_cw20_escrow
-      - package_cw1
-      - package_cw20
-      - lint
+      - build_and_upload_contracts
+          # filters:
+          #   tags:
+          #     only: /^v[0-9]+\.[0-9]+\.[0-9]+.*/
```
ethanfrey, line 18: Yeah, and other code to enable once it works. Got it.

.circleci/config.yml
```diff
@@ -292,3 +300,52 @@ jobs:
...
+          command: |
+            # create a dummy container which will hold a volume with config
+            docker create -v /code --name with_code alpine /bin/true
+            # copy a config file into this volume
+            docker cp ./contracts with_code:/code
+            docker cp ./packages with_code:/code
+      - run:
+          name: Build development contracts
+          command: |
+            for contract_dir in ./contracts/*/; do
```
ethanfrey, line 330: Okay, this needs to be updated for 0.10.0 rust-optimizer, but the rest looks quite straight-forward.

Thanks for this PR

ethanfrey:
I updated this to rust-optimizer:0.10.0. Let's see if it works now

ethanfrey:
Note cw1-whitelist is super small, this is due to the fact the "library" feature is activated from the cw1-subkeys import, which affects the normal `cw1_whitelist.wasm` build as well. Since we don't use this contract, I can live with it, but time to revisit the build process a bit

.circleci/config.yml
webmaster128: Yes, in order to test the new job. You can revert or remove the commit "Dactivate test workflow" later

.circleci/config.yml
webmaster128: Yes, to activate the filter, you need to add a `:` after `build_and_upload_contracts` as well

.circleci/config.yml
```diff
@@ -292,3 +300,51 @@ jobs:
...
+            docker cp Cargo.toml with_code:/code
+            docker cp Cargo.lock with_code:/code
+            docker cp ./contracts with_code:/code
+            docker cp ./packages with_code:/code
+      - run:
+          name: Build development contracts
+          command: |
+            docker run --volumes-from with_code cosmwasm/rust-optimizer:0.10.0
+            docker cp with_code:/code/artifacts ./artifacts
+            mv artifacts/contracts.txt artifacts/checksums.txt
```
webmaster128, line 331: If we do that, we better call the file `checksums.txt` everywhere. I used that name in ComsJS but we can switch it to `contracts.txt`. However, I think `checksums.txt` is a bit more user friendly.

ethanfrey: Okay, I will update rust-optimizer to 0.10.1 with that naming. Then I can merge this, right?

webmaster128: Yeah, looks good to me. Just the deactivated publishing is missing but this you will figure out by bruteforceing it

ethanfrey: Addressed by https://github.com/CosmWasm/rust-optimizer/pull/20

webmaster128: Perfect. We'll benefit from this consistency in the long run

ethanfrey:
Oops, I need to add a github token here. I will add it and try re-running the build

PR #34, opened 2020-08-12
Cw2 migrate spec

ethanfrey:
Add a new spec for standardized migration data.

This is just a singleton each contract can store to allow future contracts to more easily migrate from this implementation.

Question: is the flexibility of multiple implementaition/versions good, or should we just store `ContractVersion` rather than the more complex `ContractInfo`?

### DIFF
--- a/packages/cw2/README.md
+++ b/packages/cw2/README.md
@@ -24,12 +24,14 @@ paths.
 All CW2-compliant contracts must store the following data:
 
 * key: `\x00\x0dcontract_info` (length prefixed "contract_info" using Singleton pattern)
-* data: Json-serialized `ContractInfo`
+* data: Json-serialized `ContractVersion`
 
 ```rust
 pub struct ContractVersion {
-    /// contract is the crate name of the implementing contract, eg. `crate:cw20-base`
-    /// we will use other prefixes for other languages, and their standard global namespacing
+    /// contract is a globally unique identifier for the contract.
+    /// it should build off standard namespacing for whichever language it is in,
+    /// and prefix it with the registry we use. 
+    /// For rust we prefix with `crates.io:`, to give us eg. `crates.io:cw20-base`
     pub contract: String,
     /// version is any string that this implementation knows. It may be simple counter "1", "2".
     /// or semantic version on release tags "v0.6.2", or some custom feature flag list.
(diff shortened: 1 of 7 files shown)


packages/cw2/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+All CW2-compliant contracts must store the following data:
+
+* key: `\x00\x0dcontract_info` (length prefixed "contract_info" using Singleton pattern)
+* data: Json-serialized `ContractInfo`
+
+```rust
+pub struct ContractVersion {
+    /// contract is the crate name of the implementing contract, eg. `crate:cw20-base`
+    /// we will use other prefixes for other languages, and their standard global namespacing
```
webmaster128, line 32: If you want uniqueness, I think we need to prefix with the registry instead of the build system type. Then this would become `crates.io:cw20-base`

packages/cw2/README.md
```diff
@@ -0,0 +1,78 @@
...
+    pub contract: String,
+    /// version is any string that this implementation knows. It may be simple counter "1", "2".
+    /// or semantic version on release tags "v0.6.2", or some custom feature flag list.
+    /// the only code that needs to understand the version parsing is code that knows how to
+    /// migrate from the given contract (and is tied to it's implementation somehow)
+    pub version: String,
+}
+
+pub struct ContractInfo {
+    versions: Vec<ContractVersion>,
```
webmaster128, line 42: Can you explain why this is an array?

packages/cw2/README.md
```diff
@@ -0,0 +1,78 @@
...
+    "contract": "crate:cw20-base",
+    "version": "v0.1.0"
+  }]
+}
+```
+
+Note we explicitly allow multiple version for mix-ins. For example,
+`cw1-subkeys` imports the config/admin list bucket from `cw1-whitelist`
+as well as adding some more info to it. We may want to explicitly mention
+the source of the state to be migrated:
```
webmaster128, line 60: What is a mix-in and why does it differ from all the other dependencies a contract has?

packages/cw2/README.md
ethanfrey: Sure, I was overengineering it.

I will make it simply ContractVersion. That was my second thought last night and thanks for the feedback.

PR #39, opened 2020-08-12
Unique singleton names

### LINKED ISSUE #37: Use more unique names for store/queries (not "Config" / "Meta")
We start having contracts importing code from other contracts (building on base code).
When multiple contracts use the same generic prefixes or data-types like "config", "meta", this can get confusing.

Use clearer names, like "AdminList" for `cw1-whitelist` and "TokenInfo" for `cw20-base`.

ethanfrey:
Closes #37 

Rename some generic names "Config" / "Meta" to clearer, contract-specific naming

### DIFF
--- a/contracts/cw1-subkeys/examples/schema.rs
+++ b/contracts/cw1-subkeys/examples/schema.rs
@@ -1,7 +1,7 @@
 use std::env::current_dir;
 use std::fs::create_dir_all;
 
-use cosmwasm_schema::{export_schema, remove_schemas, schema_for};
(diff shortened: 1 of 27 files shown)


ethanfrey:
@Thunnini FYI, a minor naming change in the cw20 spec.

It is no longer "Meta" ("meta"), but "TokenInfo" ("token_info") on the queries to get the symbol, decimals, etc.

I will be releasing 1.0 tomorrow hopefully, this is the last change to the public API.

PR #52, opened 2020-08-23
Atomic swap

maurolacy:
Implementation of atomic swaps, for native tokens.


maurolacy:
Integration tests done. Except for `query()` integ test, that are failing with:

`thread 'test_query' panicked at 'VM error: CommunicationErr { source: InvalidRegion { source: LengthExceedsCapacity { length: 1931622946, capacity: 1919247470, backtrace: Backtrace(()) } } }', ...`

The error is when calling the `query()` VM wrapper. Will investigate more tomorrow, but posting here in case you know something about it.

Thanks.

ethanfrey:
You have to add the iterator feature to both cosmwasm-vm and cosmwasm-storage, but I assume you did.

This seems to show you are trying to send an enormous buffer or one with negative size. Can you link to the exact line that returns this error? I will eyeball it and see if I see anything obvious before digging down. You should never be able to create such an error with the higher-level APIs you use

maurolacy:
> You have to add the iterator feature to both cosmwasm-vm and cosmwasm-storage, but I assume you did.

Yes.
 
> This seems to show you are trying to send an enormous buffer or one with negative size. Can you link to the exact line that returns this error? I will eyeball it and see if I see anything obvious before digging down.

There's now PoC code in the [atomic-swap-integ_test_query](https://github.com/CosmWasm/cosmwasm-plus/tree/atomic-swap-integ_test_query) branch.

The [last line of integ_test_query()]((https://github.com/CosmWasm/cosmwasm-plus/blob/95ba038f4403c64c2a30bb81e11ca0aa4e31f541/contracts/atomic-swap/tests/integration.rs#L351) produces the error.

ethanfrey:
Interesting. Two points:

I don't see the equivalent test here: https://github.com/CosmWasm/cosmwasm-plus/blob/atomic-swap-integ_test_query/contracts/atomic-swap/src/contract.rs rather just a details query. Can you make a unit test on the ListQuery to see what it says? It will give you more detail info on any error.

Looking at this https://github.com/CosmWasm/cosmwasm-examples/blob/master/erc20/tests/integration.rs#L378 it seems you do the same thing (unless you want to just `unwrap` it there, but that is not a functional change).

ethanfrey:
Also, I have been removing integration tests in this repo, as I hadn't found a bug with them before (besides the cosmwasm contracts that were there to debug it). I would like to hunt down what happens here, but if we figure out why this is broken, I would generally, just focus on the unit tests and maybe remove all these.

Also, good to add your contract to `.circleci/config` so this is covered by CI (maybe after removing the integration tests that slow things down a lot)

maurolacy:
> Interesting. Two points:
> 
> I don't see the equivalent test here: https://github.com/CosmWasm/cosmwasm-plus/blob/atomic-swap-integ_test_query/contracts/atomic-swap/src/contract.rs rather just a details query. Can you make a unit test on the ListQuery to see what it says? It will give you more detail info on any error.

I converted it to a minimal test, just to show / pinpoint the issue. I have the full integ test for query, and I can commit it if you like, for example in the original branch (it will break the build, though). Anyway, the unit test for query() has a `QueryMsg::List`, and it works fine. What also works, as I've checked it manually, is `QueryMsg::Details`. It's only `List` that's broken in the VM.
 
> Looking at this https://github.com/CosmWasm/cosmwasm-examples/blob/master/erc20/tests/integration.rs#L378 it seems you do the same thing (unless you want to just `unwrap` it there, but that is not a functional change).

If I add the `unwrap()`it's the same; this is just the minimal way to reproduce the problem. In fact, I've checked and the `query()` implementation (the one in `contract.rs`) is never called... there seem to be an issue with arguments parsing in the VM, as far as I can tell at this point.

maurolacy:
> Also, I have been removing integration tests in this repo, as I hadn't found a bug with them before (besides the cosmwasm contracts that were there to debug it). I would like to hunt down what happens here, but if we figure out why this is broken, I would generally, just focus on the unit tests and maybe remove all these.

OK.

> Also, good to add your contract to `.circleci/config` so this is covered by CI (maybe after removing the integration tests that slow things down a lot)

Will do.

ethanfrey:
Huh, I ran this code and played with unit tests and get the same issue.

The `all_swap_ids` code seems proper as well. Maybe we have some vm error iterating over an empty range. I should add a test for that to the cosmwasm.

Update: I did update the test to List a non-empty bucket and the same issue.

ethanfrey:
@maurolacy 

I am off for a bit, but if you want to dig deeper, can you try to add an equivalent test to https://github.com/CosmWasm/cosmwasm/blob/master/contracts/queue/tests/integration.rs (which uses the iterator a bunch). You may just want to add a new query that works like you current List one here: https://github.com/CosmWasm/cosmwasm/blob/master/contracts/queue/src/contract.rs#L115-L124 and then test it with unit test and integration test.

If you get that to fail, then we are much closer to a fix - iterating in one repo. You can also import cosmwasm-storage for the same prefix store approach if it doesn't fail without it.

maurolacy:
Will do.

maurolacy:
OK, I've created unit and integration tests under the queue contract in cosmwasm, and the integration test fails.

The error is similar, but not exactly the same:
```
thread 'query_list' panicked at 'VM error: CommunicationErr { source: RegionLengthTooBig { length: 1663187490, max_length: 65536, backtrace: Backtrace(()) } }', ...
```
(Numbers are always the same).

I cannot push the branch as I don't have permission.

I'll create an issue to track this, and document stuff there.

maurolacy:
Created issue: https://github.com/CosmWasm/cosmwasm/issues/508

maurolacy:
Pushing the `query()` integration test just for reference / completeness.

Will remove integrations tests later as agreed.

ethanfrey:
Thanks, I will review tomorrow... been super busy the last days - sorry.

Can you rebase on master since I did merge a lot recently?

.circleci/config.yml
```diff
@@ -23,6 +24,47 @@ workflows:
              ignore: /.*/
 
 jobs:
+  contract_atomic_swap:
+    docker:
+      - image: rust:1.44.1
+    working_directory: ~/project/contracts/atomic-swap
```
ethanfrey, line 30: please rename dir to cw20-atomic-swap (both here and the actual directory)

It will be the same contract that supports both native and cw20 (just like cw20-escrow), and nice to keep the link there

Cargo.toml
```diff
@@ -1,6 +1,13 @@
 [workspace]
 members = ["packages/*", "contracts/*"]
 
+[profile.release.package.atomic-swap]
```
ethanfrey, line 4: again, `cw20-atomic-swap` good to catch this one

contracts/atomic-swap/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "atomic-swap"
```
ethanfrey, line 2: ditto: `cw20-atomic-swap`

contracts/atomic-swap/examples/schema.rs
```diff
@@ -0,0 +1,21 @@
...
+use atomic_swap::msg::InitMsg;
+
+fn main() {
+    let mut out_dir = current_dir().unwrap();
+    out_dir.push("schema");
+    create_dir_all(&out_dir).unwrap();
+    remove_schemas(&out_dir).unwrap();
+
+    export_schema(&schema_for!(InitMsg), &out_dir);
+    export_schema_with_title(&mut schema_for!(HandleMsg), &out_dir, "HandleMsg");
```
ethanfrey, line 16: This doesn't need to include variants (that are pulled in from `HandleMsg`, but should also include return values, which are not. Please add:

* `QueryMsg`
* `ListResponse`
* `DetailsResponse`

contracts/atomic-swap/src/msg.rs
```diff
@@ -0,0 +1,84 @@
...
+    if bytes.len() < 3 || bytes.len() > 20 {
+        return false;
+    }
+    true
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum QueryMsg {
+    /// Show all open swaps. Return type is ListResponse.
```
ethanfrey, line 52: This obviously will need pagination. Look at https://github.com/CosmWasm/cosmwasm-plus/pull/64 where I add some iterators to cw20-base. Keep point is the pagination.

Pagination input (2 fields on query): https://github.com/CosmWasm/cosmwasm-plus/blob/master/contracts/cw20-base/src/msg.rs#L149-L150

Logic to apply the start/limit: https://github.com/CosmWasm/cosmwasm-plus/blob/master/contracts/cw20-base/src/enumerable.rs#L45-L51

Note that you can use `Option<String>` for the `start_after` as that is the typical id. Otherwise the same. (I need to write up this pattern in cw0... just testing it out with various contracts before moving it into a standard.

contracts/atomic-swap/src/state.rs
```diff
@@ -0,0 +1,92 @@
...
+
+use cosmwasm_std::{
+    CanonicalAddr, Coin, Env, Order, ReadonlyStorage, StdError, StdResult, Storage,
+};
+use cosmwasm_storage::{bucket, bucket_read, prefixed_read, Bucket, ReadonlyBucket};
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug, Default)]
+pub struct AtomicSwap {
+    /// This is hex-encoded sha-256 hash of the preimage (must be 32*2 = 64 chars)
+    pub hash: String,
```
ethanfrey, line 12: Question - why not just use `Binary` here?

For the public API using hex-encoded string makes sense, but for the internal storage, we can store the hash in base64, which is auto-decoded to `Vec<u8>`

contracts/atomic-swap/src/state.rs
```diff
@@ -0,0 +1,92 @@
...
+    pub recipient: CanonicalAddr,
+    pub source: CanonicalAddr,
+    pub end_height: u64,
+    pub end_time: u64,
+    /// Balance in native tokens
+    pub balance: Vec<Coin>,
+}
+
+impl AtomicSwap {
```
ethanfrey, line 21: I would use [`cw0::Expiration`](https://github.com/CosmWasm/cosmwasm-plus/blob/master/packages/cw0/src/expiration.rs#L6-L15) instead of `end_height` and `end_time`. We have the logic elsewhere and it standardized this input field

contracts/atomic-swap/src/state.rs
```diff
@@ -0,0 +1,92 @@
...
+            .save("lazy".as_bytes(), &dummy_swap())
+            .unwrap();
+        atomic_swaps(&mut storage)
+            .save("assign".as_bytes(), &dummy_swap())
+            .unwrap();
+        atomic_swaps(&mut storage)
+            .save("zen".as_bytes(), &dummy_swap())
+            .unwrap();
+
+        let ids = all_swap_ids(&storage).unwrap();
```
ethanfrey, line 85: :+1:

contracts/atomic-swap/src/msg.rs
```diff
@@ -0,0 +1,84 @@
...
+pub struct ListResponse {
+    /// List all open swap ids
+    pub swaps: Vec<String>,
+}
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
+pub struct DetailsResponse {
+    /// Id of this swap
+    pub id: String,
+    /// This is hex-encoded sha-256 hash of the preimage (must be 32*2 = 64 chars)
```
ethanfrey, line 69: Yes, here, very nice to have it hex

contracts/atomic-swap/src/msg.rs
```diff
@@ -0,0 +1,84 @@
...
+    pub hash: String,
+    /// If released, funds go to the recipient
+    pub recipient: HumanAddr,
+    /// If refunded, funds go to the source
+    pub source: HumanAddr,
+    /// When end height greater than zero and block height exceeds this value, the swap is expired.
+    /// Once a swap is expired, it can be returned to the original source (via "refund").
+    pub end_height: u64,
```
ethanfrey, line 77: Again, if we use `cw0::Expiration` in the internals, we can use it in this API as well

contracts/atomic-swap/src/contract.rs
```diff
@@ -0,0 +1,619 @@
...
+use cw2::{set_contract_version, ContractVersion};
+
+use crate::msg::{
+    is_valid_name, CreateMsg, DetailsResponse, HandleMsg, InitMsg, ListResponse, QueryMsg,
+};
+use crate::state::{all_swap_ids, atomic_swaps, atomic_swaps_read, AtomicSwap};
+
+// Version info, for migration info
+const CONTRACT_NAME: &str = "crates.io:atomic-swap";
+const CONTRACT_VERSION: &str = "v0.1.0";
```
ethanfrey, line 16: I realized this is prone to not be updated (I left it 0.1.0 on many contracts for the 0.1.1 release).

Please use: `env!("CARGO_PKG_VERSION");` (see: https://github.com/CosmWasm/cosmwasm-plus/blob/master/contracts/cw1-whitelist/src/contract.rs#L16)

contracts/atomic-swap/src/contract.rs
```diff
@@ -0,0 +1,619 @@
...
+    }
+
+    if env.message.sent_funds.is_empty() {
+        return Err(StdError::generic_err(
+            "Send some coins to create an atomic swap",
+        ));
+    }
+
+    // Ensure this is 32 bytes hex-encoded
+    let _ = parse_hex_32(&msg.hash)?;
```
ethanfrey, line 60: You can use the result and store it as binary (`hash: Binary::from(vec_u8)` or `hash: vec_u8.into()`)

contracts/atomic-swap/src/contract.rs
```diff
@@ -0,0 +1,619 @@
...
+    let swap = AtomicSwap {
+        hash: msg.hash.clone(),
+        recipient: recipient_raw,
+        source: deps.api.canonical_address(&env.message.sender)?,
+        end_height: msg.end_height,
+        end_time: msg.end_time,
+        balance: env.message.sent_funds.clone(),
+    };
+
+    if swap.is_expired(&env) {
```
ethanfrey, line 73: If we used expired object, we can do this check before creating the swap

contracts/atomic-swap/src/contract.rs
```diff
@@ -0,0 +1,619 @@
...
+    if !swap.is_expired(&env) {
+        return Err(StdError::generic_err("Atomic swap not yet expired"));
+    }
+
+    let rcpt = deps.api.human_address(&swap.source)?;
+
+    // We delete the swap
+    atomic_swaps(&mut deps.storage).remove(id.as_bytes());
+
+    let msgs = send_native_tokens(&env.contract.address, &rcpt, env.message.sent_funds);
```
ethanfrey, line 145: Nope. We do not send back the user the coins they sent in this transaction. We need to return the values stored in the swap.

`env.message.sent_funds` => `swap.balance` (just like release above)

contracts/atomic-swap/src/contract.rs
```diff
@@ -0,0 +1,619 @@
...
+        end_height: swap.end_height,
+        end_time: swap.end_time,
+        balance: swap.balance,
+    };
+    Ok(details)
+}
+
+fn query_list<S: Storage, A: Api, Q: Querier>(deps: &Extern<S, A, Q>) -> StdResult<ListResponse> {
+    Ok(ListResponse {
+        swaps: all_swap_ids(&deps.storage)?,
```
ethanfrey, line 212: We will want pagination on this. I like the idea to keep the logic in the `all_swap_ids` function in state, just add pagination there as well. (Maybe you calculate `start_after` and `limit` here from the user input and pass in the args that need to be applied to `range`)

contracts/atomic-swap/src/contract.rs
```diff
@@ -0,0 +1,619 @@
...
+            end_time: 0,
+            end_height: 123456,
+        };
+        handle(&mut deps, env.clone(), HandleMsg::Create(create.clone())).unwrap();
+
+        // Cannot refund, wrong id
+        let refund = HandleMsg::Refund {
+            id: "swap0002".to_string(),
+        };
+        let res = handle(&mut deps, env.clone(), refund);
```
ethanfrey, line 499: We should use a different env here:

`let env = mock_env(HumanAddr::from("anyone"), &[])`

showing that no permissions are needed, and *no balance*, which would have caught the logic bug above

contracts/atomic-swap/src/contract.rs
```diff
@@ -0,0 +1,619 @@
...
+            Ok(_) => panic!("expected error"),
+            Err(StdError::NotFound { .. }) => {}
+            Err(e) => panic!("unexpected error: {:?}", e),
+        }
+
+        // Cannot refund, not expired yet
+        let refund = HandleMsg::Refund {
+            id: "swap0001".to_string(),
+        };
+        let res = handle(&mut deps, env.clone(), refund);
```
ethanfrey, line 510: again, this should be "anyone" with no sent_funds

contracts/atomic-swap/src/contract.rs
```diff
@@ -0,0 +1,619 @@
...
+        match res {
+            Ok(_) => panic!("expected error"),
+            Err(StdError::GenericErr { msg, .. }) => {
+                assert_eq!(msg, "Atomic swap not yet expired".to_string())
+            }
+            Err(e) => panic!("unexpected error: {:?}", e),
+        }
+
+        // Can refund, already expired
+        let env = mock_env_height("anybody", &balance, 123457);
```
ethanfrey, line 520: Once again, you do not send `balance` (all the fund) in the `refund`

contracts/atomic-swap/src/contract.rs
```diff
@@ -0,0 +1,619 @@
...
+            end_height: 123456,
+        };
+        handle(&mut deps, env.clone(), HandleMsg::Create(create.clone())).unwrap();
+
+        // Cannot release, wrong id
+        let release = HandleMsg::Release {
+            id: "swap0002".to_string(),
+            preimage: preimage(),
+        };
+        let res = handle(&mut deps, env.clone(), release);
```
ethanfrey, line 398: env for release should have no tokens. Use a different `mock_env` like `mock_env(HumanAddr::from("anyone"), &[])`

contracts/atomic-swap/src/contract.rs
```diff
@@ -0,0 +1,619 @@
...
+        // Cannot refund again
+        let res = handle(&mut deps, env.clone(), refund);
+        match res.unwrap_err() {
+            StdError::NotFound { .. } => {}
+            e => panic!("Expected NotFound, got {}", e),
+        }
+    }
+
+    #[test]
+    fn test_query() {
```
ethanfrey, line 545: No list query tests anymore? It is working with 0.10.1 cosmwasm

ethanfrey: Please add some list queries and some tests with pagination. Reference: https://github.com/CosmWasm/cosmwasm-plus/blob/master/contracts/cw20-base/src/enumerable.rs#L201-L215

ethanfrey (commented):
Thanks for building this out. I have also made some enhancements to the patterns in cosmwasm-plus the last week: pagination, new Expiration type, version calculation, etc. And make a lot of references to update to those newer specs. I think I only see one bug, and a few missing tests (and a few types that can be adjusted).

Please make the requested updates and this should be good to merge.

contracts/atomic-swap/src/contract.rs
maurolacy: Oops. Good catch!

contracts/atomic-swap/src/contract.rs
maurolacy: Yeah, sorry about these. Was based on the original version of the swaps, that had the funds in the contract, and I forgot to change it.

maurolacy: No, this was a copy / paste error. Good catch.

contracts/atomic-swap/src/contract.rs
maurolacy: > No list query tests anymore? It is working with 0.10.1 cosmwasm

I would leave the query unit test, and remove (all of) the integration test(s).

contracts/atomic-swap/src/contract.rs
maurolacy: I don't know what I was thinking. And, it's unfortunate that the broken sender balance in unit tests didn't flag this.

contracts/atomic-swap/src/msg.rs
maurolacy: I was thinking about this and a good idea would be to encapsulate the pagination functionality inside `range()` directly, or by using an additional function like `interval()`. And with a couple extra options like `Inclusive` (default), `Exclusive`, `OpenBelow`, `OpenAbove`. Limits can be handled the same way: with an option like `Option<usize>`, or with naming (`range()` vs. `range_all()`, etc.)

One of the few things I don't like about Rust is lack of default parameters, by the way (other being the excessive need for `clone()`).

maurolacy: Also, I don't like the approach in `calc_range_start()`. Maybe it's better to just get an inclusive range, and then filter out the first / last elements, in case of open intervals / exclusive ranges.

The logic can be a bit complex or cumbersome, but it needs to be written just once, and it can be thoroughly tested. 

It would probably be easier to write a general function using generic types this way, too.

ethanfrey (approved):
Looks good. A few minor nitpics, I will merge this and you can address those in your next PR if you wish.

Also note #69 which you can gladly take on if you wish.

contracts/cw20-atomic-swap/src/msg.rs
```diff
@@ -0,0 +1,81 @@
+use cosmwasm_std::{Coin, HumanAddr};
+use cw20::Expiration;
```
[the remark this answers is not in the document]
maurolacy: OK. Will do it in the next PR.

contracts/cw20-atomic-swap/src/state.rs
```diff
@@ -0,0 +1,96 @@
...
+use cosmwasm_std::{
+    Binary, CanonicalAddr, Coin, Env, Order, ReadonlyStorage, StdError, StdResult, Storage,
+};
+use cosmwasm_storage::{bucket, bucket_read, prefixed_read, Bucket, ReadonlyBucket};
+use cw20::Expiration;
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug, Default)]
+pub struct AtomicSwap {
+    /// This is the sha-256 hash of the preimage
+    pub hash: Binary,
```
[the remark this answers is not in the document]
maurolacy: Glad to do it. The end result is better.

contracts/cw20-atomic-swap/src/state.rs
```diff
@@ -0,0 +1,96 @@
...
+    pub recipient: CanonicalAddr,
+    pub source: CanonicalAddr,
+    pub expires: Expiration,
+    /// Balance in native tokens
+    pub balance: Vec<Coin>,
+}
+
+impl AtomicSwap {
+    pub fn is_expired(&self, env: &Env) -> bool {
```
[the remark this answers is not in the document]
maurolacy: You're right. Duly noted.

contracts/cw20-atomic-swap/examples/schema.rs
```diff
@@ -0,0 +1,23 @@
...
+use cw20_atomic_swap::msg::QueryMsg;
+
+fn main() {
+    let mut out_dir = current_dir().unwrap();
+    out_dir.push("schema");
+    create_dir_all(&out_dir).unwrap();
+    remove_schemas(&out_dir).unwrap();
+
+    export_schema(&schema_for!(InitMsg), &out_dir);
+    export_schema_with_title(&mut schema_for!(HandleMsg), &out_dir, "HandleMsg");
```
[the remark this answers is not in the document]
maurolacy: OK.

PR #61, opened 2020-08-26
Fix order of contract version setup

maurolacy:
Small PR to fix version setup order in init().


ethanfrey:
Please add a test (ideally before and after)

ethanfrey:
Thanks for the commit. I went with #62, which is functionally the same, just to let a new contributor into the repo.

PR #62, opened 2020-08-26
workshop subkey PR

whalelephant:
- add test for contract name check
- use whitelist init before set_contract_version to reset version on contract

### DIFF
--- a/contracts/cw1-subkeys/src/contract.rs
+++ b/contracts/cw1-subkeys/src/contract.rs
@@ -32,7 +32,9 @@ pub fn init<S: Storage, A: Api, Q: Querier>(
         version: CONTRACT_VERSION.to_string(),
     };
     let result = whitelist_init(deps, env, msg);
-    set_contract_version(&mut deps.storage, &version)?;
+    if result.is_ok() {
+        set_contract_version(&mut deps.storage, &version)?;
+    };
     result
 }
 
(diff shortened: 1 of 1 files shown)


contracts/cw1-subkeys/src/contract.rs
```diff
@@ -344,6 +348,42 @@ mod tests {
         }
     }
 
+        
+    #[test]
+    fn get_contract_version_works() {
```
ethanfrey, line 353: Thanks for the testcase. This could be a bit shorter (simple init), but this works as well.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -31,8 +31,11 @@ pub fn init<S: Storage, A: Api, Q: Querier>(
         contract: CONTRACT_NAME.to_string(),
         version: CONTRACT_VERSION.to_string(),
     };
-    set_contract_version(&mut deps.storage, &version)?;
-    whitelist_init(deps, env, msg)
+    let result = whitelist_init(deps, env, msg);
+    if result.is_ok() {
```
ethanfrey, line 35: Nice one, only set on success. 

I guess

```rust
let result = whitelist_init(deps, env, msg)?;
set_contract_version(&mut deps.storage, &version)?;
Ok(result);
```

would be the same, but this works as well

ethanfrey (approved):
Thanks for the contribution

PR #70, opened 2020-08-30
Update with new wasm, new queries

ethanfrey:
Update c20 helper to 0.2.0 wasm binary
Expose queries for `all_allowances` and `all_account`


ethanfrey:
@abefernan You can use this helper for building the cw20 wallet.

CodeID is 28: https://coralnet.wasm.glass/#/codes/28

PR #72, opened 2020-09-03
Cw20 atomic swaps

maurolacy:
Cw20 swap functionality.

### DIFF
--- a/contracts/cw20-atomic-swap/Cargo.toml
+++ b/contracts/cw20-atomic-swap/Cargo.toml
@@ -18,6 +18,7 @@ singlepass = ["cosmwasm-vm/default-singlepass"]
 library = []
 
 [dependencies]
+cw20-escrow = { path = "../cw20-escrow", version = "0.2.0", features = ["library"] }
 cw2 = { path = "../../packages/cw2", version = "0.2.0" }
 cw20 = { path = "../../packages/cw20", version = "0.2.0" }
 cosmwasm-std = { version = "0.10.0", features = ["iterator"] }
--- a/contracts/cw20-atomic-swap/src/contract.rs
+++ b/contracts/cw20-atomic-swap/src/contract.rs
@@ -75,7 +75,7 @@ pub fn try_create<S: Storage, A: Api, Q: Querier>(
         return Err(StdError::generic_err("Invalid atomic swap id"));
     }
 
-    // FIXME? What about a non-empty array, but with zero-valued coins?
+    // FIXME: normalize array first (remove zero-valued coins), and then check for empty
     if balance.is_empty() {
         return Err(StdError::generic_err(
             "Send some coins to create an atomic swap",
--- a/contracts/cw20-atomic-swap/src/contract.rs
+++ b/contracts/cw20-atomic-swap/src/contract.rs
@@ -1,17 +1,18 @@
 use sha2::{Digest, Sha256};
 
 use cosmwasm_std::{
-    log, to_binary, Api, BankMsg, Binary, CosmosMsg, Env, Extern, HandleResponse, HumanAddr,
-    InitResponse, Querier, StdError, StdResult, Storage, WasmMsg,
+    from_binary, log, to_binary, Api, BankMsg, Binary, CosmosMsg, Env, Extern, HandleResponse,
+    HumanAddr, InitResponse, Querier, StdError, StdResult, Storage, WasmMsg,
 };
 use cw2::set_contract_version;
+use cw20::{Cw20HandleMsg, Cw20ReceiveMsg};
 
-use crate::balance::Balance;
+use crate::balance::{Balance, Cw20Coin};
 use crate::msg::{
     is_valid_name, CreateMsg, DetailsResponse, HandleMsg, InitMsg, ListResponse, QueryMsg,
+    ReceiveMsg,
 };
 use crate::state::{all_swap_ids, atomic_swaps, atomic_swaps_read, AtomicSwap};
-use cw20::Cw20HandleMsg;
 
 // Version info, for migration info
 const CONTRACT_NAME: &str = "crates.io:cw20-atomic-swap";
@@ -41,6 +42,25 @@ pub fn handle<S: Storage, A: Api, Q: Querier>(
         ),
         HandleMsg::Release { id, preimage } => try_release(deps, env, id, preimage),
         HandleMsg::Refund { id } => try_refund(deps, env, id),
+        HandleMsg::Receive(msg) => try_receive(deps, env, msg),
+    }
+}
+
+pub fn try_receive<S: Storage, A: Api, Q: Querier>(
+    deps: &mut Extern<S, A, Q>,
+    env: Env,
+    wrapper: Cw20ReceiveMsg,
+) -> StdResult<HandleResponse> {
+    let msg: ReceiveMsg = match wrapper.msg {
+        Some(bin) => from_binary(&bin),
+        None => Err(StdError::parse_err("ReceiveMsg", "no data")),
+    }?;
+    let token = Cw20Coin {
+        address: deps.api.canonical_address(&env.message.sender)?,
+        amount: wrapper.amount,
+    };
+    match msg {
+        ReceiveMsg::Create(create) => try_create(deps, env, create, Balance::Cw20(token)),
     }
 }
 
@@ -54,7 +74,7 @@ pub fn try_create<S: Storage, A: Api, Q: Querier>(
         return Err(StdError::generic_err("Invalid atomic swap id"));
     }
 
-    // FIXME? A Non-empty array, but with zero-valued coins
+    // FIXME? What about a non-empty array, but with zero-valued coins?
     if balance.is_empty() {
         return Err(StdError::generic_err(
             "Send some coins to create an atomic swap",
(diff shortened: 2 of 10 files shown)


contracts/cw20-atomic-swap/Cargo.toml
```diff
@@ -18,6 +18,7 @@ singlepass = ["cosmwasm-vm/default-singlepass"]
 library = []
 
 [dependencies]
+cw20-escrow = { path = "../cw20-escrow", version = "0.2.0", features = ["library"] }
```
orkunkl, line 21: I guess it is better to resolve https://github.com/CosmWasm/cosmwasm-plus/issues/73 before finalizing this PR. Packages are getting tangled...

maurolacy: It shouldn't be a problem. Just use the `library` feature, as above, when importing stuff from other contracts into your contract.

maurolacy: We can always refactor all contracts with the changes proposed in #73 later.

ethanfrey: Yeah, library does allow that. I agree we should not have that dependency, but rather pull out cw20 coin into cw20 package or such. But happy to merge that and tackle the change in #73

contracts/cw20-atomic-swap/src/contract.rs
```diff
@@ -31,22 +37,48 @@ pub fn handle<S: Storage, A: Api, Q: Querier>(
     msg: HandleMsg,
 ) -> StdResult<HandleResponse> {
     match msg {
-        HandleMsg::Create(msg) => try_create(deps, env, msg),
+        HandleMsg::Create(msg) => try_create(
+            deps,
+            env.clone(),
```
ethanfrey, line 42: Why clone? We never use it elsewhere

contracts/cw20-atomic-swap/src/contract.rs
```diff
@@ -31,22 +37,48 @@ pub fn handle<S: Storage, A: Api, Q: Querier>(
...
+            msg,
+            Balance::Native(env.message.sent_funds),
```
ethanfrey, line 44: This is interesting... so `try_create` works for both native and cw20 coins.
It would be good to write some small docs explaining this pattern (for the docs site - talk with @orkunkl )

contracts/cw20-atomic-swap/src/contract.rs
```diff
@@ -31,22 +37,48 @@ pub fn handle<S: Storage, A: Api, Q: Querier>(
...
+        ),
         HandleMsg::Release { id, preimage } => try_release(deps, env, id, preimage),
         HandleMsg::Refund { id } => try_refund(deps, env, id),
+        HandleMsg::Receive(msg) => try_receive(deps, env, msg),
+    }
+}
+
+pub fn try_receive<S: Storage, A: Api, Q: Querier>(
```
ethanfrey, line 52: Nice

contracts/cw20-atomic-swap/src/contract.rs
```diff
@@ -31,22 +37,48 @@ pub fn handle<S: Storage, A: Api, Q: Querier>(
...
     env: Env,
     msg: CreateMsg,
+    balance: Balance,
 ) -> StdResult<HandleResponse> {
     if !is_valid_name(&msg.id) {
         return Err(StdError::generic_err("Invalid atomic swap id"));
     }
 
-    if env.message.sent_funds.is_empty() {
+    // FIXME? What about a non-empty array, but with zero-valued coins?
```
ethanfrey, line 80: That should be rejected. We can have a "normalize" function that removes all 0 values

contracts/cw20-atomic-swap/src/contract.rs
```diff
@@ -160,16 +192,37 @@ fn parse_hex_32(data: &str) -> StdResult<Vec<u8>> {
     }
 }
 
-fn send_native_tokens(from: &HumanAddr, to: &HumanAddr, amount: Vec<Coin>) -> Vec<CosmosMsg> {
+fn send_tokens<A: Api>(
```
ethanfrey, line 195: This is a nice one.

Let's remember this one for #73 I think there are quite a few helpers we may want to pull out.

contracts/cw20-atomic-swap/src/contract.rs
```diff
@@ -622,8 +684,121 @@ mod tests {
                 recipient: create2.recipient,
                 source: sender2,
                 expires: create2.expires,
-                balance
+                balance: BalanceHuman::Native(balance),
             }
         );
     }
+
+    #[test]
```
ethanfrey, line 692: In general, looks good. I need to read through this again to make sure all the test cases are covered.

ethanfrey (commented):
Looks good.

I think some of the helpers to combine native and cw20 coins can be extended to make this even cleaner. But that is really for #73 - happy to see how you combined them already and I think we could do something similar with cw20-escrow as well.

I want to go through the test cases again to make sure it is all covered, but in general looks quite good.

contracts/cw20-atomic-swap/src/contract.rs
maurolacy: OK.

contracts/cw20-atomic-swap/src/contract.rs
maurolacy: OK, I'll assign me #73 next.

contracts/cw20-atomic-swap/src/contract.rs
maurolacy: Yes... one of the advantages of joining both coins in one set of functions is that most of the tests for invalid values / checks work for both coins, and so have to be written only once.

I'll recheck everything too, just to err on the safe side.

ethanfrey: > most of the tests for invalid values / checks work for both coins, and so have to be written only once.

This is key and helps avoid not just duplicate code but bugs. In fact the opyn hack was based on them testing the erc20 path extensively but not the ETH path. I think making helpers that unify cw20 and native coins is super useful to both make it easier to build and to avoid such bugs

contracts/cw20-atomic-swap/src/contract.rs
maurolacy: It's needed because of the passing by value, just below (`Balance::Native(env.message.sent_funds)`). Structs have this behaviour in Rust by default... will try to refactor into an assignment and a construction.

contracts/cw20-atomic-swap/src/contract.rs
maurolacy: OK. I'll change the comment now, and address that as part of #73.

contracts/cw20-atomic-swap/src/contract.rs
ethanfrey: Coin contains strings, and cannot be Copy, only Clone.
I missed the other usage below... my bad. It was fine as it was, I just needed :coffee:

maurolacy: No problem. I'll just close the issue I wrote regarding Copy coins then.

contracts/cw20-atomic-swap/src/contract.rs
```diff
@@ -622,8 +682,121 @@ mod tests {
...
+            hash: real_hash(),
+            recipient: cw20_rcpt.clone(),
+            expires: Expiration::AtHeight(123000),
+        };
+        let receive = Cw20ReceiveMsg {
+            sender: cw20_sender,
+            amount: cw20_coin.amount,
+            msg: Some(to_binary(&HandleMsg::Create(create)).unwrap()),
+        };
+        let token_contract = cw20_coin.address; // TODO: Confirm
```
ethanfrey, line 738: Yes, this is correct

contracts/cw20-atomic-swap/src/contract.rs
```diff
@@ -622,8 +682,121 @@ mod tests {
...
+        assert_eq!(1, res.messages.len());
+        assert_eq!(log("action", "release"), res.log[0]);
+        assert_eq!(log("id", cw20_swap_id), res.log[1]);
+
+        // Verify the resulting Cw20 transfer message
+        let send_msg = Cw20HandleMsg::Transfer {
+            recipient: cw20_rcpt,
+            amount: cw20_coin.amount,
+        };
+        assert_eq!(
```
ethanfrey, line 765: Yup. I wonder if some of these asserts could be made easier with a few helpers, but not sure

contracts/cw20-atomic-swap/src/contract.rs
```diff
@@ -622,8 +682,121 @@ mod tests {
...
+                send: vec![]
+            })
+        );
+
+        // Now somebody (typically, B) releases the original offer on the Native (X) blockchain,
+        // using the (now public) preimage
+        let env = mock_env("other_somebody", &[]);
+
+        // First, let's obtain the preimage from the logs of the release() transaction on Y
+        let preimage_log = &res.log[2];
```
ethanfrey, line 779: :+1:

ethanfrey (approved):
Looks good and thanks for the code cleanup

PR #75, opened 2020-09-07
cw1-subkeys: Implement permission functionality

orkunkl:
Implements permission functionality for messages:

- Delegate
- Undelegate
- Redelegate
- Withdraw

### DIFF
--- a/contracts/cw1-subkeys/README.md
+++ b/contracts/cw1-subkeys/README.md
@@ -16,6 +16,9 @@ account can try to execute a `CosmosMsg::Bank(BankMsg::Send{})` from this
 contract and if they have the required allowances, their allowance will be
 reduced and the send message relayed. If they don't have sufficient authorization,
 or if they try to proxy any other message type, then the attempt will be rejected.
+Admin can give permissions to subkeys to relay specific types of messages 
+(covers _Delegate, Undelegate, Redelegate, Withdraw_ for now). Subkeys have no permission
+on creation, it can be setup with `SetupPermission` message.
 
 ### Messages
 
@@ -34,6 +37,10 @@ enum HandleMsg {
         denom: String,
         amount: Uint128,
         expires: Option<Expiration>,
+    },
+    SetupPermissions {
+        spender: HumanAddr,
+        permissions: Permissions,
     }
 }
 ```
@@ -46,12 +53,22 @@ It also adds one more query type:
 enum QueryMsg {
     Allowance {
         spender: HumanAddr,
-    }
+    },
+    AllAllowances {
+        start_after: Option<HumanAddr>,
+        limit: Option<u32>,
+    },
 }
 
-pub struct AllowanceResponse {
-    pub allowance: Vec<Coin>,
+pub struct AllowanceInfo {
+    pub spender: HumanAddr,
+    pub balance: Balance,
     pub expires: Expiration,
+    pub permissions: Permissions,
+}
+
+pub struct AllAllowancesResponse {
+    pub allowances: Vec<AllowanceInfo>,
 }
 ```
 
--- a/contracts/cw1-subkeys/helpers.ts
+++ b/contracts/cw1-subkeys/helpers.ts
@@ -135,9 +135,12 @@ const useOptions = (options: Options): Network => {
 
 type Expiration = { at_height: { height: number } } | { at_time: { time: number } } | { never: {}}
 
+type Permissions = { delegate: boolean } | { undelegate: boolean } | { redelegate: boolean } | { withdraw: boolean }
+
 interface AllowanceResponse {
   readonly balance: readonly Coin[],
   readonly expires: Expiration,
+  readonly permission: Permissions,
 }
 
 interface AdminListResponse {
@@ -176,6 +179,7 @@ interface CW1Instance {
   updateAdmins: (admins: readonly string[]) => Promise<string>
   increaseAllowance: (recipient: string, amount: Coin, expires?: Expiration) => Promise<string>
   decreaseAllowance: (recipient: string, amount: Coin, expires?: Expiration) => Promise<string>
+  setupPermissions: (recipient: string, permission: Permissions) => Promise<string>
 }
 
 interface CW1Contract {
@@ -231,7 +235,12 @@ const CW1 = (client: SigningCosmWasmClient): CW1Contract => {
       const result = await client.execute(contractAddress, {decrease_allowance: {spender, amount, expires}});
       return result.transactionHash;
     }
-    
+
+    const setupPermissions = async (spender: string, permission: Permissions): Promise<string> => {
+      const result = await client.execute(contractAddress, {setup_permissions: {spender, permission}});
+      return result.transactionHash;
+    }
+
     return {
       contractAddress,
       admins,
@@ -241,6 +250,7 @@ const CW1 = (client: SigningCosmWasmClient): CW1Contract => {
       updateAdmins,
       increaseAllowance,
       decreaseAllowance,
+      setupPermissions
     };
   }
 
--- a/contracts/cw1-subkeys/helpers.ts
+++ b/contracts/cw1-subkeys/helpers.ts
@@ -140,7 +140,7 @@ type Permissions = { delegate: boolean } | { undelegate: boolean } | { redelegat
 interface AllowanceResponse {
   readonly balance: readonly Coin[],
   readonly expires: Expiration,
-  readonly permission: Permissions,
+  readonly permissions: Permissions,
 }
 
 interface AdminListResponse {
@@ -154,14 +154,23 @@ interface InitMsg {
 }
 
 // TODO: define more of these
-type CosmosMsg = SendMsg; 
+type CosmosMsg = {};
 
-interface SendMsg {
+interface SendMsg extends CosmosMsg {
   readonly bank: {
     readonly send: {
       readonly from_address: string,
       readonly to_address: string,
       readonly amount: readonly Coin[],    
+    }
+  }
+}
+
+interface DelegateMsg extends CosmosMsg {
+  readonly staking: {
+    readonly delegate: {
+      readonly validator: string,
+      readonly amount: readonly Coin[],
     }
   }
 }
@@ -179,7 +188,7 @@ interface CW1Instance {
   updateAdmins: (admins: readonly string[]) => Promise<string>
   increaseAllowance: (recipient: string, amount: Coin, expires?: Expiration) => Promise<string>
   decreaseAllowance: (recipient: string, amount: Coin, expires?: Expiration) => Promise<string>
-  setupPermissions: (recipient: string, permission: Permissions) => Promise<string>
+  setupPermissions: (recipient: string, permissions: Permissions) => Promise<string>
 }
 
 interface CW1Contract {
@@ -236,8 +245,8 @@ const CW1 = (client: SigningCosmWasmClient): CW1Contract => {
       return result.transactionHash;
     }
 
-    const setupPermissions = async (spender: string, permission: Permissions): Promise<string> => {
-      const result = await client.execute(contractAddress, {setup_permissions: {spender, permission}});
+    const setupPermissions = async (spender: string, permissions: Permissions): Promise<string> => {
+      const result = await client.execute(contractAddress, {setup_permissions: {spender, permissions}});
       return result.transactionHash;
     }
 
@@ -283,7 +292,7 @@ const CW1 = (client: SigningCosmWasmClient): CW1Contract => {
 
 // Demo:
 // const client = await useOptions(coralnetOptions).setup(PASSWORD);
-// const { address} = await client.getAccount()
+// const { address } = await client.getAccount()
 // const factory = CW1(client)
 //
 // const codeId = await factory.upload();
--- a/contracts/cw1-subkeys/helpers.ts
+++ b/contracts/cw1-subkeys/helpers.ts
@@ -135,7 +135,12 @@ const useOptions = (options: Options): Network => {
 
 type Expiration = { at_height: { height: number } } | { at_time: { time: number } } | { never: {}}
 
-type Permissions = { delegate: boolean } | { undelegate: boolean } | { redelegate: boolean } | { withdraw: boolean }
+interface Permissions {
+  readonly delegate: boolean
+  readonly undelegate: boolean
+  readonly redelegate: boolean
+  readonly withdraw: boolean
+}
 
 interface AllowanceResponse {
   readonly balance: readonly Coin[],
@@ -361,3 +366,13 @@ const CW1 = (client: SigningCosmWasmClient): CW1Contract => {
 // contract.execute([{bank: {send: {from_address: contractAddress, to_address: address, amount: [{denom: "ushell", amount: "440000"}]}}}])
 // client.getAccount(contractAddress)
 // client.getAccount()
+
+// let permissions: Permissions = { delegate: true, undelegate: true, redelegate: true, withdraw: true }
+// contract.setupPermissions(randomAddress, permissions)
+
+// test delegating and undelegating from another account
+// let dmsg: DelegateMsg = {staking: {delegate: {validator:"coralvaloper1hf50trj7plz2sd8cmcvn7c8ruh3tjhc2uch4gp", amount:{denom:"ureef",amount:"999"}}}}
+// contract.execute([dmsg])
+//
+// let unmsg: UndelegateMsg = {staking: {undelegate: {validator:"coralvaloper1hf50trj7plz2sd8cmcvn7c8ruh3tjhc2uch4gp", amount:{denom:"ureef",amount:"999"}}}}
+// contract.execute([unmsg])
--- a/contracts/cw1-subkeys/helpers.ts
+++ b/contracts/cw1-subkeys/helpers.ts
@@ -241,7 +241,7 @@ interface CW1Instance {
   updateAdmins: (admins: readonly string[]) => Promise<string>
   increaseAllowance: (recipient: string, amount: Coin, expires?: Expiration) => Promise<string>
   decreaseAllowance: (recipient: string, amount: Coin, expires?: Expiration) => Promise<string>
-  setStakingPermissions: (recipient: string, permissions: Permissions) => Promise<string>
+  setPermissions: (recipient: string, permissions: Permissions) => Promise<string>
 }
 
 interface CW1Contract {
@@ -314,8 +314,8 @@ const CW1 = (client: SigningCosmWasmClient): CW1Contract => {
       return result.transactionHash;
     }
 
-    const setStakingPermissions = async (spender: string, permissions: Permissions): Promise<string> => {
-      const result = await client.execute(contractAddress, {setup_permissions: {spender, permissions}});
+    const setPermissions = async (spender: string, permissions: Permissions): Promise<string> => {
+      const result = await client.execute(contractAddress, {set_permissions: {spender, permissions}});
       return result.transactionHash;
     }
 
@@ -332,7 +332,7 @@ const CW1 = (client: SigningCosmWasmClient): CW1Contract => {
       updateAdmins,
       increaseAllowance,
       decreaseAllowance,
-      setStakingPermissions
+      setPermissions
     };
   }
 
--- a/contracts/cw1-subkeys/helpers.ts
+++ b/contracts/cw1-subkeys/helpers.ts
@@ -221,7 +221,7 @@ interface CW1Instance {
   updateAdmins: (admins: readonly string[]) => Promise<string>
   increaseAllowance: (recipient: string, amount: Coin, expires?: Expiration) => Promise<string>
   decreaseAllowance: (recipient: string, amount: Coin, expires?: Expiration) => Promise<string>
-  setupPermissions: (recipient: string, permissions: Permissions) => Promise<string>
+  setStakingPermissions: (recipient: string, permissions: Permissions) => Promise<string>
 }
 
 interface CW1Contract {
@@ -278,7 +278,7 @@ const CW1 = (client: SigningCosmWasmClient): CW1Contract => {
       return result.transactionHash;
     }
 
-    const setupPermissions = async (spender: string, permissions: Permissions): Promise<string> => {
+    const setStakingPermissions = async (spender: string, permissions: Permissions): Promise<string> => {
       const result = await client.execute(contractAddress, {setup_permissions: {spender, permissions}});
       return result.transactionHash;
     }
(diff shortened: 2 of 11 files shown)


packages/cw0/src/expiration.rs
```diff
@@ -14,6 +15,16 @@ pub enum Expiration {
     Never {},
 }
 
+impl fmt::Display for Expiration {
```
orkunkl, line 18: Useful for debugging

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -88,6 +90,44 @@ where
...
+                    allowances.save(owner_raw.as_slice(), &allowance)?;
+                }
+                CosmosMsg::Staking(StakingMsg::Undelegate {
+                    validator: _,
+                    amount: _,
+                }) => {
+                    if !allowance.permissions.undelegate {
+                        return Err(StdError::generic_err(PermissionErr::Undelegate{}));
+                    }
+                    // TODO Undelegation takes 21 days, it is not logical to increase balance
```
orkunkl, line 111: My only question is this @ethanfrey

contracts/cw1-subkeys/README.md
```diff
@@ -34,6 +37,10 @@ enum HandleMsg {
         denom: String,
         amount: Uint128,
         expires: Option<Expiration>,
+    },
+    SetupPermissions {
```
ethanfrey, line 41: Maybe `StakingPermissions` as a better name? Or `SetStakingPermissions`?

contracts/cw1-subkeys/README.md
```diff
@@ -46,12 +53,22 @@ It also adds one more query type:
...
         spender: HumanAddr,
-    }
+    },
+    AllAllowances {
+        start_after: Option<HumanAddr>,
+        limit: Option<u32>,
+    },
 }
 
-pub struct AllowanceResponse {
```
ethanfrey, line 52: You removed the `AllowanceResponse`? I think you need both.

Oh, this is README....

contracts/cw1-subkeys/helpers.ts
```diff
@@ -135,9 +135,17 @@ const useOptions = (options: Options): Network => {
 
 type Expiration = { at_height: { height: number } } | { at_time: { time: number } } | { never: {}}
 
+interface Permissions {
```
ethanfrey, line 138: Question:

If we add permissions for eg. governance votes, or calling wasm contracts, would you keep them here? Or use a separate type? 

To see if we want one large "Permissions" or "StakingPermissions", "VotingPermissions", etc.

contracts/cw1-subkeys/src/state.rs
```diff
@@ -6,11 +6,47 @@ use cosmwasm_storage::{bucket, bucket_read, Bucket, ReadonlyBucket};
...
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema, Default, Copy)]
+pub struct Permissions {
+    pub delegate: bool,
+    pub redelegate: bool,
+    pub undelegate: bool,
+    pub withdraw: bool,
+}
+
+impl fmt::Display for Permissions {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
```
ethanfrey, line 20: Nice

packages/cw0/src/expiration.rs
ethanfrey: Makes better error messages, right? sounds good.

contracts/cw1-subkeys/helpers.ts
```diff
@@ -231,7 +277,12 @@ const CW1 = (client: SigningCosmWasmClient): CW1Contract => {
       const result = await client.execute(contractAddress, {decrease_allowance: {spender, amount, expires}});
       return result.transactionHash;
     }
-    
+
+    const setupPermissions = async (spender: string, permissions: Permissions): Promise<string> => {
```
ethanfrey, line 281: Thanks for adding this here as well.

I think the `all_allowances` query is missing as well (my bad), can you add that too?

contracts/cw1-subkeys/helpers.ts
```diff
@@ -151,9 +159,9 @@ interface InitMsg {
 }
 
 // TODO: define more of these
-type CosmosMsg = SendMsg; 
+interface CosmosMsg {}
```
ethanfrey, line 162: I would rather make this tighter check;

`type CosmosMsg = SendMsg | DelegateMsg | UndelegateMsg | RedelegateMsg | WithdrawMsg`

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -88,6 +90,44 @@ where
                     allowance.balance = allowance.balance.sub(amount.clone())?;
                     allowances.save(owner_raw.as_slice(), &allowance)?;
                 }
+                CosmosMsg::Staking(StakingMsg::Delegate {
```
ethanfrey, line 93: This makes a quite large branch.
I would do something like:

`CosmosMsg::Staking(staking_msg) => check_staking_message(&staking_msg, &mut allowance)?` 

or such and then include all these branches inside `check_staking_message`. That function would then be simpler:

```rust
match staking_msg {
  StakingMsg:::Delegate { amount, ..} => 
                    if !allowance.permissions.delegate {
                        Err(StdError::generic_err(PermissionErr::Delegate{}))
                        // or maybe this (and cast to StdError above when `?` is used)
                        Err(PermissionErr::Delegate{})
                    } else {
                       Ok(())
                   }
}
```

Try to pull some out and simplify

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -88,6 +90,44 @@ where
...
+                    validator: _,
+                    amount,
+                }) => {
+                    if !allowance.permissions.delegate {
+                        return Err(StdError::generic_err(PermissionErr::Delegate{}));
+                    }
+                    // Decrease allowance
+                    allowance.balance = allowance.balance.sub(amount.clone())?;
```
ethanfrey, line 101: I would keep the staking items independent of balances. You can delegate and undelegate all you want, but not take money from the account. By requiring the balances here, I need to give someone the power to take my money to let them select the validators I invest in. Just the first check is needed (and see above comment about refactor to make it simpler)

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -196,6 +236,47 @@ where
...
+
+    if !cfg.is_admin(&owner_raw) {
+        return Err(StdError::unauthorized());
+    }
+    if spender_raw == owner_raw {
+        return Err(StdError::generic_err("Cannot set allowance to own account"));
+    }
+
+    // update allowance permissions
+    allowances(&mut deps.storage).update(spender_raw.as_slice(), |allow| {
```
ethanfrey, line 260: We should be able to set the staking permissions on accounts without any balances. 

It may just have balances, or just permissions, or both (and they should not interfere with each other)

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -329,15 +413,20 @@ mod tests {
         };
         init(deps, env.clone(), init_msg).unwrap();
 
-        // Add subkeys with initial allowances
-        for (spender, expiration) in spenders.iter().zip(expirations) {
+        // Add subkeys with initial allowances and permissions
+        for ((spender, expiration), permissions) in spenders.iter().zip(expirations).zip(permissions) {
```
ethanfrey, line 417: I would not add permissions here. The setup script is already too confusing for me. And this makes it harder to test individual cases. (Just set balance, just set permissions, set both). I guess they are somehow covered here, but it should be more explicit.

I was already a bit unhappy with this giant setup script and adding more makes it super hard to see what is being tested. I would rather just use this for a few of the targetted tests below, and just call `init` and a few Allowance/Permssions settings explicitly for the tests you are writing

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -939,6 +1055,84 @@ mod tests {
...
+            &initial_expirations,
+            &initial_permissions,
+        );
+
+        let msg_delegate = vec![StakingMsg::Delegate{
+            validator: HumanAddr(String::from("validator")),
+            amount: allow.clone(),
+        }.into()];
+        let msg_redelegate = vec![StakingMsg::Redelegate{
+            src_validator: Default::default(),
```
ethanfrey, line 1102: Good to have some non-defaults here... let's pick a name or two.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -939,6 +1055,84 @@ mod tests {
...
+            &env,
+            &admins,
+            &initial_spenders,
+            &initial_allowances,
+            &initial_expirations,
+            &initial_permissions,
+        );
+
+        let msg_delegate = vec![StakingMsg::Delegate{
+            validator: HumanAddr(String::from("validator")),
```
ethanfrey, line 1098: `HumanAddr::from("validator")` does the same and is less typing

ethanfrey (commented):
Thanks for your first contract PR @orkunkl 

In general, you got all the pieces connected and it looks generally correct. I added a number of comments to improve code quality. Can you go through them and also take another pass on naming design and I will then re-review this

contracts/cw1-subkeys/README.md
orkunkl: `SetStakingPermission` sounds good

contracts/cw1-subkeys/src/contract.rs
orkunkl: On it

contracts/cw1-subkeys/helpers.ts
orkunkl: Separate permissions design brings a lot of extra code(bucket) to both smart contract and helpers.ts. This struct will not contain hundreds of booleans(max 15-20?). I guess using single permissions struct is the optimal way

orkunkl: If rust and buckets provides a way to define permissions generically, then multiple permissions that extend `Permissions` trait is better.

contracts/cw1-subkeys/README.md
orkunkl: Depends on https://github.com/CosmWasm/cosmwasm-plus/pull/75/files#r484310769

contracts/cw1-subkeys/README.md
orkunkl: I wanted to make helpers.ts and msg.rs naming identical

packages/cw0/src/expiration.rs
orkunkl: Exactly

contracts/cw1-subkeys/helpers.ts
ethanfrey: Fair enough, we keep one Permissions and extend it for other packages.
Please add a comment on the type explaining that choice (actually only in helpers.rs, not here)

ethanfrey:
Looks good. One comment requested, and please `cargo fmt` so the linter is happy. Then I will do the final review to merge

contracts/cw1-subkeys/helpers.ts
orkunkl: I assume you mean state.rs by helpers.rs? or helpers.ts? I am confused 😕

ethanfrey: I meant `state.rs`
I should re-review again

contracts/cw1-subkeys/helpers.ts
```diff
@@ -150,8 +174,7 @@ interface InitMsg {
   readonly mutable: boolean,
 }
 
-// TODO: define more of these
-type CosmosMsg = SendMsg; 
+type CosmosMsg = SendMsg | DelegateMsg | UndelegateMsg | RedelegateMsg | WithdrawMsg
```
ethanfrey, line 177: :+1:

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -73,17 +84,25 @@ where
...
         Ok(res)
     } else {
-        let mut allowances = allowances(&mut deps.storage);
-        let allow = allowances.may_load(owner_raw.as_slice())?;
-        let mut allowance =
-            allow.ok_or_else(|| StdError::not_found("No allowance for this account"))?;
         for msg in &msgs {
             match msg {
+                CosmosMsg::Staking(staking_msg) => {
+                    let permissions = permissions(&mut deps.storage);
```
ethanfrey, line 90: This is fine and explicit. I generally make this more compact.

```rust
let perm = permissions(&mut deps.storage).load(owner_raw.as_slice())?;
check_staking_permissions(staking_msg, perm)?;
```

But it is fine as is. (and with custom error message)

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -103,6 +122,48 @@ where
     }
 }
 
+pub fn check_staking_permissions(
+    staking_msg: &StakingMsg,
+    permissions: Permissions,
+) -> Result<bool, PermissionErr> {
+    match staking_msg {
```
ethanfrey, line 129: Much nicer

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -103,6 +122,48 @@ where
...
+        StakingMsg::Delegate {
+            validator: _,
+            amount: _,
+        } => {
+            if !permissions.delegate {
+                return Err(PermissionErr::Delegate {});
+            }
+        }
+        StakingMsg::Undelegate {
```
ethanfrey, line 138: You can ignore the attributes like: 

```rust
StakingMsg::Undelegate { .. } => { 
            if !permissions.undelegate {
                return Err(PermissionErr::Undelegate {});
            }
}
```

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -196,17 +257,55 @@ where
...
+{
+    let cfg = admin_list_read(&deps.storage).load()?;
+    let spender_raw = &deps.api.canonical_address(&spender)?;
+    let owner_raw = &deps.api.canonical_address(&env.message.sender)?;
+
+    if !cfg.is_admin(&owner_raw) {
+        return Err(StdError::unauthorized());
+    }
+    if spender_raw == owner_raw {
+        return Err(StdError::generic_err("Cannot set allowance to own account"));
```
ethanfrey, line 277: `s/Cannot set allowance/Cannot set permission/`

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -196,17 +257,55 @@ where
...
 ) -> StdResult<Binary> {
     match msg {
         QueryMsg::AdminList {} => to_binary(&query_admin_list(deps)?),
         QueryMsg::Allowance { spender } => to_binary(&query_allowance(deps, spender)?),
+        QueryMsg::Permissions { spender } => to_binary(&query_permissions(deps, spender)?),
         QueryMsg::CanSend { sender, msg } => to_binary(&query_can_send(deps, sender, msg)?),
         QueryMsg::AllAllowances { start_after, limit } => {
             to_binary(&query_all_allowances(deps, start_after, limit)?)
         }
+        QueryMsg::AllPermissions { start_after, limit } => {
```
ethanfrey, line 306: Nice addition

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -243,19 +354,27 @@ fn can_send<S: Storage, A: Api, Q: Querier>(
...
+            // now we check if there is enough allowance for this message
+            let allowance = allowances_read(&deps.storage).may_load(owner_raw.as_slice())?;
+            match allowance {
+                // if there is an allowance, we subtract the requested amount to ensure it is covered (error on underflow)
+                Some(allow) => Ok(allow.balance.sub(amount).is_ok()),
+                None => Ok(false),
+            }
+        }
+        CosmosMsg::Staking(staking_msg) => {
+            permissions_read(&deps.storage)
```
ethanfrey, line 368: Interesting chain. 

I think the `?` and `match` approach for allowances is more legible. But nice that you learned a bit more about map and map_or and handling Options and Results. Key skills for working with rust

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -505,6 +650,162 @@ mod tests {
...
+
+        let setup_perm_msg2 = HandleMsg::SetPermissions {
+            spender: spender2.clone(),
+            // default is no permission
+            permissions: Default::default(),
+        };
+        // default is no permission
+        handle(&mut deps, env.clone(), setup_perm_msg2).unwrap();
+
+        let permissions = query_permissions(&deps, spender1.clone()).unwrap();
```
ethanfrey, line 696: Looks good. I would add one more test, where you set the allowance for someone with a permission and then ensure that both are set properly (setting allowance doesn't unset the permission)

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -939,6 +1240,88 @@ mod tests {
...
+
+        // spender1 can execute
+        for msg in &msgs {
+            let env = mock_env(&spender1, &[]);
+            let res = handle(&mut deps, env, HandleMsg::Execute { msgs: msg.clone() });
+            assert!(res.is_ok())
+        }
+
+        // spender2 cannot execute (no permission)
+        for msg in &msgs {
```
ethanfrey, line 1319: I would add a few more tests where one sender has eg. 2 permissions set, 2 unset and ensure that those two pass, and the other 2 fail. 

Just a double check that there is no typo in which permission is checked for which message. And ensures future changes don't break this (what if we swap delegate and redelegate permission checks? these tests would still pass)

ethanfrey (commented):
Much nicer.

If you could cover a few more cases in the test code, I can merge this

contracts/cw1-subkeys/src/contract.rs
orkunkl: Just some old scala and functional programming skills λ

orkunkl: Applied the `?` and `match` case.
But `can_send` method looks weird to me. The return value is `StdResult` but in tests, it is expected to be always boolean. Which throws the error information to the thrash 🤔

contracts/cw1-subkeys/src/contract.rs
orkunkl: Added the test. Developer errors must be more distinctive and functional than just a str :( https://github.com/CosmWasm/cosmwasm/issues/518

contracts/cw1-subkeys/src/contract.rs
ethanfrey: Error should only throw if we have trouble parsing from disk, and we cannot *prove* that will never happen.

In all normal cases (non-corrupt db), we should get true or false. You would have to write invalid json to the key to trigger an error

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -365,14 +354,14 @@ fn can_send<S: Storage, A: Api, Q: Querier>(
...
-                .may_load(owner_raw.as_slice())
-                .map(|perm_opt| {
-                    // if permission exists
-                    perm_opt.map_or(false, |perm| {
-                        check_staking_permissions(&staking_msg, perm).is_ok()
-                    })
-                })
+            let perm_opt = permissions_read(&deps.storage).may_load(owner_raw.as_slice())?;
+            match perm_opt {
+                Some(permission) => match check_staking_permissions(&staking_msg, permission) {
```
ethanfrey, line 359: This could actually be:

`Some(permission) => Ok(check_staking_permissions(&staking_msg, permission).is_ok())`

Error only returned from `may_load` unable to parse.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -1253,7 +1243,7 @@ mod tests {
         let spender2 = HumanAddr::from("spender0002");
         let denom = "token1";
         let amount = 10000;
-        let allow = coin(amount, denom);
+        let coin = coin(amount, denom);
```
ethanfrey, line 1246: This shadows the function `coin` for the rest of this test function. I would avoid that 

(I only shadow one variable with another when transforming it, eg. `let msg = msg.to_string();`)

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -1321,7 +1316,128 @@ mod tests {
...
+        handle(&mut deps, env.clone(), setup_perm_msg3).unwrap();
+        let env = mock_env(&spender3, &[]);
+        let res = handle(
+            &mut deps,
+            env.clone(),
+            HandleMsg::Execute {
+                msgs: msg_delegate.clone(),
+            },
+        );
+        // FIXME need better error check here
```
ethanfrey, line 1340: You want typed errors... I know we will make this happen

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -1321,7 +1316,128 @@ mod tests {
...
+            HandleMsg::Execute {
+                msgs: msg_withdraw.clone(),
+            },
+        );
+        assert!(res.is_err())
+    }
+
+    // tests permissions and allowances are independent features and does not affect each other
+    #[test]
+    fn permissions_allowances_independent() {
```
ethanfrey, line 1370: Thank you.

ethanfrey (approved):
Looks good.

Two minor nits on code style. You can ideally change them (or ignore them if you disagree strongly) and then ready for merge.

contracts/cw1-subkeys/src/contract.rs
orkunkl: I see :+1:  Renamed it to `coin1`. `allow` was not a good fit there.

orkunkl:
2 points left:
1. `StdErr` which is taken care of. We can keep this PR on hold until it is there. It's a minor improve so good to merge IMO.
2. This is important. In helpers.ts, [wasm source URL](https://github.com/CosmWasm/cosmwasm-plus/blob/e88048b51d5b7a7ccfc8b448847943f85a1cc75a/contracts/cw1-subkeys/helpers.ts#L352) is pointed to v0.1.1 tag. After merge, cli helper will be broken. It is a good practice to use tags in `helpers.ts` files but to make it work we need to release new version after every contract change.... I don't see a problem with releasing minor versions. @ethanfrey ideas?

ethanfrey:
`helpers.ts` is a bit of an issue. current flow is this:

1. Create 0.2.1 tag of rust code
2. Package is auto-built (20 minutes later)
3. Update helpers.ts to point to new code binary

Meaning 0.2.1 helpers.ts points to 0.2.0 (or 0.1.1) code.

I am happy to make a new tag shortly and then we have the updated helpers, but this chicken-and-egg problem is what stops us from using tagged helpers.ts, which is the real solution. Any ideas for work-arounds?

ethanfrey:
I am happy to merge this when you give the :+1:

And we can cut a release of cosmwasm-plus contract today

orkunkl:
Houston, the shuttle is ready to launch 🚀  👍

Issue #76, opened 2020-09-07
Check for / reject zero amount tokens


maurolacy:
Add a check for considering zero amounts as missing.

For native coins, filter out zeros, and then ensure then remaining balance has length > 0. Only save the filtered list.

For cw20 coins, reject zero amounts directly. Add a check to also reject such sends in cw20-base.

ethanfrey:
As part of this, let's pull out the [`Balance` helper](https://github.com/CosmWasm/cosmwasm-plus/blob/master/contracts/cw1-subkeys/src/balance.rs) from cw1-subkeys into `cw0` (as it only relies on cosmwasm-std and used by many different protocols).

This already has a [`normalize` function](https://github.com/CosmWasm/cosmwasm-plus/blob/master/contracts/cw1-subkeys/src/balance.rs#L26-L27) which you could use (or add a more efficient `is_empty` one).

This would be a great help to build the higher-level checks for native coins in all contracts.

I would also suggest 2 PRs here: one for the cw20 additions, one for the native coin changes (that I mentioned here)

maurolacy:
Yes, I was taking a look at the subkeys code, and was going to suggest exactly that.

ethanfrey:
This is done with #86 and #87, correct?

maurolacy:
Correct.

PR #78, opened 2020-09-10
Cw20coin refactoring

### LINKED ISSUE #73: Cw20Coin duplication
When I was looking into how to implement cw20 compatible atomic swaps in noticed the contract needs to define a construct for Cw20 tokens. I checked other contracts and saw Cw20Token is defined in [cw20-escrow contract](http://github.com/CosmWasm/cosmwasm-plus/blob/84d9ba6f406ee2c6954afa5fea78ad24865a29e4/contracts/cw20-escrow/src/state.rs#L33-L37). We need to define this struct either in cw20-base or cw20 to prevent duplication in contracts that use cw20

maurolacy:
A small PR to move `Cw20Coin` to the `cw20` package (closes #73).

Will also add coin normalization (#76) to cw20-atomic-swaps here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -275,12 +275,12 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-channel"
-version = "0.4.3"
+version = "0.4.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "09ee0cc8804d5393478d743b035099520087a5186f3b93fa58cec08fa62407b6"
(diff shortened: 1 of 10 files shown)


ethanfrey (approved):
The reorg looks good. This could actually be merged as-is as just the de-dup.

And then another PR to add some helpers to it (like check for 0).
And another PR to handle the normalizing for Balance (I will check those PRs)

maurolacy:
OK.

ethanfrey:
If you agree for me to merge as is, please mark as non-draft. Then I can do it

Issue #57, opened 2020-08-26
Look at serde(flatten) to simplify return value composition


ethanfrey:
https://serde.rs/attr-flatten.html

When we try to "extend" structs, especially those we serialize but don't deserialize (query responses), rather than copy the definition and add more fields, we could embed it.

Here is an example to explain what I mean. [`AllowanceResponse`](https://github.com/CosmWasm/cosmwasm-plus/blob/master/packages/cw20/src/query.rs#L58-L62) is extended to [`AllowanceInfo`](https://github.com/CosmWasm/cosmwasm-plus/blob/master/packages/cw20/src/query.rs#L71-L76). Currently that looks like:

```rust
#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug, Default)]
pub struct AllowanceResponse {
    pub allowance: Uint128,
    pub expires: Expiration,
}

#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug, Default)]
pub struct AllowanceInfo {
    pub spender: HumanAddr,
    pub allowance: Uint128,
    pub expires: Expiration,
}
```

Maybe it would be possible to do:

```rust
#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug, Default)]
pub struct AllowanceInfo {
    #[serde(flatten)]
    pub base: AllowanceResponse,
    pub expires: Expiration,
}
```

This PR should not only worry about the code compiling and working in Rust, but also check the JSON schema we expose externally. This can be demo-ed on a few values like above and then we can see if this works at all, and if it works, if it simplifies or complicates things.

**Note**: This may not work with the custom serde-json-wasm library we use. If flatten is not supported, or raises errors, then please raise an issue on that repo (our custom json lib) and close this issue. (We are unlikely to implement it)

Deliverables:
- [ ] Update `AllowanceInfo` as defined
- [ ] Update all contracts to use it properly
- [ ] Check generated json schema (should be the same) and manually print out an example of serializing it
- [ ] Test serializing/deserializing with this struct

In addition, please provide a short response if this seems to make the coding simpler (extending single query to enumerable) and if you feel it improves code quality, a list of other structs to consider in future issues. (This may just add confusion, and we may not make this change, but I wanted to try it out before building too many List queries)

ethanfrey:
[...]

PR #84, opened 2020-09-11
WIP: Feedback needed - serde[(flatten)]

whalelephant:
Hi,

This PR is ref to #57 

- Th current code has been tested to build to wasm successfully and pass tests in /packages and /contracts.
- The additional tests may be misplaced
- Not sure what to call the flattened field name, currently `allowance_response` which is helpful for maintainers but probably not so for the users.

I think it will be nicer to maintain "extended" structs by using flatten, this assumes we can reuse the basic structs many times. However, it is more verbose when developers have to use "extended" structs.


contracts/cw20-base/src/enumerable.rs
```diff
@@ -27,8 +27,10 @@ pub fn query_all_allowances<S: Storage, A: Api, Q: Querier>(
             let (k, v) = item?;
             Ok(AllowanceInfo {
                 spender: api.human_address(&CanonicalAddr::from(k))?,
-                allowance: v.allowance,
-                expires: v.expires,
+                allowance_response: AllowanceResponse {
```
ethanfrey, line 30: Can't we just do `allowance_response: v`?

Much less verbose and helps make use of flatten

packages/cw20/src/query.rs
```diff
@@ -84,3 +84,60 @@ pub struct AllAllowancesResponse {
...
+            spender: HumanAddr::from("spender"),
+        };
+        let allowance_info = serde_json::to_value(info.clone()).unwrap();
+        let serialized = serde_json::to_string(&allowance_info).unwrap();
+        let deserialized: AllowanceInfo = serde_json::from_str(&serialized).unwrap();
+        assert!(info.eq(&deserialized));
+    }
+
+    #[test]
+    fn it_should_make_correct_schema() {
```
ethanfrey, line 113: Oh, this is a very interesting idea.

What I would do would be to define the old AllowanceInfo type for tests and generate json schema for the old and new AllowanceInfo (with and without flatten) and you could compare those two schemas (asset equal)

packages/cw20/src/query.rs
```diff
@@ -84,3 +84,60 @@ pub struct AllAllowancesResponse {
 pub struct AllAccountsResponse {
     pub accounts: Vec<HumanAddr>,
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use cosmwasm_schema::schema_for;
+    use serde_json;
```
ethanfrey, line 92: Note we use a custom json serde lib for cosmwasm_std.

You can import cosmwasm_std::to_vec to use that. It would be good to ensure our custom lib supports flatten as well as the standard one

ethanfrey (commented):
Nice start. Added some thoughts on where to go with this

PR #86, opened 2020-09-12
Cw20 zero amount checks

maurolacy:
Add checks for rejecting zero-amount Cw20 tokens.

Also improves the use of `Cw20Coin`(Human) accross contracts.

Closes the Cw20 part of #76.

### DIFF
--- a/contracts/cw20-base/src/allowances.rs
+++ b/contracts/cw20-base/src/allowances.rs
@@ -265,10 +265,10 @@ mod tests {
 
     use cosmwasm_std::testing::{mock_dependencies, mock_env};
     use cosmwasm_std::{coins, CosmosMsg, StdError, WasmMsg};
-    use cw20::TokenInfoResponse;
+    use cw20::{Cw20CoinHuman, TokenInfoResponse};
 
     use crate::contract::{handle, init, query_balance, query_token_info};
-    use crate::msg::{HandleMsg, InitMsg, InitialBalance};
+    use crate::msg::{HandleMsg, InitMsg};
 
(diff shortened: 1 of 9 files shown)


contracts/cw20-base/src/allowances.rs
```diff
@@ -287,7 +287,7 @@ mod tests {
             name: "Auto Gen".to_string(),
             symbol: "AUTO".to_string(),
             decimals: 3,
-            initial_balances: vec![InitialBalance {
+            initial_balances: vec![Cw20CoinHuman {
```
ethanfrey, line 290: didn't realize these were the same one. good one

contracts/cw20-escrow/src/contract.rs
```diff
@@ -98,6 +98,11 @@ pub fn try_cw20_create<S: Storage, A: Api, Q: Querier>(
     token: Cw20Coin,
     msg: CreateMsg,
 ) -> StdResult<HandleResponse> {
+    if token.is_empty() {
```
ethanfrey, line 101: Good check on the receiver (and needed).

Can we also add a check to cw20-base that we don't even send a 0 value? We cannot rely on it here, but nice check there as well.

ethanfrey (approved):
Nice one. I would add another safety check to `cw20-base::handle_send` to not send 0 tokens, but that is a different pr/issue

contracts/cw20-escrow/src/contract.rs
maurolacy: Sure. I will create the issue.

maurolacy: Issue #89.

PR #87, opened 2020-09-12
Native balance refactoring

maurolacy:
Refactor cw1-subkeys native `Balance` into `cw0`.

Use it in the cw20-atomic-swap contract `Balance`.

Closes the native part of #76.


contracts/cw20-atomic-swap/src/balance.rs
```diff
@@ -1,27 +1,34 @@
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
-use cosmwasm_std::Coin;
 use cw20::Cw20Coin;
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
 #[serde(rename_all = "snake_case")]
 pub enum Balance {
```
ethanfrey, line 8: Balance and `cw0::Balance` is a bit confusing.

Do you have an idea for a different name for one of them? `cw0::NativeBalance`?

contracts/cw20-atomic-swap/src/contract.rs
```diff
@@ -239,7 +240,7 @@ fn query_details<S: Storage, A: Api, Q: Querier>(
 
     // Convert balance to human balance
     let balance_human = match swap.balance {
-        Balance::Native(coins) => BalanceHuman::Native(coins),
+        Balance::Native(cw0::Balance(coins)) => BalanceHuman::Native(coins),
```
ethanfrey, line 243: It would be nice to simplify this. Maybe:

`Balance::Native(coins) => BalanceHuman::Native(coins.0)`

OR

`Balance::Native(coins) => BalanceHuman::Native(coins.into())` (this may need another function)

contracts/cw20-atomic-swap/src/contract.rs
```diff
@@ -66,13 +66,14 @@ pub fn try_create<S: Storage, A: Api, Q: Querier>(
...
-    balance: Balance,
+    mut balance: Balance,
 ) -> StdResult<HandleResponse> {
     if !is_valid_name(&msg.id) {
         return Err(StdError::generic_err("Invalid atomic swap id"));
     }
 
-    // FIXME: normalize array first (remove zero-valued coins), and then check for empty
+    // Normalize coins first
+    balance.normalize();
```
ethanfrey, line 76: Maybe we should avoid the two calls and just make balance.is_empty more intelligent. That would make it hard to anyone else using this to get it wrong.

contracts/cw20-atomic-swap/src/balance.rs
```diff
@@ -1,27 +1,34 @@
...
+        Balance::Native(cw0::Balance(vec![]))
     }
 }
 
 impl Balance {
     pub fn is_empty(&self) -> bool {
         match self {
-            Balance::Native(coins) => coins.is_empty(),
-            Balance::Cw20(_) => false, // FIXME: Reject zero-valued coins
+            Balance::Native(balance) => balance.is_empty(),
```
ethanfrey, line 22: You could make this more robust with something like this:

`! balance.iter().any(|x| x.amount != 0)`

Then we don't need to normalize before checking for empty (actually this logic should be in the `balance.rs` file)

ethanfrey (approved):
Looks quite good. I just left some comments on naming and some usage details to make this a bit clearer. Logic looks good. Except maybe the `is_empty`

Also, I would rebase on master since #86 was merged.

contracts/cw20-atomic-swap/src/balance.rs
maurolacy: Mi idea was that the `cw0` of `cw0::Balance` implies native. But I agree it's confusing. It could be `NativeBalance` or `AtomBalance`. As you like / prefer.

contracts/cw20-atomic-swap/src/contract.rs
maurolacy: I thought about that, but they are different operations. And, they are being called separately in the subkeys contract.

Anyway, I don't think we'll never call `is_empty()` more than once per call / endpoint. So, I'm OK with joining them.

contracts/cw20-atomic-swap/src/balance.rs
maurolacy: Good idea. We don't put `normalize()` inside `is_empty()`, and iterate here. I'll change it.

ethanfrey: Is this done in CW-escrow as well? The empty check?

It supports a vec of cw20coin.

Maybe there is some unification possible here. Or maybe not. I don't mind is empty as is zero... nothing is there

maurolacy: In escrows it's done directly on the `amount` field, i. e. there's no use of `Coin` / `Cw20Coin` there. I looked at it to introduce it, but it wasn't worth it.

The `any`check you propose already works for empty arrays. So, I'm OK with doing that change, and we can keep the `normalize()` outside, as an improvement than can be used by the implementation.

contracts/cw20-atomic-swap/src/balance.rs
maurolacy: Refactored in favor of `NativeBalance`.

maurolacy:
I like how it is now. Normalizing is a relatively costly operation, and in this way we don't enforce it on every `is_empty()` call.

Also, with your proposal for checking empty arrays using an iterator, `is_empty` is now independent of normalization.

Finally, I moved that `is_empty()` impl to the `NativeBalance` struct directly.

packages/cw0/src/balance.rs
```diff
@@ -65,7 +65,7 @@ impl Balance {
     }
 
     pub fn is_empty(&self) -> bool {
-        self.0.is_empty()
+        !self.0.iter().any(|x| x.amount != Uint128(0))
```
ethanfrey, line 68: Thanks for updating this

contracts/cw20-atomic-swap/src/contract.rs
ethanfrey: With the change to is_empty below, you can remove this line, right?

ethanfrey (approved):
Looks good,  I would just remove the normalize and merge

packages/cw0/src/balance.rs
maurolacy: You're welcome. That's a clever check by the way.

maurolacy:
> With the change to is_empty below, you can remove this line, right?

You can, in the sense that the `is_empty()` check will still work. That is, it doesn't depend on a previous normalization anymore.

But, keeping it does no harm, as normalization is a good thing, i. e. it's better to store a normalized balance that an un-normalized one.

ethanfrey:
> But, keeping it does no harm, as normalization is a good thing, i. e. it's better to store a normalized balance that an un-normalized one.

This is a good point and if we did other manipulations on the array (like with escrow), I would add it.

For now, I would remove it and save the gas costs. And we use that where we want to do more array manipulations (and maybe optimize all the NativeBalance ops to work with normalize balances)

PR #90, opened 2020-09-15
Cw3 spec

### LINKED ISSUE #79: Define cw3 spec for multisigs
This should define pubic operations like `create_proposal`, `vote`, and `execute`. The flow should be similar to cw1 (proposal can contain any number of `CosmosMsg` but they must be proposed and voted on in various steps before they can be executed. Each successful proposal can be executed exactly once.

The multisig contains a group and one set of criteria for successful proposals (how many votes needed, voting duration)

ethanfrey:
Closes #79 

Defines a spec we can implement for various voting contracts. Happy for reviews on this interface.

There are at least 3 different cases we want to cover in this spec:

- K of N immutible multisig. One key proposes a set of messages, after K-1 others approve it, it can be executed with the multisig address.
- K of N flexible, mutable multisig. Like above, but with multiple contracts. One contract stores the group, which is referenced from multiple multisig contracts (which in turn implement cw3). One cw3 contracts is able to update the group content (maybe needing 67% vote). Other cw3 contracts may hold tokens, staking rights, etc with various execution thresholds, all controlled by one group.

Maybe this can fit in as an extension of this interface:

- Token weighted voting. People lock tokens in a contract for voting rights. There is a vote threshold to execute messages. The voting set is dynamic. We may also want to support clear YES/NO votes and quora not absolute thresholds


ethanfrey:
@webmaster128 happy if you want to take an early look at the README and comment. I will be defining types for this.

packages/cw3/README.md
```diff
@@ -0,0 +1,107 @@
...
+have failed for the given proposal. If so (eg. time expired
+and insufficient votes), then the proposal is marked `Failed`.
+This is not strictly necessary, as it will only act when
+it is impossible the contract would ever be executed,
+but can be triggered to provide some better UI.
+
+### Queries
+
+`Proposal{proposal_id}` - Returns the information set when
+creating the contract, along with the current status.
```
webmaster128, line 101:
```suggestion
creating the proposal, along with the current status.
```

packages/cw3/README.md
```diff
@@ -0,0 +1,107 @@
...
+`ReverseProposals{start_before, limit}` - Returns the same info
+as `Proposal`, but for all proposals along with pagination.
+Starts at latest proposal_id and descending. (Often this
+is what you will want for a UI)
+
+`Vote{proposal_id, voter}` - Returns how the given 
+voter (HumanAddr) voted on the proposal. (May be null)
+
+`ListVotes{proposal_id, start_after, limit}` - Returns the same info
+as `Proposal`, but for all proposals along with pagination.
```
webmaster128, line 116:
```suggestion
as `Vote`, but for all votes along with pagination.
```

packages/cw3/examples/schema.rs
```diff
@@ -0,0 +1,23 @@
...
+    out_dir.push("schema");
+    create_dir_all(&out_dir).unwrap();
+    remove_schemas(&out_dir).unwrap();
+
+    export_schema_with_title(&mut schema_for!(Cw3HandleMsg), &out_dir, "HandleMsg");
+    export_schema_with_title(&mut schema_for!(Cw3QueryMsg), &out_dir, "QueryMsg");
+    export_schema(&schema_for!(ProposalResponse), &out_dir);
+    export_schema(&schema_for!(ProposalListResponse), &out_dir);
+    export_schema(&schema_for!(VoteResponse), &out_dir);
+    export_schema(&schema_for!(VoteListResponse), &out_dir);
```
webmaster128, line 22: All those schemas are only generated for `T = Empty`, so `T` cannot be used together with the schema files. If we want to stick with the custom `T` in `CosmosMsg`, I think we should remove the schemas here. Then only the implementations generate schemas, which either set `T` or actually use `Empty`.

(the same is true for the schemas in cosmwasm-std (packages/std/schema), which are kind of pointless)

packages/cw3/src/helpers.rs
```diff
@@ -0,0 +1,80 @@
...
+        self.0.clone()
+    }
+
+    /// Convert this address to a form fit for storage
+    pub fn canonical<A: Api>(&self, api: &A) -> StdResult<Cw3CanonicalContract> {
+        let canon = api.canonical_address(&self.0)?;
+        Ok(Cw3CanonicalContract(canon))
+    }
+
+    pub fn encode_msg(&self, msg: Cw3HandleMsg) -> StdResult<CosmosMsg> {
```
webmaster128, line 27: - This does not support custom message types
- If we always return a `WasmMsg`, why not return `StdResult<WasmMsg>`. This would remove all the issues with a generic result type

packages/cw3/src/helpers.rs
```diff
@@ -0,0 +1,80 @@
...
+        }
+        .into())
+    }
+
+    /// helper doesn't support custom messages now
+    pub fn proposal<T: Into<String>, U: Into<String>>(
+        &self,
+        title: T,
+        description: U,
+        msgs: Vec<CosmosMsg>,
```
webmaster128, line 41: This does not support custom message types

packages/cw3/src/msg.rs
```diff
@@ -0,0 +1,64 @@
...
+}
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
+#[serde(rename_all = "lowercase")]
+pub enum Vote {
+    YES,
+    NO,
+    ABSTAIN,
+    VETO,
+}
```
webmaster128, line 40: In the rest of the cosmwasm codebase, we use `PascalCase` for enum cases (Yes, No, …)

webmaster128 (commented):
Some initial thoughs. But looks good overall

packages/cw3/README.md
```diff
@@ -0,0 +1,107 @@
...
+an auto-generated ID in the `Data` field (and the logs) that
+can be used to reference the proposal later. You can specify
+an expiration time/height in Propose, but this may be set 
+automatically by the contract (overriding or just enforcing
+min/max/default values).
+ 
+Many implementations will want to restrict who can propose.
+Maybe only people in the voting set. Maybe there is some
+deposit to be made along with the proposal. This is not
+in the spec but left open to the implementation.
```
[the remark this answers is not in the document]
ethanfrey: Should I specify that Propose does not include a yes vote from the sender if applicable?
Specify that it does only if the proposer is in the valid voter set?
Or leave it unspecified?

I'd like to specify this, as then wallets know if they need to propose and vote, or just propose.

packages/cw3/examples/schema.rs
ethanfrey: These work for generic queriers that don't support custom messages, and are actually the basis of most generic clients.

I would expect some custom contracts to add more and then we have custom clients that know that, but those would be chain-specific naturally, and this is the only set of schemas that would work over all contracts and all chains.

packages/cw3/src/helpers.rs
ethanfrey: I will leave a TODO here to make it more generic and fix it when using it.
That will only happen with an on-chain cw3 consumer.

packages/cw3/README.md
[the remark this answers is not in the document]
webmaster128: I think it is clean to make propose and vote two messages. It would be nice if they could be sent in one tranaction, but then we can't rely on execution data. However, this would work with client-side generated IDs.

packages/cw3/src/helpers.rs
ethanfrey: There is a comment to that effect on line 36 ( ` /// helper doesn't support custom messages now` )

packages/cw3/src/msg.rs
ethanfrey: Will update. Thanks

packages/cw3/src/helpers.rs
ethanfrey: I added a FIXME for this

packages/cw3/README.md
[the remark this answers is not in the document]
ethanfrey: I defined it as the first in the README

PR #91, opened 2020-09-15
Cw3 fixed multisig

### LINKED ISSUE #8: Basic multisig contract
Builds on #79 

Define a basic, usable multisig

- [x] Define fixed participant set in the beginning
- [x] Each participant may have a different weight
- [x] "Pass" threshold defined on init
- [x] Proposal expiration defined on init
- [x] Any member may create a proposal (`&[CosmosMsg]`) with an expiration
- [x] Any member may vote *once* for a proposal.
- [x] If enough yes votes are applied, any member may *execute* the proposal
- [x] Implement all queries
- [ ] Unit test coverage

In addition, specify a standard interface for the multisig, which can be used by more complex implementations.

ethanfrey:
Closes #8 

Status:

* Init completed
* Handle logic done (minus events/data on result)
* Query logic done

TODO:

- [x] Write unit tests
- [x] Return info on HandleResult (data, logs)
- [x] Write proper README


ethanfrey:
@maurolacy this is a super rough draft, but if you want to take a quick look and let me know if I forgot any error cases or such, please do so.

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -0,0 +1,267 @@
...
+        description,
+        expires,
+        msgs,
+        status: Status::Open,
+        yes_weight: weight,
+        required_weight: cfg.required_weight,
+    };
+
+    // get next id
+    let id = next_id(&mut deps.storage)?;
```
maurolacy, line 93: I guess concurrency checks / contention is handled at the framework / blockchain level.

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -0,0 +1,267 @@
...
+pub fn handle_execute<S: Storage, A: Api, Q: Querier>(
+    deps: &mut Extern<S, A, Q>,
+    env: Env,
+    proposal_id: u64,
+) -> StdResult<HandleResponse<Empty>> {
+    // anyone can trigger this if the vote passed
+
+    let mut prop = proposal_read(&deps.storage).load(&proposal_id.to_be_bytes())?;
+    if prop.status != Status::Open {
+        return Err(StdError::generic_err("Proposal is not open for voting"));
```
maurolacy, line 155: "Proposal is not open for voting" -> "Proposal is not open for execution" /  "Proposal is not open"?

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -0,0 +1,267 @@
...
+    }
+    // we enforce this for now, but maybe it doesn't make sense... since no one can *vote*
+    // after expiration, if it is expired but there were enough yes votes, it means it passed
+    // in the period. Do we enforce execution then as well?
```
maurolacy, line 159: > Do we enforce execution then as well?

I think so.

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -0,0 +1,267 @@
...
+    proposal_id: u64,
+) -> StdResult<HandleResponse<Empty>> {
+    // anyone can trigger this if the vote passed
+
+    let mut prop = proposal_read(&deps.storage).load(&proposal_id.to_be_bytes())?;
+    if [Status::Executed, Status::Failed]
+        .iter()
+        .any(|x| *x == prop.status)
+    {
+        return Err(StdError::generic_err("Cannot closed completed proposals"));
```
maurolacy, line 196: closed -> close.

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -0,0 +1,267 @@
...
+    }
+    if !prop.expires.is_expired(&env.block) {
+        return Err(StdError::generic_err(
+            "Proposal must expire before you can close it",
+        ));
+    }
+    // ensure it did not pass (think about the above... passed and expired should be EITHER closeable or executable)
+    if prop.yes_weight >= prop.required_weight {
+        return Err(StdError::generic_err("Already passed, try to execute it"));
```
maurolacy, line 205: This is consistent with allowing execution of (passed) expired proposals.

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    // set it to failed
+    prop.status = Status::Failed;
```
maurolacy, line 209: This is a personal preference, but I don't like the name `Failed`, as it makes me think there was an error or something. Maybe `DidntPass`, or `Rejected`?

Also, `Executed` can be maybe be renamed to `Passed`? Though in that case I like the naming more, because it indicates the associated messages have been dispatched.

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -0,0 +1,267 @@
...
+) -> StdResult<HandleResponse<Empty>> {
+    // only members of the multisig can create a proposal
+    let raw_sender = deps.api.canonical_address(&env.message.sender)?;
+    let weight = voter_weight_read(&deps.storage)
+        .may_load(raw_sender.as_slice())?
+        .ok_or_else(StdError::unauthorized)?;
+
+    // TODO: using max as default here, also enforce max
+    let cfg = config_read(&deps.storage).load()?;
+    let expires = latest.unwrap_or(cfg.max_voting_period);
```
maurolacy, line 79: No need to enforce that `latest` (if it is set) is less than or equal than `cfg.max_voting_period`?

maurolacy: Also, if and when `earliest` is not ignored, some consistency checks (`earliest` < `latest` <= `max_voting_period`) may be a good idea.

maurolacy (commented):
Logic looks good to me.

ethanfrey:
Thanks for the feedback. I will update the spec in some places, code in others

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -0,0 +1,267 @@
+use cosmwasm_std::{
+    to_binary, Api, Binary, CosmosMsg, Empty, Env, Extern, HandleResponse, HumanAddr, InitResponse,
+    Querier, StdError, StdResult, Storage,
+};
+use cw2::set_contract_version;
+
+use crate::msg::{HandleMsg, InitMsg, QueryMsg};
```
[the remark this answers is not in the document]
ethanfrey: Yes, those are good for clients. We may want to add additional methods there. I guess I could alias now, but I wanted to make it easier to add more (like cw20-base)

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -0,0 +1,267 @@
...
+
+pub fn init<S: Storage, A: Api, Q: Querier>(
+    deps: &mut Extern<S, A, Q>,
+    _env: Env,
+    msg: InitMsg,
+) -> StdResult<InitResponse> {
+    set_contract_version(&mut deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+    let total_weight = msg.voters.iter().map(|v| v.weight).sum();
+    let cfg = Config {
+        required_weight: msg.required_weight,
```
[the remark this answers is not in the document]
ethanfrey: Yup, this will be the simple contract. I plan more complex ones - dynamic, different thresholds, etc. This should be simple and solid.

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -0,0 +1,267 @@
...
+        total_weight,
+        max_voting_period: msg.max_voting_period,
+    };
+    config(&mut deps.storage).save(&cfg)?;
+
+    // add all voters
+    let mut voters = voter_weight(&mut deps.storage);
```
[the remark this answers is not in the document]
ethanfrey: I was ending up with lots of variables and methods named `voters` and didn't have a better name. I will update, thanks.

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -0,0 +1,267 @@
...
+
+    // get next id
+    let id = next_id(&mut deps.storage)?;
+
+    // save the proposal
+    proposal(&mut deps.storage).save(&id.to_be_bytes(), &prop)?;
+
+    // add the first yes vote from voter
+    let ballot = Ballot {
+        weight,
```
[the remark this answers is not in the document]
ethanfrey: To display it a bit nicer in queries. But yeah, it is a dup.

contracts/cw3-fixed-multisig/src/contract.rs
ethanfrey: Yeah, we are gauranteed single-threaded no race conditions.

contracts/cw3-fixed-multisig/src/contract.rs
ethanfrey: I renamed them. Also added a pass status (for passed, but not executed). We currently do not display these all, but I will try to improve the info in query to do better.

contracts/cw3-fixed-multisig/src/contract.rs
ethanfrey: Definitely, that is the meaning of the TODO - to set a max, and I abused it as default.

contracts/cw3-fixed-multisig/src/contract.rs
ethanfrey: Updated in 025e717

ethanfrey:
@maurolacy I think I finished all logic and updated it with your feedback

When you have time, can you start writing some unit tests (also for the logic I added in cw0)? Just make a branch and push the commits when you finish them. I will merge them in and add some more tests myself possibly. Thanks

contracts/cw3-fixed-multisig/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[features]
+backtraces = ["cosmwasm-std/backtraces"]
+# use library feature to disable all init/handle/query exports
+library = []
+
+[dependencies]
+cosmwasm-std = { version = "0.10.1", features = ["iterator"] }
```
orkunkl, line 21: Added task migration to issue https://github.com/CosmWasm/cosmwasm-plus/issues/96

contracts/cw3-fixed-multisig/src/state.rs
```diff
@@ -0,0 +1,111 @@
...
+
+pub fn proposal<S: Storage>(storage: &mut S) -> Bucket<S, Proposal> {
+    bucket(PREFIX_PROPOSAL, storage)
+}
+
+pub fn proposal_read<S: ReadonlyStorage>(storage: &S) -> ReadonlyBucket<S, Proposal> {
+    bucket_read(PREFIX_PROPOSAL, storage)
+}
+
+pub fn next_id<S: Storage>(storage: &mut S) -> StdResult<u64> {
```
orkunkl, line 86: [storage/sequence](https://github.com/CosmWasm/cosmwasm/blob/master/packages/storage/src/sequence.rs) does exactly this logic.

contracts/cw3-fixed-multisig/Cargo.toml
maurolacy: Hi, better to do that in another MR in my opinion.

contracts/cw3-fixed-multisig/src/state.rs
ethanfrey: I didn't know that. And I wrote it :facepalm:

contracts/cw3-fixed-multisig/Cargo.toml
ethanfrey: Yes, let us merge this first (and the nft contract) before migrating to 0.11.

For contract obligations, I need https://github.com/CosmWasm/cosmwasm-plus/pull/97 and would like this PR to be merged (0.10.x compatible) before end of month, and published as 0.2.2. Once those are merged, let's get the 0.11 branch going and prepare a 0.3.0 release.

ethanfrey:
I just gotta fix up the readme, then this is ready for merge. Last reviews welcome

Issue #88, opened 2020-09-12
Unify handling of native and cw20 coins in contracts


maurolacy:
Refactor cw20-enabled contracts to use the same pattern introduced in cw0-atomic-swap, of using the same function to handle both native and cw20 coins.

The main candidate for implementing this is cw20-escrow.

Also, document the pattern in the cosmwasm / cosmwasm-plus wiki docs.

PR #92, opened 2020-09-18
cw20-escrow refactoring: Unify handling of native and cw20

maurolacy:
A first iteration on #88.

This can be unified / simplified further, by example by introducing a `Cw20Balance` type. Or by introducing a generic `Coin`, that is a wrapper around `NativeCoin` and `Cw20Coin`. But, better leave those for another task / PR.

I'm not so happy with some of the trade-offs here. Like using a tuple type for the escrow balance. I also couldn't avoid code repetition in some parts, due to the structural differences between a `Coin` and a `Cw20Coin`. Maybe wrapping everything into a generic `Coin` is a good idea?


contracts/cw20-escrow/src/state.rs
```diff
@@ -8,7 +8,7 @@ use cosmwasm_storage::{bucket, bucket_read, prefixed_read, Bucket, ReadonlyBucke
 
 use cw20::Cw20Coin;
 
-pub(crate) type EscrowBalance = (Vec<Coin>, Vec<Cw20Coin>);
+pub(crate) type GenericBalance = (Vec<Coin>, Vec<Cw20Coin>);
```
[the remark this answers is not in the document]
ethanfrey: I also favor a struct here:

```rust
pub struct GenericBalance {
  pub native: Vec<Coin>,
  pub cw20: Vec<Cw20Coin>,
}
```

We end up with `escrow.balance.native`, which is much more clear than `escrow.balance.0`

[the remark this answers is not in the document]
ethanfrey: And as I commented above, I think a fair bit of the logic can be added as methods to this struct rather than external functions. We may well want to pull this struct out later for use in other types that store lots of coins, so adding coins to it removing coins from it, and converting to `Vec<CosmosMsg>` would be nice to have as methods. Along with various constructor helpers.

ethanfrey (commented):
Nice integration and ideas.

Added some comments to clean it up a bit more.

contracts/cw20-escrow/src/state.rs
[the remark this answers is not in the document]
maurolacy: Good ideas. Thanks.

contracts/cw20-atomic-swap/src/contract.rs
```diff
@@ -37,7 +36,7 @@ pub fn handle<S: Storage, A: Api, Q: Querier>(
     match msg {
         HandleMsg::Create(msg) => {
             let sent_funds = env.message.sent_funds.clone();
-            try_create(deps, env, msg, Balance::Native(NativeBalance(sent_funds)))
+            try_create(deps, env, msg, Balance::from(sent_funds))
```
ethanfrey, line 39: Nitpick:

`Balance::from(sent_funds)` == `sent_fund.into()` right?

I prefer the second, even if it is a bit more magic. I will merge without, just a note for the future.

contracts/cw20-atomic-swap/src/contract.rs
```diff
@@ -198,11 +197,11 @@ fn send_tokens<A: Api>(
         Ok(vec![])
     } else {
         match amount {
-            Balance::Native(NativeBalance(coins)) => {
+            Balance::Native(coins) => {
```
ethanfrey, line 200: Reads cleaner

contracts/cw20-escrow/src/contract.rs
```diff
@@ -37,15 +36,11 @@ pub fn handle<S: Storage, A: Api, Q: Querier>(
...
+            Balance::from(env.message.sent_funds),
             &env.message.sender,
         ),
         HandleMsg::Approve { id } => try_approve(deps, env, id),
-        HandleMsg::TopUp { id } => try_top_up(
-            deps,
-            id,
-            Balance::Native(NativeBalance(env.message.sent_funds)),
-        ),
+        HandleMsg::TopUp { id } => try_top_up(deps, id, Balance::from(env.message.sent_funds)),
```
ethanfrey, line 43: Also cleaner, and with `.into()`, even more concise

ethanfrey (approved):
Looks good, merging.

Left some comment on a few style nitpicks, things for the future

contracts/cw20-atomic-swap/src/contract.rs
maurolacy: Sure, will take those into account; along with implementing and using the `From` trait from the beginning.

PR #93, opened 2020-09-19
add_tokens() uses generic coin trait

maurolacy:
Simplifies `add_tokens()` by using `GenericCoin`.

Requires [#533](https://github.com/CosmWasm/cosmwasm/pull/533) in cosmwasm-plus.


ethanfrey (changes requested):
While I see you desire to simplify one function, this adds quite a bit of cognitive complexity to the whole system. Coins and Cw20Coins should be easy to pick up for new coders (as the basis of much other contracts). I would not go this approach.

ethanfrey (changes requested):
After looking at https://github.com/CosmWasm/cosmwasm-plus/pull/92 and https://github.com/CosmWasm/cosmwasm-plus/pull/93 I am pretty convinced that making this Trait does not simplify the system, but rather add systemtic complexity in order to simplify one function and thus adds more work to the end user.

I would not pursue this path, but thank you for the idea.

maurolacy:
> After looking at #92 and #93 I am pretty convinced that making this Trait does not simplify the system, but rather add systemtic complexity in order to simplify one function and thus adds more work to the end user.
> 
> I would not pursue this path, but thank you for the idea.

Yes, this was just an idea, and I wanted to hear your opinion about it. That's why I put it in a different PR. I'm OK with not doing it, and I agree it adds unnecessary complexity.

PR #94, opened 2020-09-20
Add zero amount checks / tests

### LINKED ISSUE #89: Avoid sending zero amount cw20 tokens
Add checks to cw20-base so that we don't allow sending a 0 value.

maurolacy:
Add zero amount checks to `cw20-base` handles.

Closes #89.


ethanfrey (approved):
Nice one

PR #95, opened 2020-09-24
Unit tests for cw3-fixed-multisig

maurolacy:
There are still some extra checks to do, but I think these now are acceptable unit tests.

### DIFF
--- a/contracts/cw3-fixed-multisig/src/contract.rs
+++ b/contracts/cw3-fixed-multisig/src/contract.rs
@@ -401,46 +401,56 @@ fn calc_range_start(start_after: Option<HumanAddr>) -> Option<Vec<u8>> {
 
 #[cfg(test)]
 mod tests {
+    use cosmwasm_std::from_binary;
     use cosmwasm_std::testing::{mock_dependencies, mock_env};
+
     use cw0::Duration;
     use cw2::{get_contract_version, ContractVersion};
 
     use crate::msg::Voter;
 
     use super::*;
-    use cosmwasm_std::from_binary;
+
+    fn voter<T: Into<HumanAddr>>(addr: T, weight: u64) -> Voter {
+        Voter {
+            addr: addr.into(),
+            weight,
+        }
+    }
 
     // this will set up the init for other tests
     fn setup_test_case<S: Storage, A: Api, Q: Querier>(
-        deps: &mut Extern<S, A, Q>,
+        mut deps: &mut Extern<S, A, Q>,
         env: Env,
-        voters: Vec<Voter>,
         required_weight: u64,
         max_voting_period: Duration,
     ) {
-        // Init a contract with admins
+        // Init a contract with voters
+        let voters = vec![
+            voter(env.message.sender.clone(), 1),
+            voter("voter0001", 1),
+            voter("voter0002", 2),
+            voter("voter0003", 3),
+        ];
+
         let init_msg = InitMsg {
             voters,
             required_weight,
             max_voting_period,
         };
-        init(deps, env, init_msg).unwrap();
+        init(&mut deps, env, init_msg).unwrap();
     }
 
     #[test]
     fn test_init_works() {
         let mut deps = mock_dependencies(20, &[]);
-
         let owner = HumanAddr::from("admin0001");
-        let voters = vec![Voter {
-            addr: HumanAddr::from("voter0001"),
-            weight: 1,
-        }];
+        let env = mock_env(owner, &[]);
+
         let required_weight = 1;
         let voting_period = Duration::Time(1234567);
 
-        let env = mock_env(owner, &[]);
-        setup_test_case(&mut deps, env, voters, required_weight, voting_period);
+        setup_test_case(&mut deps, env, required_weight, voting_period);
 
         assert_eq!(
             ContractVersion {
@@ -457,36 +467,12 @@ mod tests {
     fn test_propose_works() {
         let mut deps = mock_dependencies(20, &[]);
 
-        let owner = HumanAddr::from("admin0001");
-        let voters = vec![
-            Voter {
-                addr: owner.clone(),
-                weight: 0,
-            },
-            Voter {
-                addr: HumanAddr::from("voter0001"),
-                weight: 1,
-            },
-            Voter {
-                addr: HumanAddr::from("voter0002"),
-                weight: 2,
-            },
-            Voter {
-                addr: HumanAddr::from("voter0003"),
-                weight: 3,
-            },
-        ];
         let required_weight = 3;
         let voting_period = Duration::Time(2000000);
 
+        let owner = HumanAddr::from("admin0001");
         let env = mock_env(owner, &[]);
-        setup_test_case(
-            &mut deps,
-            env.clone(),
-            voters,
-            required_weight,
-            voting_period,
-        );
+        setup_test_case(&mut deps, env.clone(), required_weight, voting_period);
 
         // Propose
         let msgs = vec![CosmosMsg::Custom(Empty {})];
@@ -507,36 +493,12 @@ mod tests {
         let mut deps = mock_dependencies(20, &[]);
 
         let owner = HumanAddr::from("admin0001");
-        let voter1 = HumanAddr::from("voter0001");
-        let voters = vec![
-            Voter {
-                addr: owner.clone(),
-                weight: 0,
-            },
-            Voter {
-                addr: voter1.clone(),
-                weight: 1,
-            },
-            Voter {
-                addr: HumanAddr::from("voter0002"),
-                weight: 2,
-            },
-            Voter {
-                addr: HumanAddr::from("voter0003"),
-                weight: 3,
-            },
-        ];
+
         let required_weight = 3;
         let voting_period = Duration::Time(2000000);
 
         let env = mock_env(owner, &[]);
-        setup_test_case(
-            &mut deps,
-            env.clone(),
-            voters,
-            required_weight,
-            voting_period,
-        );
+        setup_test_case(&mut deps, env.clone(), required_weight, voting_period);
 
         // Propose
         let msgs = vec![CosmosMsg::Custom(Empty {})];
@@ -559,7 +521,7 @@ mod tests {
         assert_eq!(1, proposals.len());
         let proposal = &proposals[0];
 
-        // owner cannot Vote (again)
+        // owner cannot vote (again)
         let vote = HandleMsg::Vote {
             proposal_id: proposal.id,
             vote: Vote::Yes,
@@ -574,7 +536,7 @@ mod tests {
         }
 
         // But voter1 can
-        let env = mock_env(voter1, &[]);
+        let env = mock_env("voter0001", &[]);
         let vote = HandleMsg::Vote {
             proposal_id: proposal.id,
             vote: Vote::Yes,
--- a/contracts/cw3-fixed-multisig/src/contract.rs
+++ b/contracts/cw3-fixed-multisig/src/contract.rs
@@ -104,13 +104,19 @@ pub fn handle_propose<S: Storage, A: Api, Q: Querier>(
         return Err(StdError::generic_err("Wrong expiration option"));
     }
 
+    let status = if vote_power < cfg.required_weight {
+        Status::Open
+    } else {
+        Status::Passed
+    };
+
     // create a proposal
     let prop = Proposal {
         title,
         description,
         expires,
         msgs,
-        status: Status::Open,
+        status,
         yes_weight: vote_power,
         required_weight: cfg.required_weight,
     };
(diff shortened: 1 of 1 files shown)


contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -399,5 +401,826 @@ fn calc_range_start(start_after: Option<HumanAddr>) -> Option<Vec<u8>> {
...
+        // But voter1 can
+        let env = mock_env(voter1, &[]);
+        let vote = HandleMsg::Vote {
+            proposal_id: proposal.id,
+            vote: Vote::Yes,
+        };
+        let res = handle(&mut deps, env, vote).unwrap();
+
+        // Verify
+        assert_eq!(res, HandleResponse::default());
```
ethanfrey, line 585: And then ensure the vote was updated. Also, turned to passed with enough votes.

A test case that it immediately goes to Passed if the proposed has the required weight alone would be good as well (I think this doesn't work now and please improve the code as well)

ethanfrey (commented):
Thank you for the start. Added some comments on how to simplify the test code, but also some points that need testing and likely don't work properly in the contract code, which I didn't see until reading these test cases. Thank you for them

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -399,5 +401,826 @@ fn calc_range_start(start_after: Option<HumanAddr>) -> Option<Vec<u8>> {
...
+    use cw0::Duration;
+    use cw2::{get_contract_version, ContractVersion};
+
+    use crate::msg::Voter;
+
+    use super::*;
+    use cosmwasm_std::from_binary;
+
+    // this will set up the init for other tests
+    fn setup_test_case<S: Storage, A: Api, Q: Querier>(
```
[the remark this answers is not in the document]
maurolacy: Yes, I started with it, and now I see it's not very useful. Will refactor, probably with no args, or just a couple args for flexibility.

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -399,5 +401,826 @@ fn calc_range_start(start_after: Option<HumanAddr>) -> Option<Vec<u8>> {
...
+
+    // TODO: query() tests
+
+    #[test]
+    fn test_propose_works() {
+        let mut deps = mock_dependencies(20, &[]);
+
+        let owner = HumanAddr::from("admin0001");
+        let voters = vec![
+            Voter {
```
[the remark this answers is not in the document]
maurolacy: OK.

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -399,5 +401,826 @@ fn calc_range_start(start_after: Option<HumanAddr>) -> Option<Vec<u8>> {
...
+        setup_test_case(
+            &mut deps,
+            env.clone(),
+            voters,
+            required_weight,
+            voting_period,
+        );
+
+        // Propose
+        let msgs = vec![CosmosMsg::Custom(Empty {})];
```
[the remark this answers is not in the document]
maurolacy: All right, will change it.

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -399,5 +401,826 @@ fn calc_range_start(start_after: Option<HumanAddr>) -> Option<Vec<u8>> {
...
+
+        let env = mock_env(owner, &[]);
+        setup_test_case(&mut deps, env, voters, required_weight, voting_period);
+
+        assert_eq!(
+            ContractVersion {
+                contract: CONTRACT_NAME.to_string(),
+                version: CONTRACT_VERSION.to_string(),
+            },
+            get_contract_version(&deps.storage).unwrap()
```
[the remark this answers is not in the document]
maurolacy: Yes. My plan was to write a skeleton for all the tests first, and then I'll iterate adding more checks and cases to cover all the variants / errors in the main code.

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -399,5 +401,826 @@ fn calc_range_start(start_after: Option<HumanAddr>) -> Option<Vec<u8>> {
...
+
+    #[test]
+    fn test_propose_works() {
+        let mut deps = mock_dependencies(20, &[]);
+
+        let owner = HumanAddr::from("admin0001");
+        let voters = vec![
+            Voter {
+                addr: owner.clone(),
+                weight: 0,
```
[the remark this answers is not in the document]
maurolacy: All right. I thought zero weight was a good idea for a proposal creator that don't effectively votes on it. (He votes it, but without changing the tally).

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -399,5 +401,826 @@ fn calc_range_start(start_after: Option<HumanAddr>) -> Option<Vec<u8>> {
...
+        let proposal = HandleMsg::Propose {
+            title: "Title".to_string(),
+            description: "Description".to_string(),
+            msgs,
+            latest: None,
+        };
+        handle(&mut deps, env.clone(), proposal).unwrap();
+
+        // Query the proposal id
+        let query_msg = QueryMsg::ListProposals {
```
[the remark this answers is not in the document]
maurolacy: Sure.

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -399,5 +401,826 @@ fn calc_range_start(start_after: Option<HumanAddr>) -> Option<Vec<u8>> {
...
+            start_after: None,
+            limit: None,
+        };
+        let proposals: ProposalListResponse =
+            from_binary(&query(&deps, query_msg).unwrap()).unwrap();
+        let proposals = &proposals.proposals;
+        assert_eq!(1, proposals.len());
+        let proposal = &proposals[0];
```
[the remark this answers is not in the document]
maurolacy: I'll add those later, yes.

contracts/cw3-fixed-multisig/src/contract.rs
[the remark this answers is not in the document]
ethanfrey: Oh, that is an interesting idea. A way to give someone proposer but not voting rights.

Okay, then let's leave that valid. Empty voters vec (or total_weight == 0) should be invalid, as should required_weight > total_weight.

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -401,46 +401,56 @@ fn calc_range_start(start_after: Option<HumanAddr>) -> Option<Vec<u8>> {
...
+        mut deps: &mut Extern<S, A, Q>,
         env: Env,
-        voters: Vec<Voter>,
         required_weight: u64,
         max_voting_period: Duration,
     ) {
-        // Init a contract with admins
+        // Init a contract with voters
+        let voters = vec![
+            voter(env.message.sender.clone(), 1),
```
ethanfrey, line 430: Minor point, I think you can use `&env.message.sender` here which is *slightly* less verbose. Both are equivalent though

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -574,7 +536,7 @@ mod tests {
         }
 
         // But voter1 can
-        let env = mock_env(voter1, &[]);
+        let env = mock_env("voter0001", &[]);
```
ethanfrey, line 539: Good to define these strings as constants in `mod test`, and then use those same constants in the tests as well as in the setup.

ethanfrey:
Nice cleanup, and you convinved me about the 0 weight members. This should be documented in the README (TODO)

contracts/cw3-fixed-multisig/src/contract.rs
maurolacy: OK. Good idea.

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -133,8 +133,15 @@ pub fn handle_propose<S: Storage, A: Api, Q: Querier>(
     };
     ballots(&mut deps.storage, id).save(raw_sender.as_slice(), &ballot)?;
 
-    // TODO: add some event attributes
-    Ok(HandleResponse::default())
+    Ok(HandleResponse {
```
ethanfrey, line 140: Thanks for adding these

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -1,32 +1,61 @@
...
+        return Err(StdError::generic_err("Required weight cannot be zero"));
+    }
+    if msg.voters.is_empty() {
+        return Err(StdError::generic_err("No voters"));
+    }
+    let weights: StdResult<Vec<u64>> = msg
+        .voters
+        .iter()
+        .map(|v| {
+            if v.weight == 0 && v.addr != env.message.sender {
```
ethanfrey, line 40: After your argument why 0 could make sense, I am happy for any of these to be 0, as long as the other checks (above and below) pass.

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -77,11 +106,15 @@ pub fn handle_propose<S: Storage, A: Api, Q: Querier>(
...
+
     // max expires also used as default
     let max_expires = cfg.max_voting_period.after(&env.block);
     let mut expires = latest.unwrap_or(max_expires);
-    if expires.partial_cmp(&max_expires) != Some(Ordering::Less) {
+    let comp = expires.partial_cmp(&max_expires);
+    if let Some(Ordering::Greater) = comp {
         expires = max_expires;
+    } else if comp.is_none() {
+        return Err(StdError::generic_err("Wrong expiration option"));
```
ethanfrey, line 117: Better check - nice eye

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -399,5 +457,485 @@ fn calc_range_start(start_after: Option<HumanAddr>) -> Option<Vec<u8>> {
...
+        let res = init(&mut deps, env.clone(), init_msg);
+
+        // Verify
+        assert!(res.is_err());
+        match res.unwrap_err() {
+            StdError::GenericErr { msg, .. } => assert_eq!(&msg, "Required weight cannot be zero"),
+            e => panic!("unexpected error: {}", e),
+        }
+
+        // Zero weights for voters other than sender not allowed
```
ethanfrey, line 549: As above, I would actually allow this case - you convinced me with the "non-voting proposer" concept.

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -399,5 +457,485 @@ fn calc_range_start(start_after: Option<HumanAddr>) -> Option<Vec<u8>> {
...
+        // Verify
+        assert_eq!(
+            res,
+            HandleResponse {
+                messages: vec![],
+                log: vec![
+                    log("action", "vote"),
+                    log("sender", VOTER3),
+                    log("proposal_id", proposal_id),
+                    log("status", "Passed"),
```
ethanfrey, line 823: This is very nice addition to the log messages

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -399,5 +457,485 @@ fn calc_range_start(start_after: Option<HumanAddr>) -> Option<Vec<u8>> {
...
+                    log("action", "vote"),
+                    log("sender", VOTER1),
+                    log("proposal_id", proposal_id),
+                    log("status", "Open"),
+                ],
+                data: None
+            }
+        );
+
+        // TODO: No/Veto votes have no effect on the tally
```
ethanfrey, line 737: The only thing I would add here are the voting cases you already marked with TODO.

Other than that, looking great and happy to merge when the last points are finished

ethanfrey (approved):
Looking good, just those last TODOs and it is golden.

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -809,7 +811,17 @@ mod tests {
             }
         );
 
-        // TODO: non-Open proposals cannot be voted
+        // non-Open proposals cannot be voted
+        // Vote it again
+        let env = mock_env(VOTER5, &[]);
+        let res = handle(&mut deps, env, yes_vote);
```
maurolacy, line 817: After adding this test, I was left wondering that maybe is a good idea to allow voting on `Passed` proposals (L165](https://github.com/CosmWasm/cosmwasm-plus/blob/f979892bb7d32339769ad83ccb289e92e5a2f300/contracts/cw3-fixed-multisig/src/contract.rs#L165) above).

That way, we can get better statistics on the final result of a given proposal. That is, despite a proposal having passed, it can have a large number of No / Veto votes (some of which can be cast after it passed). Or, conversely, a proposal can have a large adoption, than the one reflected by the results, which currently "freeze" after it passed.


That way you, can get better statistics

ethanfrey:
Thank you

PR #97, opened 2020-09-25
Implement cw721-base nft contract

### LINKED ISSUE #44: Implement base NFT contract
Builds on #43 

Implement a simple contract that implements the NFT spec (like cw20-base does). This is means to be used stand-alone or further customized.

Not not implement enumeration, or at least enumeration by owner, as that requires secondary indexes to do cleanly: https://github.com/CosmWasm/cosmwasm/issues/468

ethanfrey:
Closes #44

Most logic writen.

TODO:
- [x] Send proper message on `SendNftToken`
- [x] Add unit tests
- [x] Write a better README


ethanfrey:
@maurolacy This is roughly done for the main code path, only basic unit tests.
It also could use a good review to point out any issues in the design.

maurolacy:
OK, I'll review it tomorrow morning.

contracts/cw721-base/src/contract.rs
```diff
@@ -0,0 +1,499 @@
...
+}
+
+pub fn handle_send_nft<S: Storage, A: Api, Q: Querier>(
+    deps: &mut Extern<S, A, Q>,
+    env: Env,
+    contract: HumanAddr,
+    token_id: String,
+    _msg: Option<Binary>,
+) -> StdResult<HandleResponse> {
+    let mut token = tokens(&mut deps.storage).load(token_id.as_bytes())?;
```
maurolacy, line 130: This is just `handle_transfer_nft()` (plus the send msg), and can probably be simplified as such.

contracts/cw721-base/src/contract.rs
```diff
@@ -0,0 +1,499 @@
...
+    Ok(HandleResponse::default())
+}
+
+pub fn handle_revoke<S: Storage, A: Api, Q: Querier>(
+    deps: &mut Extern<S, A, Q>,
+    env: Env,
+    spender: HumanAddr,
+    token_id: String,
+) -> StdResult<HandleResponse> {
+    let mut token = tokens(&mut deps.storage).load(token_id.as_bytes())?;
```
maurolacy, line 186: Not sure it's worth it, but this logic is shared between `approve` and `revoke`. In fact, `approve` is just `revoke` plus adding a new spender.

maurolacy (commented):
Looks pretty good to me. Besides what I think is a serious error (`token_id` overwrite in minter), logic looks good.

Will continue reviewing the tests later.

contracts/cw721-base/src/contract.rs
```diff
@@ -0,0 +1,499 @@
...
+
+    // create the token
+    let token = TokenInfo {
+        owner: deps.api.canonical_address(&owner)?,
+        approvals: vec![],
+        name,
+        description: description.unwrap_or_default(),
+        image,
+    };
+    tokens(&mut deps.storage).save(token_id.as_bytes(), &token)?;
```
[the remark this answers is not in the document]
ethanfrey: This is very true. Good catch and major bug

ethanfrey:
Addressed the major issues raised - good points, thank you.

maurolacy:
I've made some changes directly on your branch. Please review them.

contracts/cw721-base/src/contract.rs
```diff
@@ -397,7 +397,7 @@ fn humanize_approvals<A: Api>(api: A, info: &TokenInfo) -> StdResult<Vec<cw721::
 fn humanize_approval<A: Api>(api: A, approval: &Approval) -> StdResult<cw721::Approval> {
     Ok(cw721::Approval {
         spender: api.human_address(&approval.spender)?,
-        expires: approval.expires.clone(),
```
ethanfrey: Yup, this is a nice fix building on cw3 pr.

contracts/cw721-base/src/contract.rs
```diff
@@ -0,0 +1,499 @@
...
+const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+pub fn init<S: Storage, A: Api, Q: Querier>(
+    deps: &mut Extern<S, A, Q>,
+    _env: Env,
+    msg: InitMsg,
+) -> StdResult<InitResponse> {
+    set_contract_version(&mut deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+
+    let info = ContractInfoResponse {
```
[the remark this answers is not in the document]
ethanfrey: Yeah, I figure it is 1:1 what we return (and only for returning that info) no need to define a new type.

But in general, nice to keep state and api separate.

PR #99, opened 2020-09-29
cw721-nft-base unit tests

maurolacy:
Basic unit tests for most of the handlers (will write `approve_all()` / `revoke_all()` tests tomorrow).

Tests are not exhaustive, but are a good start. They can be made exhaustive with more time / effort.

### DIFF
--- a/contracts/cw721-base/src/contract.rs
+++ b/contracts/cw721-base/src/contract.rs
@@ -569,5 +569,66 @@ mod tests {
                 approvals: vec![]
             }
         );
+
+        // TODO: Cannot mint same token again
+    }
+
+    #[test]
+    fn transfering_nft() {
+        let mut deps = mock_dependencies(20, &[]);
+        setup_contract(&mut deps);
+
+        // Mint a token
+        let token_id = "melt".to_string();
+        let name = "Melting power".to_string();
+        let description = "Allows the owner to melt anyone looking at him or her".to_string();
+
+        let mint_msg = HandleMsg::Mint {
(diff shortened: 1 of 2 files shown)


ethanfrey:
Nice start. Thank you for this.

Please ping me on discord when you start coding tomorrow, or I will if I start first. If I start first, I would merge this (already a nice step forward) and then start another PR.

contracts/cw721-base/src/contract.rs
```diff
@@ -783,4 +805,103 @@ mod tests {
...
+    fn approving_all_revoking_all() {
+        let mut deps = mock_dependencies(20, &[]);
+        setup_contract(&mut deps);
+
+        // Mint a couple tokens (from the same owner)
+        let token_id1 = "grow1".to_string();
+        let name1 = "Growing power".to_string();
+        let description1 = "Allows the owner the power to grow anything".to_string();
+        let token_id2 = "grow2".to_string();
+        let name2 = "Growing power".to_string();
```
ethanfrey, line 819: This should be slightly different that name1. Even calling it "Growing Power 2" is fine.

contracts/cw721-base/src/contract.rs
```diff
@@ -783,4 +805,103 @@ mod tests {
...
+        let description2 = "Allows the owner the power to grow anything".to_string();
```
ethanfrey, line 820: Ditto

ethanfrey (commented):
Nice start minor comments - plus those around the commented out code on discord

PR #100, opened 2020-09-30
Fix CLI call command

webmaster128:
Before, the linked file did not point to a typescript file. Also the script would break once 0.23 is out.


ethanfrey (approved):
nice fix

Issue #96, opened 2020-09-24
Migrate contracts to v0.11


maurolacy:
- [ ] Migrate https://github.com/CosmWasm/cosmwasm-plus/pull/91 cw3-fixed-multisig

PR #104, opened 2020-10-02
Migrate to v0.11

maurolacy:
WIP for #96.

### DIFF
--- a/contracts/cw1-subkeys/src/contract.rs
+++ b/contracts/cw1-subkeys/src/contract.rs
@@ -172,7 +172,7 @@ where
         return Err(StdError::generic_err("Cannot set allowance to own account"));
     }
 
-    allowances(&mut deps.storage).update(spender_raw.as_slice(), |allow| {
+    allowances(&mut deps.storage).update::<_, StdError>(spender_raw.as_slice(), |allow| {
         let mut allowance = allow.unwrap_or_default();
         if let Some(exp) = expires {
             allowance.expires = exp;
@@ -216,16 +216,17 @@ where
         return Err(StdError::generic_err("Cannot set allowance to own account"));
     }
 
-    let allowance = allowances(&mut deps.storage).update(spender_raw.as_slice(), |allow| {
-        // Fail fast
-        let mut allowance =
-            allow.ok_or_else(|| StdError::not_found("No allowance for this account"))?;
-        if let Some(exp) = expires {
-            allowance.expires = exp;
-        }
-        allowance.balance = allowance.balance.sub_saturating(amount.clone())?; // Tolerates underflows (amount bigger than balance), but fails if there are no tokens at all for the denom (report potential errors)
-        Ok(allowance)
-    })?;
+    let allowance =
+        allowances(&mut deps.storage).update::<_, StdError>(spender_raw.as_slice(), |allow| {
+            // Fail fast
+            let mut allowance =
+                allow.ok_or_else(|| StdError::not_found("No allowance for this account"))?;
+            if let Some(exp) = expires {
+                allowance.expires = exp;
+            }
+            allowance.balance = allowance.balance.sub_saturating(amount.clone())?; // Tolerates underflows (amount bigger than balance), but fails if there are no tokens at all for the denom (report potential errors)
+            Ok(allowance)
+        })?;
     if allowance.balance.is_empty() {
         allowances(&mut deps.storage).remove(spender_raw.as_slice());
     }
--- a/contracts/cw1-subkeys/src/contract.rs
+++ b/contracts/cw1-subkeys/src/contract.rs
@@ -100,7 +100,7 @@ where
                 }) => {
                     let mut allowances = allowances(&mut deps.storage);
                     let allow = allowances.may_load(owner_raw.as_slice())?;
-                    let mut allowance = allow.ok_or_else(|| ContractError::NotFound {})?;
+                    let mut allowance = allow.ok_or_else(|| ContractError::NoAllowance {})?;
                     // Decrease allowance
                     allowance.balance = allowance.balance.sub(amount.clone())?;
                     allowances.save(owner_raw.as_slice(), &allowance)?;
@@ -218,7 +218,7 @@ where
         spender_raw.as_slice(),
         |allow| {
             // Fail fast
-            let mut allowance = allow.ok_or_else(|| ContractError::NotFound {})?;
+            let mut allowance = allow.ok_or_else(|| ContractError::NoAllowance {})?;
             if let Some(exp) = expires {
                 allowance.expires = exp;
             }
@@ -1167,7 +1167,7 @@ mod tests {
         let env = mock_env(&spender2, &[]);
         let res = handle(&mut deps, env, handle_msg.clone());
         match res.unwrap_err() {
-            ContractError::NotFound { .. } => {}
+            ContractError::NoAllowance { .. } => {}
             e => panic!("unexpected error: {}", e),
         }
 
--- a/contracts/cw1-subkeys/src/contract.rs
+++ b/contracts/cw1-subkeys/src/contract.rs
@@ -1,5 +1,6 @@
 use schemars::JsonSchema;
 use std::fmt;
+use std::ops::{AddAssign, Sub};
 
 use cosmwasm_std::{
     attr, to_binary, Api, BankMsg, Binary, CanonicalAddr, Coin, CosmosMsg, Empty, Env, Extern,
@@ -15,15 +16,14 @@ use cw1_whitelist::{
 };
 use cw2::set_contract_version;
 
+use crate::error::ContractError;
 use crate::msg::{
     AllAllowancesResponse, AllPermissionsResponse, AllowanceInfo, HandleMsg, PermissionsInfo,
     QueryMsg,
 };
 use crate::state::{
-    allowances, allowances_read, permissions, permissions_read, Allowance, PermissionErr,
-    Permissions,
+    allowances, allowances_read, permissions, permissions_read, Allowance, Permissions,
 };
-use std::ops::{AddAssign, Sub};
 
 // version info for migration info
 const CONTRACT_NAME: &str = "crates.io:cw1-subkeys";
@@ -45,11 +45,11 @@ pub fn handle<S: Storage, A: Api, Q: Querier>(
     // Note: implement this function with different type to add support for custom messages
     // and then import the rest of this contract code.
     msg: HandleMsg<Empty>,
-) -> StdResult<HandleResponse<Empty>> {
+) -> Result<HandleResponse<Empty>, ContractError> {
     match msg {
         HandleMsg::Execute { msgs } => handle_execute(deps, env, msgs),
-        HandleMsg::Freeze {} => handle_freeze(deps, env),
-        HandleMsg::UpdateAdmins { admins } => handle_update_admins(deps, env, admins),
+        HandleMsg::Freeze {} => Ok(handle_freeze(deps, env)?),
+        HandleMsg::UpdateAdmins { admins } => Ok(handle_update_admins(deps, env, admins)?),
         HandleMsg::IncreaseAllowance {
             spender,
             amount,
@@ -71,7 +71,7 @@ pub fn handle_execute<S: Storage, A: Api, Q: Querier, T>(
     deps: &mut Extern<S, A, Q>,
     env: Env,
     msgs: Vec<CosmosMsg<T>>,
-) -> StdResult<HandleResponse<T>>
+) -> Result<HandleResponse<T>, ContractError>
 where
     T: Clone + fmt::Debug + PartialEq + JsonSchema,
 {
@@ -89,8 +89,7 @@ where
                 CosmosMsg::Staking(staking_msg) => {
                     let permissions = permissions(&mut deps.storage);
                     let perm = permissions.may_load(owner_raw.as_slice())?;
-                    let perm =
-                        perm.ok_or_else(|| StdError::not_found("No permissions for this account"))?;
+                    let perm = perm.ok_or_else(|| ContractError::NotAllowed {})?;
 
                     check_staking_permissions(staking_msg, perm)?;
                 }
@@ -101,14 +100,13 @@ where
                 }) => {
                     let mut allowances = allowances(&mut deps.storage);
                     let allow = allowances.may_load(owner_raw.as_slice())?;
-                    let mut allowance = allow
-                        .ok_or_else(|| StdError::not_found("No allowance for this account"))?;
+                    let mut allowance = allow.ok_or_else(|| ContractError::NoAllowance {})?;
                     // Decrease allowance
                     allowance.balance = allowance.balance.sub(amount.clone())?;
                     allowances.save(owner_raw.as_slice(), &allowance)?;
                 }
                 _ => {
-                    return Err(StdError::generic_err("Message type rejected"));
+                    return Err(ContractError::MessageTypeRejected {});
                 }
             }
         }
@@ -125,26 +123,26 @@ where
 pub fn check_staking_permissions(
     staking_msg: &StakingMsg,
     permissions: Permissions,
-) -> Result<bool, PermissionErr> {
+) -> Result<bool, ContractError> {
     match staking_msg {
         StakingMsg::Delegate { .. } => {
             if !permissions.delegate {
-                return Err(PermissionErr::Delegate {});
+                return Err(ContractError::DelegatePerm {});
             }
         }
         StakingMsg::Undelegate { .. } => {
             if !permissions.undelegate {
-                return Err(PermissionErr::Undelegate {});
+                return Err(ContractError::UnDelegatePerm {});
             }
         }
         StakingMsg::Redelegate { .. } => {
             if !permissions.redelegate {
-                return Err(PermissionErr::Redelegate {});
+                return Err(ContractError::ReDelegatePerm {});
             }
         }
         StakingMsg::Withdraw { .. } => {
             if !permissions.withdraw {
-                return Err(PermissionErr::Withdraw {});
+                return Err(ContractError::WithdrawPerm {});
             }
         }
     }
@@ -157,7 +155,7 @@ pub fn handle_increase_allowance<S: Storage, A: Api, Q: Querier, T>(
     spender: HumanAddr,
     amount: Coin,
     expires: Option<Expiration>,
-) -> StdResult<HandleResponse<T>>
+) -> Result<HandleResponse<T>, ContractError>
 where
     T: Clone + fmt::Debug + PartialEq + JsonSchema,
 {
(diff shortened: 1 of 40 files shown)


contracts/cw1-subkeys/src/contract.rs
```diff
@@ -1187,14 +1189,14 @@ mod tests {
...
-            vec![log("action", "execute"), log("owner", spender1.clone())]
+            res.attributes,
+            vec![attr("action", "execute"), attr("owner", spender1.clone())]
         );
 
         // And then cannot (not enough funds anymore)
         let res = handle(&mut deps, env, handle_msg.clone());
         match res.unwrap_err() {
-            StdError::Underflow { .. } => {}
+            ContractError::Std(StdError::Underflow { .. }) => {}
```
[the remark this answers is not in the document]
ethanfrey: But since we are are creating custom errors everywhere, maybe we want to map this.
Either where it is generated like `.map_err(|e| ...)`. Or do something like:

```rust
impl From<cw1_whitelist::ContractError> for ContractError
    fn from(err: cw1_whitelist::ContractError) -> Self {
        match err {
            cw1_whitelist::ContractError::Std(error) => ContractError::Std(error),
            cw1_whitelist::ContractError::Unauthorized {} => ContractError::Unauthorized {},
        }
    }
}
```

But mapping `StdError::Underflow{}` => `ContractError::InsufficientAllowance{}`?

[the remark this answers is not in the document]
ethanfrey: Although the underflow message does show you the amounts, which is nice and shouldn't be lost.

This is an idea, take it or leave it.

contracts/cw1-subkeys/src/error.rs
```diff
@@ -0,0 +1,47 @@
+use cosmwasm_std::StdError;
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum ContractError {
```
ethanfrey, line 5: Nice explicit error cases. However, I didn't see them all checked in the test cases. I think there are some which would be even cleaner when they assert eg. `ContractError::DelegatePerm` to see exactly what failed.

contracts/cw1-subkeys/src/state.rs
```diff
@@ -33,38 +33,13 @@ const PREFIX_PERMISSIONS: &[u8] = b"permissions";
...
+    bucket(storage, PREFIX_PERMISSIONS)
 }
 
 /// returns a bucket with all permissionsk (query by subkey)
 /// (read-only version for queries)
 pub fn permissions_read<S: ReadonlyStorage>(storage: &S) -> ReadonlyBucket<S, Permissions> {
-    bucket_read(PREFIX_PERMISSIONS, storage)
-}
-
-pub enum PermissionErr {
```
ethanfrey, line 45: This is now merged into the contract error. Which is a nice improvement.

contracts/cw1-subkeys/src/state.rs
```diff
@@ -33,38 +33,13 @@ const PREFIX_PERMISSIONS: &[u8] = b"permissions";
 
 /// returns a bucket with all permissions (query by subkey)
 pub fn permissions<S: Storage>(storage: &mut S) -> Bucket<S, Permissions> {
-    bucket(PREFIX_PERMISSIONS, storage)
+    bucket(storage, PREFIX_PERMISSIONS)
```
ethanfrey, line 36: I'm about to change this all again :wink: 
https://github.com/CosmWasm/cosmwasm/pull/559

contracts/cw1-whitelist/src/error.rs
```diff
@@ -0,0 +1,11 @@
+use cosmwasm_std::StdError;
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum ContractError {
```
ethanfrey, line 5: If the only mapping is `StdError::Unauthorized` -> `ContractError::Unauthorized` and no custom cases, I would rather just use StdError here. No need to use custom messages if we don't need them. (It would also make inheritance in cw1-subkeys easier... we can use the `cw20-base` -> `cw20-staking` extension as a good example of complex error extension.

contracts/cw20-atomic-swap/src/error.rs
```diff
@@ -0,0 +1,26 @@
...
+    #[error("Parse error: {0}")]
+    ParseError(String),
+
+    #[error("Invalid {0}")]
+    Invalid(String),
+
+    #[error("{0}")]
+    EmptyBalance(String),
+
+    #[error("Atomic swap not yet expired")]
```
ethanfrey, line 18: Nice error cases.

contracts/cw20-atomic-swap/src/contract.rs
```diff
@@ -166,22 +167,22 @@ pub fn try_refund<S: Storage, A: Api, Q: Querier>(
...
             if bin.len() == 32 {
                 Ok(bin)
             } else {
-                Err(StdError::generic_err("Hash must be 64 characters"))
+                Err(ContractError::Invalid("hash: must be 64 characters".into()))
             }
         }
-        Err(e) => Err(StdError::generic_err(format!(
-            "Error parsing hash: {}",
+        Err(e) => Err(ContractError::ParseError(format!(
```
ethanfrey, line 184: You can also give some text to the error itself rather than combine messages here.

`ContractError::ParseHash(e.to_string())`

```
   #[error("non-hex hash: {0}")]
    ParseHashError(String),
```

I generally prefer avoiding the `format!()` directive when making errors. I used that a lot in generic error, but if we have custom types, let's worry about the display in `errors.rs` and just pass in needed context here (if any)

contracts/cw20-atomic-swap/src/contract.rs
```diff
@@ -166,22 +167,22 @@ pub fn try_refund<S: Storage, A: Api, Q: Querier>(
...
 
-fn parse_hex_32(data: &str) -> StdResult<Vec<u8>> {
+fn parse_hex_32(data: &str) -> Result<Vec<u8>, ContractError> {
     match hex::decode(data) {
         Ok(bin) => {
             if bin.len() == 32 {
                 Ok(bin)
             } else {
-                Err(StdError::generic_err("Hash must be 64 characters"))
+                Err(ContractError::Invalid("hash: must be 64 characters".into()))
```
ethanfrey, line 181: For example, here we could use

`ContractError::InvalidHashLength(data.len())` and add  `"hash must be 64 characters, you provided {0}"` inside the `#[error]` text

contracts/cw20-atomic-swap/src/contract.rs
```diff
@@ -362,7 +365,7 @@ mod tests {
         let res = handle(&mut deps, env, HandleMsg::Create(create.clone()));
         match res {
             Ok(_) => panic!("expected error"),
-            Err(StdError::GenericErr { msg, .. }) => {
+            Err(ContractError::EmptyBalance(msg)) => {
```
ethanfrey, line 368: For me the great thing about `ContractError` is not needing to check the message.
We write it declarative in one file and can view/update if we want to.

For tests, we can just assert the proper type (and maybe metadata like length), but no need to compare strings intended for human consumption in a test - just that the expected branch got triggered (eg. failed due to empty balance, not due to unauthorized)

contracts/cw20-escrow/src/contract.rs
```diff
@@ -192,17 +192,17 @@ pub fn try_refund<S: Storage, A: Api, Q: Querier>(
         Err(StdError::unauthorized())
     } else {
         // we delete the escrow (TODO: expose this in Bucket for simpler API)
-        prefixed(PREFIX_ESCROW, &mut deps.storage).remove(id.as_bytes());
+        prefixed(&mut deps.storage, PREFIX_ESCROW).remove(id.as_bytes());
```
ethanfrey, line 195: Note, this is now in `Bucket.remove` you can use that now.

contracts/cw20-escrow/src/contract.rs
```diff
@@ -161,17 +161,17 @@ pub fn try_approve<S: Storage, A: Api, Q: Querier>(
         Err(StdError::generic_err("escrow expired"))
     } else {
         // we delete the escrow (TODO: expose this in Bucket for simpler API)
-        prefixed(PREFIX_ESCROW, &mut deps.storage).remove(id.as_bytes());
+        prefixed(&mut deps.storage, PREFIX_ESCROW).remove(id.as_bytes());
```
ethanfrey, line 164: And here... time to update

ethanfrey (commented):
Nice work, and quite a bit of cleanup here.

I like your usage of very clear error cases in subkeys. It is quite good in atomic swap as well, but I think there are still too many user-facing strings being created in the code, not in the error type. 

I think it is best practice to only return *data related to the error* inside the `ContractError` type, and all formatting to make it reasonable (making use of the metadata) is done in the `#[error]` text. This also gives a very clear overview for anyone trying to adjust the messages (or trace a bug)

maurolacy:
> It is quite good in atomic swap as well, but I think there are still too many user-facing strings being created in the code, not in the error type.

You're right, and I see you point: The very idea of custom errors is to move the messages to the error definition; and in any case just pass parameters, and that only if justified.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -258,22 +259,20 @@ where
...
 
     if !cfg.is_admin(&owner_raw) {
-        return Err(StdError::unauthorized());
+        return Err(ContractError::Unauthorized {});
     }
     if spender_raw == owner_raw {
-        return Err(StdError::generic_err(
-            "Cannot set permission to own account",
-        ));
+        return Err(ContractError::CannotSetPermOwnAccount {});
```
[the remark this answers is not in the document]
maurolacy: They are similar, yes. Consolidated to same error.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -1178,7 +1180,7 @@ mod tests {
         let env = mock_env(&spender2, &[]);
         let res = handle(&mut deps, env, handle_msg.clone());
         match res.unwrap_err() {
-            StdError::NotFound { .. } => {}
+            ContractError::NoAllowance { .. } => {}
```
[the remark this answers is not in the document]
maurolacy: Oops, just changed it to `NotFound` (for similarity with other contracts). Reverting it for clarity.

contracts/cw1-subkeys/src/contract.rs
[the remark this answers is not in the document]
maurolacy: I would leave it as it is. Besides the extra boiler plate, it's pretty clear what's being done.

contracts/cw1-subkeys/src/error.rs
maurolacy: That's true. Let's add extra checks in a different task / PR.

contracts/cw1-subkeys/src/state.rs
maurolacy: :-)

contracts/cw1-whitelist/src/error.rs
maurolacy: Problem is AFAIK `StdError::Unauthorized` is triggering a deprecation warning.

contracts/cw20-atomic-swap/src/contract.rs
maurolacy: Yes. I'll move all the formatting to the error itself, as mentioned before.

contracts/cw1-whitelist/src/error.rs
ethanfrey: Ah, good point. We never should create `StdError::Unauthorized`.

I just saw the deprecation warning, and understand that. Good as you have it.

ethanfrey:
I think there is a lot of work migrating to proper custom errors, and a lot of this can be done in parallel to other 0.11 upgrade stuff. I am preparing some larger changes to bucket (and adding an indexed bucket) for 0.11-alpha4. 

Now that you have fully migrated the error messages in a few contracts, and we can get some clarity there on what makes sense to us both, it would be great if you could finish up the other contracts with minimal changes. Either somehow asking clippy to ignore the deprecation warning for a minute, or just using a `StdError + Unauthorized` minimal wrapper, like in cw1-whitelist, to the other contracts you have not yet ported.

Then please take out of draft mode, so we can merge this in, and you can keep updating the error codes on one branch, while I can update the Bucket refactor on another branch. I hope to do the Bucket refactor monday or tuesday

maurolacy:
> Now that you have fully migrated the error messages in a few contracts, and we can get some clarity there on what makes sense to us both, it would be great if you could finish up the other contracts with minimal changes. Either somehow asking clippy to ignore the deprecation warning for a minute, or just using a `StdError + Unauthorized` minimal wrapper, like in cw1-whitelist, to the other contracts you have not yet ported.

Turns out is not so easy to do that, because once I change to `ContractError`, that forces me to change all the branches of the `match` in `handle()`, and so on.

What I can do is leave the warnings for now, and focus on the rebase from master and other small changes, in order to remove Draft mode, so we can merge the work so far.

maurolacy:
OK, migrated all but the two new contracts (cw3-fixed-multisig and cw721-nft).

Also rebased from master. Will remove Draft, so you can review and merge this if you want to. And we can migrate the other two contracts in a different MR.

contracts/cw20-escrow/src/contract.rs
maurolacy: Will change those in another branch / PR.

maurolacy:
It's probably a good idea to increase the version of the packages to `0.2.3`.

ethanfrey:
> It's probably a good idea to increase the version of the packages to 0.2.3, before publishing.

Good point, but I don't intend to publish anything until 0.3.0, with 0.11 support. As 0.10 -> 0.11 is breaking.

Thank you for finishing this up to a usable point. I will review and merge and we can both work on this as needed (I would just mess with the bucket and query stuff from my open PRs on cosmwasm with the next 0.11-alpha, and leave all the error work to you to ensure we don't step on toes)

contracts/cw20-escrow/src/error.rs
```diff
@@ -0,0 +1,26 @@
+use cosmwasm_std::StdError;
+use thiserror::Error;
+
+#[derive(Error, Debug)]
```
ethanfrey, line 4: I like this one

contracts/cw20-staking/src/contract.rs
```diff
@@ -84,41 +82,47 @@ pub fn handle<S: Storage, A: Api, Q: Querier>(
...
+        HandleMsg::Transfer { recipient, amount } => {
+            Ok(handle_transfer(deps, env, recipient, amount)?)
+        }
+        HandleMsg::Burn { amount } => Ok(handle_burn(deps, env, amount)?),
         HandleMsg::Send {
             contract,
             amount,
             msg,
-        } => handle_send(deps, env, contract, amount, msg),
+        } => Ok(handle_send(deps, env, contract, amount, msg)?),
```
ethanfrey, line 93: If `handle_send`, etc all return `Result<HandleResponse, ContractError>`, not `StdResult<HandleResponse>`, then we don't need the `Ok(handle_send(..)?)` idiom do we? that is just when we need to auto-convert error cases.

contracts/cw20-staking/src/contract.rs
```diff
@@ -127,22 +131,19 @@ fn get_bonded<Q: Querier>(querier: &Q, contract: &HumanAddr) -> StdResult<Uint12
     bonds.iter().fold(Ok(Uint128(0)), |racc, d| {
         let acc = racc?;
         if d.amount.denom.as_str() != denom {
-            Err(StdError::generic_err(format!(
-                "different denoms in bonds: '{}' vs '{}'",
-                denom, &d.amount.denom
-            )))
+            Err(ContractError::DifferentBondDenom(
```
ethanfrey, line 134: nice

contracts/cw20-staking/src/error.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    #[error("No allowance for this account")]
+    NoAllowance {},
+
+    #[error("Minting cannot exceed the cap")]
+    CannotExceedCap {},
+}
+
+impl From<cw20_base::ContractError> for ContractError {
+    fn from(err: cw20_base::ContractError) -> Self {
```
ethanfrey, line 47: Nice extension.

I did try to figure out how to "extend" enums in rust and it is not possible.
I wonder if we could simplify this one day with codegen? But I guess it is only a few minutes to write such code now.

contracts/cw20-staking/src/error.rs
```diff
@@ -0,0 +1,60 @@
...
+    Std(#[from] StdError),
+
+    #[error("Unauthorized")]
+    Unauthorized {},
+
+    #[error("Validator '{0}' not in current validator set")]
+    NotInValidatorSet(String),
+
+    #[error("Different denominations in bonds: '{0}' vs. '{1}'")]
+    DifferentBondDenom(String, String),
```
ethanfrey, line 16: Question: would it make sense to have these named (struct) variants rather than tuple variants, where there is more than one argument?

I think the generated code is the same, you just then use `DifferentBondDenom{contract: String, message: String}` or something like that to make it clear which side is which.

contracts/cw3-fixed-multisig/Cargo.toml
```diff
@@ -20,9 +20,9 @@ library = []
 [dependencies]
 cosmwasm-std = { version = "0.10.1", features = ["iterator"] }
 cosmwasm-storage = { version = "0.10.1", features = ["iterator"] }
-cw0 = { path = "../../packages/cw0", version = "0.2.2" }
-cw2 = { path = "../../packages/cw2", version = "0.2.2" }
-cw3 = { path = "../../packages/cw3", version = "0.2.2" }
+cw0 = { version = "0.2.2" }
```
ethanfrey, line 23: Ah, interesting way to do this... as long as no one else imports this, we can hold it back to the older package versions. Nice idea!

ethanfrey (approved):
Good stuff here, and thanks for getting it stable.

I will merge this in. Left a few comments along the way that may be useful for the next segment of your error-refactoring PR.

contracts/cw20-escrow/src/error.rs
maurolacy: I think I got those from `cosmwasm-examples`.

contracts/cw20-staking/src/contract.rs
maurolacy: Yes, AFAIK.

contracts/cw20-staking/src/error.rs
maurolacy: Yeah, these cases made me search for "rust inheritance" and the like... traits are fine and dandy, but for abstraction there's nothing better than subtypes (interface inheritance) IMO.

contracts/cw20-staking/src/error.rs
maurolacy: I think it makes sense, yes. I'll change it for clarity.

contracts/cw3-fixed-multisig/Cargo.toml
maurolacy: Yeah, that took a while... :-) happy that I made it compile (and then mergeable).

contracts/cw20-staking/src/error.rs
maurolacy: Come to think of it, this is probably more a case of requiring "plain old" (implementation) inheritance, rather than interface inheritance.

PR #111, opened 2020-10-09
Finish v0.11 migration

ethanfrey:
This finishes up the great work started in @maurolacy 's branch and makes a final cosmwasm-plus PR for the upgrades

### DIFF
--- a/contracts/cw20-base/src/contract.rs
+++ b/contracts/cw20-base/src/contract.rs
@@ -369,7 +369,7 @@ pub fn migrate<S: Storage, A: Api, Q: Querier>(
 
     // once we have "migrated", set the new version and return success
     set_contract_version(&mut deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
-    return Ok(MigrateResponse::default());
+    Ok(MigrateResponse::default())
 }
 
 #[cfg(test)]
--- a/contracts/cw20-base/src/contract.rs
+++ b/contracts/cw20-base/src/contract.rs
@@ -344,6 +344,7 @@ pub fn query_minter<S: Storage, A: Api, Q: Querier>(
 pub fn migrate<S: Storage, A: Api, Q: Querier>(
     deps: &mut Extern<S, A, Q>,
     _env: Env,
+    _info: MessageInfo,
     _msg: MigrateMsg,
 ) -> StdResult<MigrateResponse> {
     let old_version = get_contract_version(&deps.storage)?;
@@ -357,13 +358,18 @@ pub fn migrate<S: Storage, A: Api, Q: Querier>(
     // more recent versions do not have the v prefix
     if old_version.version.starts_with("v0.1.") {
         migrate_v01_to_v02(&mut deps.storage)?;
-        set_contract_version(&mut deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
-        return Ok(MigrateResponse::default());
+    } else if old_version.version.starts_with("0.2") {
+        // no migration between 0.2 and 0.3, correct?
+    } else {
+        return Err(StdError::generic_err(format!(
+            "Unknown version {}",
+            old_version.version
+        )));
     }
-    Err(StdError::generic_err(format!(
-        "Unknown version {}",
-        old_version.version
-    )))
+
+    // once we have "migrated", set the new version and return success
+    set_contract_version(&mut deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+    return Ok(MigrateResponse::default());
 }
 
 #[cfg(test)]
@@ -917,8 +923,8 @@ mod tests {
         );
 
         // run the migration
-        let (env, _info) = mock_env_info(HumanAddr::from("admin"), &[]);
-        migrate(&mut deps, env, MigrateMsg {}).unwrap();
+        let (env, info) = mock_env_info(HumanAddr::from("admin"), &[]);
+        migrate(&mut deps, env, info, MigrateMsg {}).unwrap();
 
         // make sure the version is updated
         assert_eq!(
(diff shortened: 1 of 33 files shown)


maurolacy:
Ey, thanks for finishing this! I'll review it today / tomorrow.

contracts/cw20-base/src/contract.rs
```diff
@@ -902,7 +911,7 @@ mod tests {
         );
 
         // run the migration
-        let env = mock_env(HumanAddr::from("admin"), &[]);
+        let (env, _info) = mock_env_info(HumanAddr::from("admin"), &[]);
```
ethanfrey, line 920: don't we need to add info to migrate so it compiles in wasm mode?

orkunkl: Yes you are right. I did it locally but I suppose you are working on it don't want to tip toes.

ethanfrey:
@maurolacy can you give a quick review on these changes before merging them into your branch (you should be able to do that). Then I can merge your branch into master. And we have 0.11 support!

I guess I wait til Monday to cut the v0.3.0 release

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -339,15 +350,18 @@ fn reverse_proposals<S: Storage, A: Api, Q: Querier>(
...
-        .map(map_proposal)
+        .map(|p| map_proposal(&env.block, p))
         .collect();
 
     Ok(ProposalListResponse { proposals: props? })
 }
 
-fn map_proposal(item: StdResult<(Vec<u8>, Proposal)>) -> StdResult<ProposalResponse> {
+fn map_proposal(
+    block: &BlockInfo,
```
maurolacy, line 360: Nice.

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -779,10 +794,10 @@ mod tests {
 
         // Expired proposals cannot be voted
         let env = match voting_period {
-            Duration::Time(duration) => mock_env_time(VOTER4, env.block.time + duration + 1),
-            Duration::Height(duration) => mock_env_height(VOTER4, env.block.height + duration + 1),
+            Duration::Time(duration) => mock_env_time(duration + 1),
+            Duration::Height(duration) => mock_env_height(duration + 1),
```
maurolacy, line 798: Nice.

contracts/cw3-fixed-multisig/src/state.rs
```diff
@@ -32,12 +32,15 @@ pub struct Proposal {
...
-    pub fn current_status(&self) -> Status {
+    pub fn current_status(&self, block: &BlockInfo) -> Status {
         let mut status = self.status;
 
-        // if open, check if voting is passed on timed out
+        // if open, check if voting is passed or timed out
         if status == Status::Open && self.yes_weight >= self.required_weight {
-            status = Status::Passed
+            status = Status::Passed;
+        }
```
maurolacy, line 41: Nice. No need to use `else if` here. Let's just not change the order of these checks.

contracts/cw721-base/src/contract.rs
```diff
@@ -41,63 +42,62 @@ pub fn init<S: Storage, A: Api, Q: Querier>(
...
-    description: Option<String>,
-    image: Option<String>,
+    _env: Env,
+    info: MessageInfo,
+    msg: MintMsg,
+    // token_id: String,
+    // owner: HumanAddr,
+    // name: String,
+    // description: Option<String>,
+    // image: Option<String>,
```
maurolacy, line 83: Remove commented code.

contracts/cw721-base/src/contract.rs
```diff
@@ -450,15 +463,22 @@ fn query_all_approvals<S: Storage, A: Api, Q: Querier>(
...
-        .map(|item| {
-            item.and_then(|(k, expires)| {
-                Ok(cw721::Approval {
-                    spender: deps.api.human_address(&k.into())?,
-                    expires,
-                })
-            })
+        .filter_map(|item| match item {
+            Ok((k, expires)) => {
+                // we remove all expired operators from the result
```
maurolacy, line 468: Sure about this? I'm not sure, but listing expired operators can be useful for debugging / accountability issues.

contracts/cw721-base/src/contract.rs
```diff
@@ -497,9 +518,14 @@ fn query_all_nft_info<S: Storage, A: Api, Q: Querier>(
...
 
-fn humanize_approvals<A: Api>(api: A, info: &TokenInfo) -> StdResult<Vec<cw721::Approval>> {
+fn humanize_approvals<A: Api>(
+    api: A,
+    block: &BlockInfo,
+    info: &TokenInfo,
+) -> StdResult<Vec<cw721::Approval>> {
     info.approvals
         .iter()
+        .filter(|apr| !apr.expires.is_expired(block))
```
maurolacy, line 528: Same here.

maurolacy (approved):
Quick review, LGTM.

contracts/cw721-base/src/contract.rs
ethanfrey: I guess I can keep them, but then add another flag to the response "is_expired: bool". Is it useful info on the client, without forcing calculations there

maurolacy: Good idea. Or directly add the expiration info.

Another option would be to add a boolean flag to the query, i. e. `include_expired` or so.

ethanfrey: Should be addressed here: https://github.com/CosmWasm/cosmwasm-plus/pull/111/commits/09b7b31427eb9c484e070697c91dba89c1a22310

ethanfrey: > Another option would be to add a boolean flag to the query, i. e. include_expired or so.

Just read this. I actually prefer this idea

ethanfrey: I will undo my change and redo it the other way. Let me know which commit you prefer.

maurolacy: I too think the query option is better.

ethanfrey: Look at https://github.com/CosmWasm/cosmwasm-plus/pull/111/commits/0525a341b7bc29cddaa40405c84ad4561aef6029

PR #116, opened 2020-10-10
Use Include/Exclude Bounds to define range searches

ethanfrey:
I think this is what you were suggesting Mauro - public API doesn't require bit-twiddling.

This is incomplete and definitely missing tests, but please review the idea (and maybe suggest a better path forward).

When we agree on a proper interface and implement it, then I will use the new `Item` and `Map` and `Prefix/Bound` types to reimplement the storage of `cw3-fixed-multisig` as a real world test case. 

- [x] Finalize Range API
- [x] Solid unit tests (edge conditions, both Ascending and Descending)
- [x] Document in README
- [ ] (Move/copy code samples to rustdocs?)
- [ ] Ensure wasmd has same boundary behavior
- [ ] Update `cw3-fixed-multisig` to new API

Once this is out, I would love to cut a `0.3.0` release, and then we can add secondary indexes in `0.3.1` as a non-breaking change

### DIFF
--- a/packages/storage-plus/src/prefix.rs
+++ b/packages/storage-plus/src/prefix.rs
@@ -122,6 +122,7 @@ mod test {
     use super::*;
     use cosmwasm_std::testing::MockStorage;
 
+    #[test]
     fn ensure_proper_range_bounds() {
         let mut store = MockStorage::new();
         // manually create this - not testing nested prefixes here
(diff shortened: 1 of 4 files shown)


maurolacy:
OK. Even if you already merged it, I'll review the other MR (storage-maps) first, as it can help me gain more context.

packages/storage-plus/src/prefix.rs
```diff
@@ -3,10 +3,22 @@ use serde::de::DeserializeOwned;
...
+use crate::iter_helpers::{concat, deserialize_kv, trim};
+
+/// Bound is used to defines the two ends of a range, more explicit than Option<u8>
+/// None means that we don't limit that side of the range at all.
+/// Include means we use the given bytes as a limit and *include* anything at that exact key
+/// Exclude means we use the given bytes as a limit and *exclude* anything at that exact key
+#[derive(Copy, Clone, Debug)]
+pub enum Bound<'a> {
+    Include(&'a [u8]),
+    Exclude(&'a [u8]),
```
maurolacy, line 18: Maybe `Inclusive` / `Exclusive` (like in inclusive / exclusive bounds) are better names here.

maurolacy (approved):
I like it, in that it hides the bound calculation under the impl. It also has a clearer naming IMO.

ethanfrey:
Migrating cw3-fixed-multisig is a bigger task than I thought, as I need/want to add support for u64 keys natively and not force &id.as_be_bytes() everywhere, but provide a cleaner interface. This is a big boon to the API, but more work than I want to throw in this PR.

Will tackle that as a follow up

PR #118, opened 2020-10-12
Move shared Balance struct to cw20

### LINKED ISSUE #115: Remove dependency between cw20_escrow and cw20_atomic_swap
There is an import in the escrow contract: `use cw20_atomic_swap::balance::Balance;`

We should be able to  `use cw0::Balance`

Also, see if there is anything else to de-dup.

maurolacy:
Closes #115.

### DIFF
--- a/contracts/cw20-atomic-swap/src/lib.rs
+++ b/contracts/cw20-atomic-swap/src/lib.rs
@@ -1,4 +1,3 @@
-pub mod balance;
 pub mod contract;
 mod error;
 pub mod msg;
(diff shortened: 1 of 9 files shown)


contracts/cw20-atomic-swap/src/lib.rs
```diff
@@ -1,4 +1,3 @@
-pub mod balance;
```
ethanfrey, line 1: You removed the reference, but I don't see the file removed in this diff

packages/cw20/src/balance.rs
```diff
@@ -3,7 +3,8 @@ use serde::{Deserialize, Serialize};
 
 use cosmwasm_std::Coin;
 use cw0::NativeBalance;
-use cw20::Cw20Coin;
+
+use crate::Cw20Coin;
```
ethanfrey, line 7: Ahh.... nevermind, it was moved to cw20. Excellent!

ethanfrey (approved):
Great work. Thank you

PR #108, opened 2020-10-07
Secondary indexes

ethanfrey:
Closes #123 as part of the cleanup
Closes #81 (completed in #122 but merged to master here)
Closes https://github.com/CosmWasm/cosmwasm/issues/468

This is a "from scratch" re-design of Buckets and support for secondary indexes.

Heavily inspired by work on https://github.com/CosmWasm/cosmwasm/pull/557 and https://github.com/CosmWasm/cosmwasm/pull/559

#109 takes over the ground-work, setting up the basic Keys, Path, Map types.

This PR builds on top of that and add indexes.

PR #122, opened 2020-10-13
Add TokensByOwner for cw721-base

### LINKED ISSUE #81: Implement TokensByOwner on base NFT contract
Builds on #44 

Add secondary indexes somehow and allow query on owner not just token id

Requires secondary indexes to do cleanly: https://github.com/CosmWasm/cosmwasm/issues/468

ethanfrey:
Closes #81 

Builds on #108 (merge that first)

Ports to use `cw-storage-plus` rather than `cosmwasm-storage`, then makes use of IndexedMap

- [x] Port to `cw-storage-plus`
- [x] Better support for &str Map keys
- [x] Use `IndexedMap`
- [x] Enable query by owner
- [x] Cover query by owner with tests
- [x] Allow setting bounds (pagination)
- [x] Raise issues for remaining TODOs

### DIFF
--- a/contracts/cw721-base/src/contract.rs
+++ b/contracts/cw721-base/src/contract.rs
@@ -3,7 +3,7 @@ use cosmwasm_std::{
     HumanAddr, InitResponse, MessageInfo, Order, Querier, StdResult, Storage, KV,
 };
 
-use cw0::{calc_range_start_human, calc_range_start_string};
+use cw0::maybe_canonical;
 use cw2::set_contract_version;
 use cw721::{
     AllNftInfoResponse, ApprovedForAllResponse, ContractInfoResponse, Expiration, NftInfoResponse,
@@ -13,9 +13,9 @@ use cw721::{
 use crate::error::ContractError;
 use crate::msg::{HandleMsg, InitMsg, MintMsg, MinterResponse, QueryMsg};
 use crate::state::{
-    contract_info, contract_info_read, increment_tokens, mint, mint_read, num_tokens, operators,
-    operators_read, tokens, tokens_read, Approval, TokenInfo,
+    increment_tokens, num_tokens, Approval, TokenInfo, CONTRACT_INFO, MINTER, OPERATORS, TOKENS,
 };
+use cw_storage_plus::{Bound, OwnedBound};
 
 // version info for migration info
 const CONTRACT_NAME: &str = "crates.io:cw721-base";
@@ -33,9 +33,9 @@ pub fn init<S: Storage, A: Api, Q: Querier>(
         name: msg.name,
         symbol: msg.symbol,
     };
-    contract_info(&mut deps.storage).save(&info)?;
+    CONTRACT_INFO.save(&mut deps.storage, &info)?;
     let minter = deps.api.canonical_address(&msg.minter)?;
-    mint(&mut deps.storage).save(&minter)?;
+    MINTER.save(&mut deps.storage, &minter)?;
     Ok(InitResponse::default())
 }
 
@@ -77,7 +77,7 @@ pub fn handle_mint<S: Storage, A: Api, Q: Querier>(
     info: MessageInfo,
     msg: MintMsg,
 ) -> Result<HandleResponse, ContractError> {
-    let minter = mint(&mut deps.storage).load()?;
+    let minter = MINTER.load(&deps.storage)?;
     let sender_raw = deps.api.canonical_address(&info.sender)?;
 
     if sender_raw != minter {
@@ -92,10 +92,14 @@ pub fn handle_mint<S: Storage, A: Api, Q: Querier>(
         description: msg.description.unwrap_or_default(),
         image: msg.image,
     };
-    tokens(&mut deps.storage).update(msg.token_id.as_bytes(), |old| match old {
-        Some(_) => Err(ContractError::Claimed {}),
-        None => Ok(token),
-    })?;
+    TOKENS.update(
+        &mut deps.storage,
+        msg.token_id.as_bytes(),
+        |old| match old {
+            Some(_) => Err(ContractError::Claimed {}),
+            None => Ok(token),
+        },
+    )?;
 
     increment_tokens(&mut deps.storage)?;
 
@@ -168,13 +172,13 @@ pub fn _transfer_nft<S: Storage, A: Api, Q: Querier>(
     recipient: &HumanAddr,
     token_id: &str,
 ) -> Result<TokenInfo, ContractError> {
-    let mut token = tokens(&mut deps.storage).load(token_id.as_bytes())?;
+    let mut token = TOKENS.load(&mut deps.storage, token_id.as_bytes())?;
     // ensure we have permissions
     check_can_send(&deps, env, info, &token)?;
     // set owner and remove existing approvals
     token.owner = deps.api.canonical_address(recipient)?;
     token.approvals = vec![];
-    tokens(&mut deps.storage).save(token_id.as_bytes(), &token)?;
+    TOKENS.save(&mut deps.storage, token_id.as_bytes(), &token)?;
     Ok(token)
 }
 
@@ -231,7 +235,7 @@ pub fn _update_approvals<S: Storage, A: Api, Q: Querier>(
     add: bool,
     expires: Option<Expiration>,
 ) -> Result<TokenInfo, ContractError> {
-    let mut token = tokens(&mut deps.storage).load(token_id.as_bytes())?;
+    let mut token = TOKENS.load(&mut deps.storage, token_id.as_bytes())?;
     // ensure we have permissions
     check_can_approve(&deps, env, info, &token)?;
 
@@ -257,7 +261,7 @@ pub fn _update_approvals<S: Storage, A: Api, Q: Querier>(
         token.approvals.push(approval);
     }
 
-    tokens(&mut deps.storage).save(token_id.as_bytes(), &token)?;
+    TOKENS.save(&mut deps.storage, token_id.as_bytes(), &token)?;
 
     Ok(token)
 }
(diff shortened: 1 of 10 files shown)


contracts/cw721-base/src/contract.rs
```diff
@@ -478,12 +484,14 @@ fn query_all_approvals<S: Storage, A: Api, Q: Querier>(
     start_after: Option<HumanAddr>,
     limit: Option<u32>,
 ) -> StdResult<ApprovedForAllResponse> {
-    let owner_raw = deps.api.canonical_address(&owner)?;
     let limit = limit.unwrap_or(DEFAULT_LIMIT).min(MAX_LIMIT) as usize;
-    let start = calc_range_start_human(deps.api, start_after)?;
+    let start_canon = maybe_canonical(deps.api, start_after)?;
+    let start = OwnedBound::exclusive(start_canon);
```
maurolacy, line 489: Nice, but is this really need? I think plainly doing
```suggestion
    let start_raw = deps.api.canonical_address(&start_after)?;
    let start = OwnedBound::exclusive(Some(start_raw));
```
is clearer.

Moreover, `exclusive()` can directly receive the value and build the option inside. There could be another helper for `OwnedBound::None`, if needed.

contracts/cw721-base/src/contract.rs
```diff
@@ -478,12 +484,14 @@ fn query_all_approvals<S: Storage, A: Api, Q: Querier>(
...
 
-    let res: StdResult<Vec<_>> = operators_read(&deps.storage, &owner_raw)
-        .range(start.as_deref(), None, Order::Ascending)
+    let owner_raw = deps.api.canonical_address(&owner)?;
+    let res: StdResult<Vec<_>> = OPERATORS
+        .prefix(&owner_raw)
+        .range(&deps.storage, start.bound(), Bound::None, Order::Ascending)
```
maurolacy, line 494: `OwnedBound` may  be easier to use in some cases, but having two structs for one purpose requires more helpers, like the `bound()` method above. I fr one would try to stick to one `Bound`, unless really needed / justified.

contracts/cw721-base/src/contract.rs
```diff
@@ -498,16 +506,38 @@ fn parse_approval<A: Api>(api: A, item: StdResult<KV<Expiration>>) -> StdResult<
...
 
+fn query_tokens<S: Storage, A: Api, Q: Querier>(
+    deps: &Extern<S, A, Q>,
+    owner: HumanAddr,
+    start_after: Option<String>,
+    limit: Option<u32>,
+) -> StdResult<TokensResponse> {
+    let limit = limit.unwrap_or(DEFAULT_LIMIT).min(MAX_LIMIT) as usize;
+    // TODO: get pagination working on second indexes
+    let _start = OwnedBound::exclusive(start_after.map(Vec::from));
```
maurolacy, line 517: I see now why `exclusive()` takes an `Option`.

I think there must be a clearer way for doing this, but cannot see it right now. Maybe building and using an `Option<Bound>`? `Bound` can be a struct (not an enum) with a boolean `exclusive` flag, by example. And the helpers take care of setting it... I'm not sure, just thinking out loud.

maurolacy (commented):
Looks good. I think the use of bounds is still a little bit confusing / needlessly complex.

Added some comments to try and help clarifying the design.

contracts/cw721-base/src/contract.rs
ethanfrey: The issue is `calc_range_start_human` was a "do everything" function. I could replace it, or just try to pull out the reusable parts.

`start_after` is `Option<HumanAddr>`, which is the complexity - doing `maybe_canonical` instead of `deps.api.canonical` (it just does a Option::map).  Your proposed line `let start_raw = deps.api.canonical_address(&start_after)?;` would not compile. However, I could inline the contents of `maybe_canonical`, which is just one line and maybe cleaner.

contracts/cw721-base/src/contract.rs
ethanfrey: Fair enough. I can rename `OwnedBound` to bound. Only issue is when writing tests it is more verbose... 

`Bound::Exclusive(b"foo")` -> `Bound::Exclusive(b"foo".to_vec())`

which is something I hit quite a bit, and tried to "clean up". Then again, the actual usage is easier the way you describe and maybe best I optimize for production code, not my tests :smile:

contracts/cw721-base/src/contract.rs
ethanfrey: `range` could accept `Option<Bound>` and we remove `Bound::None`. 

Then we replace this with `let _start = start_after.map(|x| Bound::Exclusive(x.into()));`

Before I had a rather verbose match statement to map None as well as Some, so I made the helpers. I guess it is often better just to type a `map` than call a one-line helper function that increases the API surface to learn.

ethanfrey:
My takeaways. Only use one type:

```
pub enum Bound {
  Inclusive(Vec<u8>),
  Exclusive(Vec<u8>)
}
```

and accept `Option<Bound>` in `range`.

This will lead to lines like:

`let start: Option<Bound> = start_after.map(|human| Ok(Bound::Exclusive(deps.api.canonical_addr(&human)?)).transpose()?;

Which is not so legible or user-friendly. I guess having a function for handling the `human_addr` case is good. For a string/int, we could get:

`let start: Option<Bound> = start_after.map(|s| Bound::Exclusive(s.into()))`

or 

`let start: Option<Bound> = start_after.map(|i| Bound::Exclusive(i.to_be_bytes().into()))`

both of which are simple enough, without all these helpers.

---

Does the above make sense to you? Other suggestion?

contracts/cw721-base/src/contract.rs
maurolacy: I see the need for `maybe_canonical` now. I still don't like it, though :-).

I think the issue here is that you are still computing a `start_after`. Wouldn't be better to pass the original `HumanAddr` wrapped in an `Bound::Exclusive` container, and handle all the range stuff inside? Not sure if that collides with use cases, though.

contracts/cw721-base/src/contract.rs
maurolacy: Yep. Though clarity of the API should reflect in unit tests, though. Maybe there's some syntactic sugar equivalent to `to_vec()` around?

contracts/cw721-base/src/contract.rs
maurolacy: > range could accept Option<Bound> and we remove Bound::None.

I think that's a good idea, yes.

maurolacy:
Makes sense. Will review later again with the IDE, to see and try some variants.

contracts/cw721-base/src/contract.rs
ethanfrey: The json interface is "no field", or `start_after: "cosmos12345..."`, thus the option.

I don't want to expose the bound to json (which is where the `Option<HumanAddr>` comes from

maurolacy: I see. Sorry about the previous comment, it was based on a misunderstanding. You are right, not exposing `Bound` to the user is a good idea.

contracts/cw721-base/src/state.rs
```diff
@@ -44,3 +44,10 @@ pub fn increment_tokens<S: Storage>(storage: &mut S) -> StdResult<u64> {
     TOKEN_COUNT.save(storage, &val)?;
     Ok(val)
 }
+
+// indexed map needs function, not const (for now at least)
```
maurolacy, line 50: I guess this is because of the lambda, right?

ethanfrey: Yes, you cannot build vec structs with const. It must all be in the stack and computer at compile time.

In the index refactor (using structs instead of this), I could theoretically create the structure as a const (and I manually created an IndexMap struct as const in a branch), but the compiler did not let me use `MultiIndex::new` as it took a function pointer as an argument - even if I declare that function pointer as a `const` (which let's me build a struct), there is no way to declare that in `MultiIndex::new` signature. I spent like 2 hours in the const rabbit hole...

contracts/cw721-base/src/state.rs
```diff
@@ -44,3 +44,10 @@ pub fn increment_tokens<S: Storage>(storage: &mut S) -> StdResult<u64> {
...
+pub fn tokens<'a, S: Storage + 'a>() -> IndexedMap<'a, 'a, &'a str, TokenInfo, S> {
+    IndexedMap::<&str, TokenInfo, S>::new(b"tokens")
+        .with_index("owner", b"tokens__owner", |d| d.owner.to_vec())
+        .unwrap()
```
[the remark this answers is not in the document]
ethanfrey: Yeah, I actually refactored `with_index` stuff here: https://github.com/CosmWasm/cosmwasm-plus/pull/108 and will merge this in. It makes the checks compile-time, rather than run-time ensuring the same index name is not used twice.

contracts/cw721-base/src/state.rs
```diff
@@ -44,3 +44,10 @@ pub fn increment_tokens<S: Storage>(storage: &mut S) -> StdResult<u64> {
     TOKEN_COUNT.save(storage, &val)?;
     Ok(val)
 }
+
+// indexed map needs function, not const (for now at least)
+pub fn tokens<'a, S: Storage + 'a>() -> IndexedMap<'a, 'a, &'a str, TokenInfo, S> {
+    IndexedMap::<&str, TokenInfo, S>::new(b"tokens")
```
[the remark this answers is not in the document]
ethanfrey: It is really `idx_fn: fn(&T) -> Vec<u8>` which is rejected. This *may* be implemented via a function stored in a `const` (and I have done so), but there is no way I found to inform Rust this in the `new` type signature.

I can try my branch again on Rust 1.47

contracts/cw721-base/src/state.rs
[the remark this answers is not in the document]
ethanfrey: This is the current demo on how to construct this - a bit harder but no unwrap and the usage is much cleaner: https://github.com/CosmWasm/cosmwasm-plus/pull/125/files#diff-e914973c827114d0c37ed9b0a9d7b0bb3105fba992674a1cb2b32369fa6d50bfR152-R170

contracts/cw721-base/src/state.rs
maurolacy: :-) I can imagine.

contracts/cw721-base/src/state.rs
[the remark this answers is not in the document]
maurolacy: Good luck with that. I'll keep looking; it's a good way to be aware of the new functionality.

PR #130, opened 2020-10-19
Sketch integration test framework

ethanfrey:
Handle most of the memory items needed (with `Cell` and `Box<dyn Trait>`). However, I hit an issue trying to construct the `Extern`. Ideally we would only need `&mut Storage`, so I could use `RefCell`, but that is acceptable. However, passing in `Router` as the `Querier`, which is also the execution environment of the contract itself, is impossible. 

I basically need `&Router` and `Router` at once! If we could loosen `Extern` to only require `&Querier` rather than `Querier` (which is all it actually uses), then this would be much easier to implement in Rust.

Raised issue https://github.com/CosmWasm/cosmwasm/issues/584 (and maybe this code belongs in `cosmwasm` to iterate closer with the APIs)

----

Update: With the 0.12 cosmwasm API this works quite well. More or less code-complete, including bank.

Done:
- [x] Test bank
- [x] Test WasmRouter
- [x] Test Router - move coins, call contract
- [x] Add Reflect check - contract calling others: success + error
- [ ] Handle rollback on error (transactional) - handled in #134 
- [x] Evaluate high level API

This rollback on error is a bit trickier than I thought... it affects cases where one step succeeds (and writes) and a sub-message fails. Maybe I merge this first, then do another PR for that.


ethanfrey:
@webmaster128 

Code is (almost) all implemented as I envisioned it. A number of TODOs and of course this needs tests with real contracts, but the Rust compiler is happy with the lifetimes here.

There is one obvious issue about a contract querying it's own address in handle, which leads to a runtime error. I can add tests to trigger with mock contracts and look at handling that better, but let's discuss this on a call Wednesday first.

ethanfrey:
I will merge this, as there is quite some work on top #134 and #135 and I haven't gotten a review here. Happy for any comments on this code and I will address it later. But quite some is modified in the other PRs anyway.

PR #132, opened 2020-10-27
Fix SendNft in Cw721-base

ethanfrey:
- [x] Fix cw721: The receiver type in code now matches the readme
- [x] Fix cw721-base: Send the proper message when SendNft is executed.

### DIFF
--- a/contracts/cw721-base/src/contract.rs
+++ b/contracts/cw721-base/src/contract.rs
@@ -1,13 +1,13 @@
 use cosmwasm_std::{
-    attr, from_binary, to_binary, Api, Binary, BlockInfo, CosmosMsg, Env, Extern, HandleResponse,
-    HumanAddr, InitResponse, MessageInfo, Order, Querier, StdError, StdResult, Storage, KV,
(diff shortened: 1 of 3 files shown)


contracts/cw721-base/src/contract.rs
```diff
@@ -139,18 +139,18 @@ pub fn handle_send_nft<S: Storage, A: Api, Q: Querier>(
     token_id: String,
     msg: Option<Binary>,
 ) -> Result<HandleResponse, ContractError> {
-    // Unwrap message first
-    let msgs: Vec<CosmosMsg> = match &msg {
-        None => vec![],
-        Some(msg) => vec![from_binary(msg)?],
-    };
-
```
maurolacy, line 147: If I understand correctly, the bug here is that we're accepting (and depending on) any kind of message from the client, whereas we should send only a `Cw721ReceiveMsg` type of message.

ethanfrey: Yes, we don't send what they tell us to send with the full permission of the contract. But rather a specified `Cw721ReceiveMsg`, that only triggers an expected path on some contract.

This is how it is done in cw20.

Also, the cw721 spec was wrong. I updated the README but not the code, so had to update that

PR #137, opened 2020-11-07
Multi test example

ethanfrey:
This adds a new "multi test" for sending cw20 tokens to an escrow and releasing it.

This should demonstrate how to use multi-test. We can also use this to see how to adjust the high-level APIs. 

Please let me know how we can improve the APIs. But this should be much easier than everything we had until now.


maurolacy (commented):
Taking a look at this now. Look good, will try to finish reviewing it by tomorrow.

contracts/cw20-escrow/src/integration_test.rs
```diff
@@ -0,0 +1,141 @@
...
+        cw20_whitelist: None,
+    });
+    let create_bin = to_binary(&create_msg).unwrap();
+    let send_msg = Cw20HandleMsg::Send {
+        contract: escrow_addr.clone(),
+        amount: Uint128(1200),
+        msg: Some(create_bin),
+    };
+    let res = router
+        .execute_contract(&cash_addr, &owner, &send_msg, &[])
```
maurolacy, line 101: This is trivial, but maybe `.execute_contract(&owner, &cash_addr, ...)` is clearer, i. e. passing sender before recipient.

contracts/cw20-escrow/src/integration_test.rs
```diff
@@ -0,0 +1,141 @@
...
+        .unwrap();
+    println!("{:?}", res.attributes);
+    assert_eq!(6, res.attributes.len());
+
+    // ensure balances updated
+    let owner_balance = cash.balance(&router, owner.clone()).unwrap();
```
maurolacy, line 107: This is totally accessory, but maybe doing a query for the balance here over `router` would be clearer. I mean, given that this is testing the integration, it would be nice to test the queries too.

contracts/cw20-escrow/src/integration_test.rs
```diff
@@ -0,0 +1,141 @@
...
+    assert_eq!(owner_balance, Uint128(3800));
+    let escrow_balance = cash.balance(&router, escrow_addr.clone()).unwrap();
+    assert_eq!(escrow_balance, Uint128(1200));
+
+    // ensure escrow properly created
+    let details: DetailsResponse = router
+        .wrap()
```
maurolacy, line 114: It would be nice it this could be avoided, i.e. handled internally or so. In any case, `get_wrapper()` or `get_querier()` would be a better name? This is exposing something, not wrapping it, AFAIK.

contracts/cw20-escrow/src/integration_test.rs
```diff
@@ -0,0 +1,141 @@
...
+        .query_wasm_smart(&escrow_addr, &QueryMsg::Details { id: id.clone() })
```
maurolacy, line 115: Very nice.

maurolacy (approved):
Please take into account that these are my first impressions. I would probably have a better opinion when using this API to write some integration tests myself.

I like it! It makes writing complex integration tests not only possible, but also relatively straightforward.

contracts/cw20-escrow/src/integration_test.rs
ethanfrey: It does the query via router.

I really need to rename/document the `cash: CW20Contract` type better. Rather than manually building QueryRequests for this contract, cash.balance will do that for you, a short-cut for 6 lines or so of boilerplate.

All the actual work happens via the router, but this provides a nice helper wrapper. 

We could add one explicit query to show that. But mainly I think we need to start documenting and improving naming of these contract-specific helpers. That would make it much easier for all the tests to just call methods rather than build up the raw request and message types.

contracts/cw20-escrow/src/integration_test.rs
ethanfrey: `router.as_querier_wrapper()` would be the most explicit. I just felt it was too verbose.

I am happy for better naming, but take a look at some of the big changes in cosmwasm-std for 0.12 when we took in `dyn Querier` and such instead of generics. (And I am happy for better approaches/naming).

Basically, the `Querier trait` only required one method (`raw_query`) and had a lot of default generic implementations. As I made this `&dyn Querier`, we could no longer include generics and defaults on the trait. You can see how [`QuerierWrapper` assumed most of the functionality that used to be in `Querier`](https://github.com/CosmWasm/cosmwasm/blob/master/packages/std/src/traits.rs#L77-L248). And can easily be created from a `Querier`.

I am not sure what a better way to handle this is. I cannot see how to use "deref" magic here to allow `router.query_wasm_smart(...)` directly, which is what I wanted. Please take a look at that wrapper struct (now you understand why) and I am happy for a concrete suggestion to make this cleaner.

contracts/cw20-escrow/src/integration_test.rs
ethanfrey: Sounds good. Nice suggestion,

contracts/cw20-escrow/src/integration_test.rs
```diff
@@ -0,0 +1,141 @@
...
+    let escrow_id = router.store_code(contract_escrow());
+    let escrow_addr = router
+        .instantiate_contract(escrow_id, &owner, &InitMsg {}, &[], "Escrow")
+        .unwrap();
+
+    // they are different
+    assert_ne!(cash_addr, escrow_addr);
+
+    // set up cw20 helpers
+    let cash = Cw20Contract(cash_addr.clone());
```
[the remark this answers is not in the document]
ethanfrey: Can you propose better naming or docs for this struct? We will be using this pattern a lot more and this should not be confusing. 

I would be happy for a PR or an issue with a clearly proposed design. I think these helpers are only used in this new multi-test

contracts/cw20-escrow/src/integration_test.rs
```diff
@@ -0,0 +1,141 @@
+#![cfg(test)]
+
+use cosmwasm_std::testing::{mock_env, MockApi, MockStorage};
+use cosmwasm_std::{coins, to_binary, HumanAddr, Uint128};
+use cw20::{Cw20CoinHuman, Cw20Contract, Cw20HandleMsg};
+use cw_multi_test::{Contract, ContractWrapper, Router, SimpleBank};
+
+use crate::msg::{CreateMsg, DetailsResponse, HandleMsg, InitMsg, QueryMsg, ReceiveMsg};
+
+fn mock_router() -> Router {
```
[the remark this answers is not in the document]
ethanfrey: Good Point. I will rename it.

ethanfrey:
Thank you for the feedback. I will make the adjustments on the two actionable items (rename and argument reordering).

The other comments highlight some places where better naming/docs/usability would be nice, but it is unclear how exactly it could be improved. A clear proposal for any of those points as issue or PR would be very well received.

contracts/cw20-escrow/src/integration_test.rs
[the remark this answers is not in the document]
ethanfrey: How about `BaseApp` or `Application`? That is what they call the equivalent in the Cosmos SDK.

The object that holds all state and the various handlers.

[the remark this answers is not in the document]
maurolacy: I think `Application` or `App` is better than `Router`. In the end, it's a matter of documenting it.

contracts/cw20-escrow/src/integration_test.rs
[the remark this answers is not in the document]
maurolacy: OK, I'll think about this, and play with it using the multi-test code base, to see if I can come up with a better name / design.

contracts/cw20-escrow/src/integration_test.rs
maurolacy: I agree. Some documentation / usage examples will solve this. Explicitly mentioning that these are wrappers / helpers on top of Router / App is a good idea.

contracts/cw20-escrow/src/integration_test.rs
maurolacy: I see, thanks for the explanation about the need for Wrapper. Naming is tricky... will play with the code base to se if I can propose something concrete.

PR #144, opened 2020-11-16
Rename Router -> App

ethanfrey:
Inspired by https://github.com/CosmWasm/cosmwasm-plus/pull/137#discussion_r523380556

Maybe this name is cleaner?


maurolacy (approved):
I think it's clearer, yes.

Issue #145, opened 2020-11-26
Asymmetries between query and execute in CW1 (subkeys)


willclarktech:
I noticed there are a couple of asymmetries between query and execute in the CW1 contract:
1. Why is it called `CanSend` rather than `CanExecute`?
1. Since `Execute` takes a vector of msgs why wouldn't `CanSend` also take a vector?

maurolacy:
You're right. The query name `CanSend` is probably for historical reasons. It would be better called `CanExecute`, as you say.
Regarding `CanExecute` accepting a vector of msgs, that's a nice improvement.

I'll implement the suggested changes.

ethanfrey:
I believe this can be closed.

The name was modified as requested, and we decided not to support mutliple messages in the query

PR #146, opened 2020-11-28
Rename CanSend to CanExecute for generality

maurolacy:
Closes the first part of #145, renaming of `CanSend` query enum and associated methods.

Will implement the seconds part, i. e. modifying `CanExecute` params, in a follow-up PR against this branch, for ease of reviewing.

### DIFF
--- a/contracts/cw1-subkeys/Cargo.toml
+++ b/contracts/cw1-subkeys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cw1-subkeys"
-version = "0.4.0"
+version = "0.3.2"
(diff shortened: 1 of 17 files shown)


contracts/cw1-subkeys/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "cw1-subkeys"
-version = "0.3.2"
+version = "0.4.0"
```
ethanfrey, line 3: Please don't update the version yet. There is a script to update all versions and we can do that once before publishing. (They all went from 0.11 -> 0.12, so everything is breaking)

ethanfrey (approved):
lgtm.

please revert the version bump, then feel free to merge yourself.

PR #147, opened 2020-11-28
Support multiple messages in CanExecute query

maurolacy:
Closes the last part of #145, supporting multiple messages in CanExecute query.

### DIFF
--- a/packages/cw1/src/query.rs
+++ b/packages/cw1/src/query.rs
@@ -21,5 +21,5 @@ where
 
 #[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
 pub struct CanExecuteResponse {
-    pub can_execute: bool,
+    pub can_execute: Vec<bool>,
 }
--- a/contracts/cw1-subkeys/schema/query_msg.json
+++ b/contracts/cw1-subkeys/schema/query_msg.json
@@ -64,12 +64,15 @@
         "can_execute": {
           "type": "object",
           "required": [
-            "msg",
+            "msgs",
             "sender"
           ],
           "properties": {
-            "msg": {
-              "$ref": "#/definitions/CosmosMsg_for_Empty"
+            "msgs": {
+              "type": "array",
+              "items": {
+                "$ref": "#/definitions/CosmosMsg_for_Empty"
+              }
             },
             "sender": {
               "$ref": "#/definitions/HumanAddr"
--- a/contracts/cw1-subkeys/src/contract.rs
+++ b/contracts/cw1-subkeys/src/contract.rs
@@ -296,7 +296,7 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
         QueryMsg::AdminList {} => to_binary(&query_admin_list(deps)?),
         QueryMsg::Allowance { spender } => to_binary(&query_allowance(deps, spender)?),
         QueryMsg::Permissions { spender } => to_binary(&query_permissions(deps, spender)?),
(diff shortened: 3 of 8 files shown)


packages/cw1/src/query.rs
```diff
@@ -21,5 +21,5 @@ where
 
 #[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
 pub struct CanExecuteResponse {
-    pub can_execute: bool,
+    pub can_execute: Vec<bool>,
```
webmaster128, line 24: Do we really want this return type? Woudn't it be sufficient and simpler if we return the && combined response? What do you think, @ethanfrey?

orkunkl: The return result `[true, true, false]` does not have precise information about which message has crashed the tx. Instead of `Vec<bool>` I would recommend returning a `StdResult<ContractErr>`

orkunkl (approved):
Only one comment. LGTM

packages/cw1/src/query.rs
webmaster128: > The return result [true, true, false] does not have precise information about which message

Oh, it does. The order of the results is the same as the order of submitted messages.

> simply returning a StdResult<ContractErr>

Such an error is raised when trying to execute the messages. For this query both _ok_ and _not ok_ should not be error cases.

maurolacy: > Do we really want this return type?

It's useful for testing, as you can see below. Besides that, I'm OK with both options... I'm not sure how useful this really can be for writing contracts / clients.

maurolacy: Another (related) thing, is that that struct can be moved to the cw1 package (it's currently being defined in each contract). If you want I can do that here or in another PR

webmaster128: Which structure? `CanExecuteResponse` is in `packages/cw1/src/query.rs`

maurolacy: You're right... when making the changes I was left under the impression `query.rs` was defined in both cw1 contracts.

ethanfrey:
Since making queries is rather cheap and we now have a rather awkward `Vec<CosmosMsg>` -> `Vec<bool>` API, I wonder if this is actually useful. I am fine with the rename, but wonder about this API change.

webmaster128:
It is also a question how inteligent the query API is expected to be. What happens if you pass a bunch of bank send messages that are fine individually but together exceed the allowance.

What's the purpose of this API at all?

ethanfrey:
The general idea was a generic way to query "can I do this?" without relying on the specifics (allowances/permissions only in cw1-subkeys). And yeah, it was not designed to be smart and handle messages interacting.

I don't know if there will be other approaches than cw1-subkeys. But this was a way to generically query "am I allowed to send 100 Atoms?" before doing so (which could help in generic UIs).

This can be removed and added later. Or just not supported in clients. I don't see adding more complexity here as beneficial.

(I almost consider making `execute` only take `CosmosMsg`, not `Vec<CosmosMsg>`, but that may be too restrictive.

maurolacy:
> It is also a question how inteligent the query API is expected to be. What happens if you pass a bunch of bank send messages that are fine individually but together exceed the allowance.

Good point. Since the checks are stateful, it seems to me that to support this API we would need to reproduce the state changes to be able to validate the sequence of messages. Which may be difficult to do.
That, or we just support one message per query, as of now.

> 
> What's the purpose/use case of this API at all?

Tests look nicer / simpler, but I agree, that's not  enough of a use case to justify this.

maurolacy:
Not merging because impl not stateful when checking multiple messages. Better to provide simpler, clearer API.

Issue #141, opened 2020-11-13
Multisig handles changes to group membership


ethanfrey:
Builds on #80,  #139 and #140

Let's build a properly flexible multisig that is robust to changes in the underlying group at any time. I will propose some approaches here, all of them require some sort of event notification on group changes - #140

----

Simple but limited: (please implement one of these two first to ensure the system works)

* If an "update group" event comes in while there is an open proposal, return error - this will block the group changes until there are no open proposals
* If an "update group" event comes in while there is an open proposal, force close all open proposals (aborting them). This ensures the new proposal will have a stable voting group (maybe close them and auto-reopen?)

Note that these (as well as the ones below) need a quick way to find all currently open proposals - maybe a secondary index? or clever use of the primary index

------

Complex: (Let's discuss these in the comments to decide)

*Count all votes with state at beginning of voting period*

- The total weight needed to pass is fixed at the beginning and never changes
- If a member is updated and they haven't voted yet, record the previous weight (the value at the beginning of the proposal) in some local cache inside `cw3-flex-multisig`. When the member votes, the weight there overrides the current weight in the group. 

*Count all votes with state at end of voting period*

- If a member is updated and they already voted, update the weight, both in their vote and the sum inside the proposal
- The total weight needed to pass will update with each change to the group
- If a member is updated and they haven't voted yet, no need to do anything special

ethanfrey:
@alpe Also asking for your feedback here. Not about code, but about the design choice. I know you implemented this for the governance/voting contracts in weave. And it kept multiple versions of the voting set.

Of the two complex solutions proposed:

* Count all votes with state at beginning of voting period
* Count all votes with state at end of voting period

Which one do you prefer and why? (I forget what thinking we ended up with before, and maybe you preferences have changed).

Also in reference to the other requested review: https://github.com/CosmWasm/cosmwasm-plus/issues/140#issuecomment-738416246 we see how the first "simple" solution depends on this being a hook (not a listener) and explicitly altering the behavior of the group contract based on our business logic.

alpe:
When using the state at "beginning of voting" it has the benefit that all numbers are set and transparent. You may possibly end an election early when a threshold is reached or can not be reached anymore. This is not an option though if votes can be changed within the voting period.
"Beginning of voting" is also closer to what I consider "fair". The outcome of the election can only be influenced by votes and not by buying more weight during the voting period. 
[...]

PR #150, opened 2020-12-03
Separate multisig from group

### LINKED ISSUE #80: Separate Groups from Multisigs
Builds on #8 and #79 

Aim:

Often we will want to use one group of actors in two different multisigs. Some actions may need 50% (to send tokens), others may need 67% (to add new members). Rather than try to keep the groups of two different multisigs in sync, we can create a group contract that just maintains a set of voters and has one contract that can update it.

The multisig contracts can then query this group contract when they want to check if the person is eligible to vote and how many votes they have. 

-----

Details:

Create a new contract called `cw3-flex-multisig`.  This works like `cw3-fixed-multisig`, except it does not store group membership inside. Instead it gets an address of a `cw4-group` contract (actually any `cw4` compatible contract) on init. Upon starting a proposal, query the total weight from the group to properly initialize the proposal. And upon each vote, check membership via the `cw4` contract. You can use [`cw4/helpers.rs`](https://github.com/CosmWasm/cosmwasm-plus/blob/master/packages/cw4/src/helpers.rs#L83-L113) to provide these calls.

You can no longer use the test cases as written in `cw3-fixed-multisig` as they require a mock querier to call into other contracts.  Please use https://github.com/CosmWasm/cosmwasm-plus/pull/137 as a reference how to write such tests and do feel free to simplify the API. In the end, you should be able to port over most of the `cw3-fixed-multisig` logic, so let's see how we can make multi-tests simple enough to set up - maybe best to just copy it and then start trimming it out.
 
*Note: We assume the group is static - or at least does not change during an open proposal. That adds a lot of complexity and we will address that in a later issue*

ethanfrey:
Closes #80

This no longer stores the members locally (in contract), but uses raw queries to read them from an associated group.

The code changes were relatively small. The tests must be completely re-written to use multi-test. Notably, this was even a couple lines of code shorter after the refactor. And that using multiple contracts in a much more complicated test harness.

In the end, this works the same as the `cw3-fixed-multisig`, but opens the door for all the other mutlisig tickets

The first two commits are quite huge and just make a clone of cw3-fixed-multisig as a new contract. To review, you can just check all commits after the first two, and make sure to expand `contract.rs`, which is where most of the changes are: https://github.com/CosmWasm/cosmwasm-plus/pull/150/files/2a70c8baea7dfe534f136cff1bfa20b3485128fa..HEAD

### DIFF
--- a/contracts/cw3-flex-multisig/src/contract.rs
+++ b/contracts/cw3-flex-multisig/src/contract.rs
@@ -11,12 +11,12 @@ use cw3::{
     ProposalListResponse, ProposalResponse, Status, ThresholdResponse, Vote, VoteInfo,
     VoteListResponse, VoteResponse, VoterListResponse, VoterResponse,
 };
+use cw4::Cw4Contract;
 use cw_storage_plus::Bound;
 
 use crate::error::ContractError;
 use crate::msg::{HandleMsg, InitMsg, QueryMsg};
 use crate::state::{next_id, parse_id, Ballot, Config, Proposal, BALLOTS, CONFIG, PROPOSALS};
-use cw4::Cw4Contract;
 
 // version info for migration info
 const CONTRACT_NAME: &str = "crates.io:cw3-flex-multisig";
@@ -445,18 +445,6 @@ mod tests {
     use cw_multi_test::{App, Contract, ContractWrapper, SimpleBank};
 
     use super::*;
-
-    // fn mock_env_height(height_delta: u64) -> Env {
-    //     let mut env = mock_env();
-    //     env.block.height += height_delta;
-    //     env
-    // }
-    //
-    // fn mock_env_time(time_delta: u64) -> Env {
-    //     let mut env = mock_env();
-    //     env.block.time += time_delta;
-    //     env
-    // }
 
     const OWNER: &str = "admin0001";
     const VOTER1: &str = "voter0001";
--- a/contracts/cw3-flex-multisig/src/contract.rs
+++ b/contracts/cw3-flex-multisig/src/contract.rs
@@ -15,12 +15,11 @@ use cw_storage_plus::Bound;
 
 use crate::error::ContractError;
 use crate::msg::{HandleMsg, InitMsg, QueryMsg};
-use crate::state::{
-    next_id, parse_id, Ballot, Config, Proposal, BALLOTS, CONFIG, PROPOSALS, VOTERS,
-};
+use crate::state::{next_id, parse_id, Ballot, Config, Proposal, BALLOTS, CONFIG, PROPOSALS};
+use cw4::Cw4Contract;
 
 // version info for migration info
-const CONTRACT_NAME: &str = "crates.io:cw3-fixed-multisig";
+const CONTRACT_NAME: &str = "crates.io:cw3-flex-multisig";
 const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
 
 pub fn init(
@@ -32,10 +31,13 @@ pub fn init(
     if msg.required_weight == 0 {
         return Err(ContractError::ZeroWeight {});
     }
-    if msg.voters.is_empty() {
-        return Err(ContractError::NoVoters {});
+    // we just convert to canonical to check if this is a valid format
+    if deps.api.canonical_address(&msg.group).is_err() {
+        return Err(ContractError::InvalidGroup {});
     }
-    let total_weight = msg.voters.iter().map(|v| v.weight).sum();
+
+    let group = Cw4Contract(msg.group);
+    let total_weight = group.total_weight(&deps.querier)?;
 
     if total_weight < msg.required_weight {
         return Err(ContractError::UnreachableWeight {});
@@ -45,16 +47,11 @@ pub fn init(
 
     let cfg = Config {
         required_weight: msg.required_weight,
-        total_weight,
         max_voting_period: msg.max_voting_period,
+        group,
     };
     CONFIG.save(deps.storage, &cfg)?;
 
-    // add all voters
-    for voter in msg.voters.iter() {
-        let key = deps.api.canonical_address(&voter.addr)?;
-        VOTERS.save(deps.storage, &key, &voter.weight)?;
-    }
     Ok(InitResponse::default())
 }
 
@@ -89,11 +86,12 @@ pub fn handle_propose(
 ) -> Result<HandleResponse<Empty>, ContractError> {
     // only members of the multisig can create a proposal
     let raw_sender = deps.api.canonical_address(&info.sender)?;
-    let vote_power = VOTERS
-        .may_load(deps.storage, &raw_sender)?
-        .ok_or_else(|| ContractError::Unauthorized {})?;
-
     let cfg = CONFIG.load(deps.storage)?;
+
+    let vote_power = cfg
+        .group
+        .is_member(&deps.querier, raw_sender)?
+        .ok_or_else(|| ContractError::Unauthorized {})?;
 
     // max expires also used as default
     let max_expires = cfg.max_voting_period.after(&env.block);
@@ -152,8 +150,11 @@ pub fn handle_vote(
 ) -> Result<HandleResponse<Empty>, ContractError> {
     // only members of the multisig can vote
     let raw_sender = deps.api.canonical_address(&info.sender)?;
-    let vote_power = VOTERS
-        .may_load(deps.storage, &raw_sender)?
+    let cfg = CONFIG.load(deps.storage)?;
+
+    let vote_power = cfg
+        .group
+        .is_member(&deps.querier, raw_sender)?
         .ok_or_else(|| ContractError::Unauthorized {})?;
 
     // ensure proposal exists and can be voted on
@@ -291,9 +292,10 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
 
 fn query_threshold(deps: Deps) -> StdResult<ThresholdResponse> {
     let cfg = CONFIG.load(deps.storage)?;
+    let total_weight = cfg.group.total_weight(&deps.querier)?;
     Ok(ThresholdResponse::AbsoluteCount {
         weight_needed: cfg.required_weight,
-        total_weight: cfg.total_weight,
+        total_weight,
     })
 }
 
@@ -401,10 +403,13 @@ fn list_votes(
 }
 
 fn query_voter(deps: Deps, voter: HumanAddr) -> StdResult<VoterResponse> {
+    let cfg = CONFIG.load(deps.storage)?;
     let voter_raw = deps.api.canonical_address(&voter)?;
-    let weight = VOTERS
-        .may_load(deps.storage, &voter_raw)?
+    let weight = cfg
+        .group
+        .is_member(&deps.querier, voter_raw)?
         .unwrap_or_default();
+
     Ok(VoterResponse {
         addr: voter,
         weight,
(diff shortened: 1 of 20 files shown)


contracts/cw3-flex-multisig/src/contract.rs
```diff
@@ -0,0 +1,993 @@
...
+    }
+    // we just convert to canonical to check if this is a valid format
+    if deps.api.canonical_address(&msg.group).is_err() {
+        return Err(ContractError::InvalidGroup {});
+    }
+
+    let group = Cw4Contract(msg.group);
+    let total_weight = group.total_weight(&deps.querier)?;
+
+    if total_weight < msg.required_weight {
```
maurolacy, line 42: Given that `total_weight` is now dynamic, does this check still makes sense?

contracts/cw3-flex-multisig/src/contract.rs
```diff
@@ -0,0 +1,993 @@
...
+        })
+        .collect();
+
+    Ok(VoteListResponse { votes: votes? })
+}
+
+fn query_voter(deps: Deps, voter: HumanAddr) -> StdResult<VoterResponse> {
+    let cfg = CONFIG.load(deps.storage)?;
+    let voter_raw = deps.api.canonical_address(&voter)?;
+    let weight = cfg
```
maurolacy, line 408: I think this is a (minor) issue with the fixed-multisig too: given that zero weights are allowed, this query cannot distiguish between a valid (in the list) "voter" with zero weight, and a non-voter (not in the list).

contracts/cw3-flex-multisig/src/contract.rs
```diff
@@ -0,0 +1,993 @@
...
+        .group
+        .is_member(&deps.querier, &voter_raw)?
+        .unwrap_or_default();
```
maurolacy, line 411: I can't help but mention here that I love Rust. This level of expressiveness and power at the same time.

contracts/cw3-flex-multisig/src/contract.rs
```diff
@@ -0,0 +1,993 @@
...
+    deps: Deps,
+    start_after: Option<HumanAddr>,
+    limit: Option<u32>,
+) -> StdResult<VoterListResponse> {
+    let cfg = CONFIG.load(deps.storage)?;
+    let voters = cfg
+        .group
+        .list_members(&deps.querier, start_after, limit)?
+        .into_iter()
+        .map(|member| VoterResponse {
```
maurolacy, line 429: Probably not worth it to simplify, but `Member` and `VoterResponse` are basically the same.

maurolacy (approved):
This is pretty good. I am amazed at how using the right abstractions simplifies the code and increases readability.
I was expecting cross-contract queries to be much more of an issue. This is as clear a build-up on top of fixed-multisig as possible.

Will review the tests later, and add single comments in any case.

contracts/cw3-flex-multisig/src/contract.rs
ethanfrey: I think it does, this will change in #141 as well, as we would use more percentages than fixed values. But creating a multisig that cannot be used is a user error.

contracts/cw3-flex-multisig/src/contract.rs
ethanfrey: We actually return `Option<u64>` internally (from the group) to make this distinction, which allows the 0 weight OWNER to propose (nice idea of yours). I think your point is that we do not expose that information on the query. 

That is a good point, I can update the return type to `Option<u64>`, which would translate to `number?` in typescript (maybe null or undefined)

contracts/cw3-flex-multisig/src/contract.rs
ethanfrey: They are the same. They are defined in two different specs, cw3 and cw4.

Maybe I should combine them into one shared type (and then make an explicit dependency between the two specs). This can be another issue, and if you think it simplifies the code, happy if you make such a PR. I wonder if the Rust compiler will reduce this `.map()` into a no-op, realizing the two types have the same byte layout.

contracts/cw3-flex-multisig/src/msg.rs
```diff
@@ -0,0 +1,73 @@
...
+use serde::{Deserialize, Serialize};
+
+use cosmwasm_std::{CosmosMsg, Empty, HumanAddr};
+use cw0::{Duration, Expiration};
+use cw3::Vote;
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
+pub struct InitMsg {
+    // this is the group contract that contains the member list
+    pub group: HumanAddr,
```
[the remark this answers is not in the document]
ethanfrey: Done

contracts/cw3-flex-multisig/src/state.rs
```diff
@@ -0,0 +1,79 @@
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use std::convert::TryInto;
+
+use cosmwasm_std::{BlockInfo, CosmosMsg, Empty, StdError, StdResult, Storage};
+
+use cw0::{Duration, Expiration};
+use cw3::{Status, Vote};
+use cw4::Cw4Contract;
```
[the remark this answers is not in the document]
ethanfrey: Can you explain please? We have an open issue to improve naming (happy if you want to make a clear proposal), but the functionality should be correct. Cw4Contract is basically a human addr (same serialization), but contains all the helpers to access the contract. It works well for me here (eg `cfg.group.is_member(addr)`).

contracts/cw3-flex-multisig/src/state.rs
```diff
@@ -0,0 +1,79 @@
...
+use cw_storage_plus::{Item, Map, U64Key};
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
+pub struct Config {
+    pub required_weight: u64,
+    pub max_voting_period: Duration,
+    // Total weight and voters are queried from this contract
+    pub group: Cw4Contract,
```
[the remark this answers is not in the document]
ethanfrey: done

contracts/cw3-flex-multisig/src/state.rs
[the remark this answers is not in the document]
maurolacy: I mean, just having a wrapper called `ContractAddr` (even when it's just a wrapper on top of `HumanAddr`, or even `String`) helps with code readability, clarity, and correctness.

This wrapper could have some generic helper methods (`.canonical_addr()`, probably not much more). `Cw4Contract`  (or better, `Cw4ContractAddr`) can in turn derive from it.

ethanfrey:
I think I have addressed all outstanding issues. Just waiting for a review of the test code to merge this.

contracts/cw3-flex-multisig/src/contract.rs
maurolacy: Good idea, returning an `Option` to signal this. I was thinking on returning -1, but that's not possible (`u64`), and slightly confusing.

contracts/cw3-flex-multisig/src/contract.rs
maurolacy: I guess we would need to dissassemble to know.

Regarding combining the types, I'll create the issue, and work on a PR to see if it's worth it.

contracts/cw3-flex-multisig/src/state.rs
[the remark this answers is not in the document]
maurolacy: Just wondering, maybe we can use type aliases for that? Just to improve readability.

contracts/cw3-flex-multisig/src/contract.rs
maurolacy: See #151.

contracts/cw3-flex-multisig/src/state.rs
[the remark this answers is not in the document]
maurolacy: I understand the `Human` vs `Canonical` is about the format (i. e. "human-**readable**" address). But still, it's confusing to me as I tend to relate `HumanAddr` to an address associated to a human. That's why I'm proposing a "contract address" type alias.

Maybe `HumanAddr` should be just called `Addr`. It's maybe too late to change it, though.

contracts/cw3-flex-multisig/src/contract.rs
```diff
@@ -468,230 +454,285 @@ mod tests {
...
-        assert!(res.is_err());
-        match res.unwrap_err() {
-            ContractError::WrongExpiration {} => {}
-            e => panic!("unexpected error: {}", e),
-        }
+        let res = app.execute_contract(OWNER, &flex_addr, &proposal_wrong_exp, &[]);
+        assert_eq!(
+            res.unwrap_err(),
+            ContractError::WrongExpiration {}.to_string()
+        );
```
maurolacy, line 659: Nice pattern.

contracts/cw3-flex-multisig/src/contract.rs
```diff
@@ -468,230 +454,285 @@ mod tests {
...
         let required_weight = 4;
         let voting_period = Duration::Time(2000000);
-
-        let info = mock_info(OWNER, &[]);
-        setup_test_case(deps.as_mut(), info.clone(), required_weight, voting_period).unwrap();
+        let flex_addr = setup_test_case(&mut app, required_weight, voting_period, coins(10, "BTC"));
 
         let bank_msg = BankMsg::Send {
-            from_address: OWNER.into(),
+            from_address: flex_addr.clone(),
```
maurolacy, line 632: If I understand correctly, this is not (not necessarily) wrong in the original (fixed-multisig) code. It's just another way to reward the recipient. Better, in a sense, as the contract must have the funds already assigned, instead of depending on a third party for them.

I wonder what would happen if a proposal is passed, contract releases the bank messages, but the sender (a third party, or the contract iself) lacks enough funds.

A sanity check in multisig contracts code, only accepting proposals if the contract is funded enough to eventually release the associated bank messages, would be a good idea here. Of course, I think this is only possible in the current scenario, that is, when the contract addr is in the `from_addr` of the `BankMsg`s.

maurolacy: If you are OK with it, I'll create an issue to add this check to both multisig contracts.

ethanfrey: If the contract passes a proposal to send from the owners account, the bank module will reject it.

I wanted to remove the from field after some confusion as it is currently only supported when from is the contract itself (if it can move funds from your account, that is a major security hole).

That is a bug in the fixed contract, I can fix that as well

ethanfrey: It is very hard in general to validate if any message will pass. Even more if it will be valid at the end of the voting period.

I would not do any such checks in the contract, but encourage any UI to do so

ethanfrey: Currently what would happen is execute fails. If it was sent from the contract and there was insufficient funds, you could try to execute later if more funds arrive

maurolacy: OK then. You're right, that is a bug / confusion in the original multisig test code.

contracts/cw3-flex-multisig/src/contract.rs
```diff
@@ -468,230 +454,285 @@ mod tests {
...
+        assert_eq!(
+            voters.voters,
+            vec![VoterResponse {
+                addr: OWNER.into(),
+                weight: 1
+            }]
+        );
     }
 
     // TODO: query() tests
```
maurolacy, line 621: This TODO can probably be removed, as there are query tests above, and below.

contracts/cw3-flex-multisig/src/contract.rs
maurolacy: > I wanted to remove the from field after some confusion as it is currently only supported when from is the contract itself (if it can move funds from your account, that is a major security hole).

I think that is a good idea. It will remove a whole class of bugs, and simplify setup for use cases, tests, etc.

In fact, allowing a message payload in the handle functions is probably not a good idea in general. Better for the contract to build the messages, and pass only the required / variable fields to the handle.

maurolacy:
Nice translation of all the tests!

contracts/cw3-flex-multisig/src/contract.rs
maurolacy: > (if it can move funds from your account, that is a major security hole).

Of course. I get confused between the owner / creator / admin address, and the contract address itself.

contracts/cw3-flex-multisig/src/contract.rs
ethanfrey: Made issue #154 for this

PR #157, opened 2020-12-04
Close proposal on membership change

ethanfrey:
Based on #156 (merge that first)
Addresses #141 (just the simple solution to show hooks, need a second PR to close issue)

- [x] Register callback hooks in all tests
- [x] Implement member changed hook handler
- [x] Test proper handling of simple solution: *If an "update group" event comes in while there is an open proposal, force close all open proposals*
- [x] Improve implementation efficiency (don't iterate over all proposals ever made to find open proposals)

We just need to optimize the query (add secondary index to proposals map), otherwise this is ready to review

### DIFF
--- a/contracts/cw3-flex-multisig/src/contract.rs
+++ b/contracts/cw3-flex-multisig/src/contract.rs
@@ -1051,17 +1051,31 @@ mod tests {
 
         // VOTER3 starts a proposal to send some tokens
         let cash_proposal = pay_somebody_proposal(&flex_addr);
-        let _ = app
+        let res = app
             .execute_contract(VOTER3, &flex_addr, &cash_proposal, &[])
             .unwrap();
         // Get the proposal id from the logs
-        let _cash_proposal_id: u64 = res.attributes[2].value.parse().unwrap();
+        let cash_proposal_id: u64 = res.attributes[2].value.parse().unwrap();
+        assert_ne!(cash_proposal_id, update_proposal_id);
+
+        // query state
+        let query_proposal = QueryMsg::Proposal {
+            proposal_id: cash_proposal_id,
+        };
+        let prop: ProposalResponse = app
+            .wrap()
+            .query_wasm_smart(&flex_addr, &query_proposal)
+            .unwrap();
+        assert_eq!(prop.status, Status::Open);
 
         // ensure VOTER3 is currently a member
-        let query = QueryMsg::Voter {
+        let query_voter = QueryMsg::Voter {
             address: VOTER3.into(),
         };
-        let power: VoterResponse = app.wrap().query_wasm_smart(&flex_addr, &query).unwrap();
+        let power: VoterResponse = app
+            .wrap()
+            .query_wasm_smart(&flex_addr, &query_voter)
+            .unwrap();
         assert_eq!(power.weight, 3);
 
         // Pass and execute first proposal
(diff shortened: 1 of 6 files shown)


contracts/cw3-flex-multisig/src/contract.rs
```diff
@@ -624,30 +695,27 @@ mod tests {
...
-        };
+        let proposal = pay_somebody_proposal(&flex_addr);
         let res = app.execute_contract(SOMEBODY, &flex_addr, &proposal, &[]);
         assert_eq!(res.unwrap_err(), ContractError::Unauthorized {}.to_string());
 
         // Wrong expiration option fails
+        let msgs = match pay_somebody_proposal(&flex_addr) {
+            HandleMsg::Propose { msgs, .. } => msgs,
+            _ => panic!("Wrong variant"),
+        };
```
maurolacy, line 714: Re-usability :-)

contracts/cw3-flex-multisig/src/contract.rs
```diff
@@ -624,30 +695,27 @@ mod tests {
...
-            description: "Do we reward her?".to_string(),
-            msgs: msgs.clone(),
-            latest: None,
-        };
+        let proposal = pay_somebody_proposal(&flex_addr);
         let res = app.execute_contract(SOMEBODY, &flex_addr, &proposal, &[]);
         assert_eq!(res.unwrap_err(), ContractError::Unauthorized {}.to_string());
 
         // Wrong expiration option fails
+        let msgs = match pay_somebody_proposal(&flex_addr) {
```
maurolacy, line 711:
```suggestion
        let msgs = match proposal.clone() {
```

contracts/cw3-flex-multisig/src/contract.rs
```diff
@@ -990,4 +1037,125 @@ mod tests {
...
+            .update_members(vec![VOTER3.into()], vec![])
+            .unwrap();
+        let update_proposal = HandleMsg::Propose {
+            title: "Kick out VOTER3".to_string(),
+            description: "He's trying to steal our money".to_string(),
+            msgs: vec![update_msg],
+            latest: None,
+        };
+        let res = app
+            .execute_contract(VOTER1, &flex_addr, &update_proposal, &[])
```
maurolacy, line 1068: Nice, using the multisig to update group memberships. But, this is unnecessarily complex. I would leave this, as it's a nice test, but write another (simpler) test (or add a TODO for now) before this one, that just does the basic thing:

- Creates a group and associated multisig.
- Install the hook.
- Check that the multisig is Open.
- Change group memership, directly against the group addr.
- Check that the multisig is now Rejected / Aborted.

maurolacy (approved):
lgtm.

contracts/cw3-flex-multisig/src/contract.rs
```diff
@@ -268,6 +271,38 @@ pub fn handle_close(
...
+        .range(deps.storage, None, None, Order::Ascending)
+        .filter(|item| match item {
+            Ok((_, prop)) => prop.status == Status::Open,
+            Err(_) => true,
+        })
+        .collect();
+
+    // close all open proposals
+    for (k, mut prop) in open? {
+        prop.status = Status::Rejected;
```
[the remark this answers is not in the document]
ethanfrey: I would do that, but I will change this in a day or two. This is just an intermediate pr to test hooks, before doing the more complex solution

ethanfrey:
> It just occurred to me that it would be nice for flex multisigs to request the hook directly to their group, on creation. But, now that I think of it, this has security concerns... I wonder if this is related to the "listeners" functionality... it would require something like a "publish & subscribe" contract / functionality.

Yeah, I assume an external administrator to set up all contracts. Most of these steps need admin access anyway and it is hard to write robust init scripts as contracts.

The "listener", light weight pub-sub solution that runs outside of the TX would open up a lot of doors. But that begs a lot of questions. Notably, who pays the gas and where we enforce limits

maurolacy:
> The "listener", light weight pub-sub solution that runs outside of the TX would open up a lot of doors. But that begs a lot of questions. Notably, who pays the gas and where we enforce limits

Yes... besides gas, I can think of timeouts, retries, blacklists, whitelists, evictions / purging, priorities, ...

PR #158, opened 2020-12-07
Weight 0 vs not member

### LINKED ISSUE #154: Distinguish between weight 0 and not member in cw3 queries
`Cw3QueryMsg::Voter{addr: HumanAddr}` should return something like `cw4::MemberResponse`, eg:

```rust
pub struct VoterInfo {
  pub weight: Option<u64>,
}
```

Do not change the `VoterListResponse` return info.

maurolacy:
Closes #154.


ethanfrey (approved):
Lgtm

PR #159, opened 2020-12-07
Weight 0 vs not member

### LINKED ISSUE #154: Distinguish between weight 0 and not member in cw3 queries

maurolacy:
Now, closes #154.

### DIFF
--- a/contracts/cw3-fixed-multisig/schema/query_msg.json
+++ b/contracts/cw3-fixed-multisig/schema/query_msg.json
@@ -163,7 +163,7 @@
       }
     },
     {
-      "description": "Returns VoterResponse",
+      "description": "Returns VoterInfo",
       "type": "object",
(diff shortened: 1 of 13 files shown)


ethanfrey (commented):
VoterInfo looks good.

But VoterResponse still has the Option<u64> weight. It shouldn't be optional

maurolacy:
Oh, I see, good catch. I reverted it in my branch, but it came back from the rebase. Shoud be fixed now.

ethanfrey (approved):
:+1:

PR #160, opened 2020-12-08
cw3-flex-multisig uses voting power from a snapshot of the block the proposal opened

### LINKED ISSUE #141: Multisig handles changes to group membership

ethanfrey:
Closes #141 

Uses the preferred "complex" solution. We no longer close open proposals when the group changes, but handle it gracefully

*Count all votes with state at beginning of voting period*

- The total weight needed to pass is fixed at the beginning and never changes
- If a member is updated and they haven't voted yet, record the previous weight (the value at the beginning of the proposal) in some local cache inside `cw3-flex-multisig`. When the member votes, the weight there overrides the current weight in the group. 

We do this with max 1 write per updated weight, only if this was the first time it has changed since the most recent proposal. -> O(diffs), regardless of how many proposals there are.

Future improvement: We currently iterate over all open proposals to find the latest start height. With a clever compound secondary index, we could do this with one read. 

TODO: 

- [x] Rename MemberDiff fields (old, new)
- [x] Test voting on second proposal doesn't use snapshot
- [x] Test complex case
- [x] Improve performance

### DIFF
--- a/contracts/cw3-flex-multisig/schema/handle_msg.json
+++ b/contracts/cw3-flex-multisig/schema/handle_msg.json
@@ -295,13 +295,13 @@
       "description": "MemberDiff shows the old and new states for a given cw4 member They cannot both be None. old = None, new = Some -> Insert old = Some, new = Some -> Update old = Some, new = None -> Delete",
       "type": "object",
       "required": [
-        "addr"
+        "key"
       ],
       "properties": {
-        "addr": {
+        "key": {
           "$ref": "#/definitions/HumanAddr"
         },
-        "new_weight": {
+        "new": {
           "type": [
             "integer",
             "null"
@@ -309,7 +309,7 @@
           "format": "uint64",
           "minimum": 0.0
         },
-        "old_weight": {
+        "old": {
           "type": [
             "integer",
             "null"
--- a/contracts/cw3-flex-multisig/src/contract.rs
+++ b/contracts/cw3-flex-multisig/src/contract.rs
@@ -1069,7 +1069,7 @@ mod tests {
             .unwrap();
         // Get the proposal id from the logs
         let proposal_id: u64 = res.attributes[2].value.parse().unwrap();
-        let prop_state = |app: &App| -> ProposalResponse {
+        let prop_state = |app: &App, proposal_id: u64| -> ProposalResponse {
             let query_prop = QueryMsg::Proposal { proposal_id };
             app.wrap()
                 .query_wasm_smart(&flex_addr, &query_prop)
@@ -1077,7 +1077,7 @@ mod tests {
         };
 
         // 1/4 votes
-        assert_eq!(prop_state(&app).status, Status::Open);
+        assert_eq!(prop_state(&app, proposal_id).status, Status::Open);
 
         // a few blocks later...
         app.update_block(|block| block.height += 2);
@@ -1105,19 +1105,36 @@ mod tests {
         assert_eq!(power.weight, None);
 
         // proposal still open
-        assert_eq!(prop_state(&app).status, Status::Open);
+        assert_eq!(prop_state(&app, proposal_id).status, Status::Open);
 
         // a few blocks later...
         app.update_block(|block| block.height += 3);
 
-        // VOTER2 can only vote with weight of 2 (not enough to pass)
+        // make a second proposal
+        let proposal2 = pay_somebody_proposal(&flex_addr);
+        let res = app
+            .execute_contract(VOTER1, &flex_addr, &proposal2, &[])
+            .unwrap();
+        // Get the proposal id from the logs
+        let proposal_id2: u64 = res.attributes[2].value.parse().unwrap();
+
+        // VOTER2 can pass this alone with the updated vote (newer height ignores snapshot)
+        let yes_vote = HandleMsg::Vote {
+            proposal_id: proposal_id2,
+            vote: Vote::Yes,
+        };
+        app.execute_contract(VOTER2, &flex_addr, &yes_vote, &[])
+            .unwrap();
+        assert_eq!(prop_state(&app, proposal_id2).status, Status::Passed);
+
+        // VOTER2 can only vote on first proposal with weight of 2 (not enough to pass)
         let yes_vote = HandleMsg::Vote {
             proposal_id,
             vote: Vote::Yes,
         };
         app.execute_contract(VOTER2, &flex_addr, &yes_vote, &[])
             .unwrap();
-        assert_eq!(prop_state(&app).status, Status::Open);
+        assert_eq!(prop_state(&app, proposal_id).status, Status::Open);
 
         // newbie cannot vote
         let err = app
@@ -1128,7 +1145,7 @@ mod tests {
         // previously removed VOTER3 can still vote, passing the proposal
         app.execute_contract(VOTER3, &flex_addr, &yes_vote, &[])
             .unwrap();
-        assert_eq!(prop_state(&app).status, Status::Passed);
+        assert_eq!(prop_state(&app, proposal_id).status, Status::Passed);
     }
 
     // Currently this just closes all open proposals
--- a/contracts/cw3-flex-multisig/src/contract.rs
+++ b/contracts/cw3-flex-multisig/src/contract.rs
@@ -18,7 +18,7 @@ use crate::error::ContractError;
 use crate::msg::{HandleMsg, InitMsg, QueryMsg};
 use crate::snapshot::{snapshot_diff, snapshoted_weight};
 use crate::state::{
-    next_id, parse_id, proposals, status_index, Ballot, Config, Proposal, BALLOTS, CONFIG,
+    max_proposal_height, next_id, parse_id, proposals, Ballot, Config, Proposal, BALLOTS, CONFIG,
 };
 
 // version info for migration info
@@ -292,20 +292,7 @@ pub fn handle_membership_hook(
     }
 
     // find the latest snapshot height
-    // TODO: this is O(open proposals), with clever composite secondary indexes (status, height), we may get this to O(1)
-    let heights: StdResult<Vec<u64>> = proposals()
-        .idx
-        .status
-        .items(
-            deps.storage,
-            &status_index(Status::Open),
-            None,
-            None,
-            Order::Ascending,
-        )
-        .map(|res| Ok(res?.1.start_height))
-        .collect();
-    let max_height = heights?.into_iter().max();
+    let max_height = max_proposal_height(deps.storage)?;
 
     // only try snapshot if there is an open proposal
     if let Some(last_height) = max_height {
--- a/contracts/cw3-flex-multisig/src/contract.rs
+++ b/contracts/cw3-flex-multisig/src/contract.rs
@@ -1047,8 +1047,7 @@ mod tests {
         assert_eq!(err, ContractError::WrongCloseStatus {}.to_string());
     }
 
-    // Currently this just closes all open proposals
-    // TODO: something more clever (lazily tracking voting power at start of election)
+    // use the power from the beginning of the voting period
     #[test]
     fn handle_group_changes_from_external() {
         let mut app = mock_app();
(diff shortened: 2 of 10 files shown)


maurolacy:
Hi, I'll review it now, so this may be asked in the code, but anyway, I have a couple questions:

>     * If a member is updated and they haven't voted yet, record the previous weight (the value at the beginning of the proposal) in some local cache inside `cw3-flex-multisig`. When the member votes, the weight there overrides the current weight in the group.

What about addition of new members? Can they vote?
What about deletion of members that existed when the proposal was created?

contracts/cw3-flex-multisig/schema/handle_msg.json
```diff
@@ -295,21 +295,21 @@
       "description": "MemberDiff shows the old and new states for a given cw4 member They cannot both be None. old = None, new = Some -> Insert old = Some, new = Some -> Update old = Some, new = None -> Delete",
       "type": "object",
       "required": [
-        "addr"
+        "key"
```
maurolacy, line 298: Why the renaming? no strong feelings, but I for one like the old name better.

ethanfrey:
I should probably add more code comments. The test covers both the new and deleted cases.

In any case, use the voting power (or lack thereof) at the block height when the proposal was created.

It doesn't snapshot per tx, but at the beginning of the block in which the proposal was created, which should be fine for any reasonable case

contracts/cw3-flex-multisig/schema/handle_msg.json
```diff
@@ -295,21 +295,21 @@
...
       ],
       "properties": {
-        "addr": {
+        "key": {
           "$ref": "#/definitions/HumanAddr"
         },
-        "new_weight": {
+        "new": {
```
maurolacy, line 304: Same here.

maurolacy (commented):
Seeing and thinking about the concept of snapshots, it occurred to me that all that logic could be implemented at the level of the **group** contract directly.
It would be more efficient (implemented once, and available for all clients of the group).
It would be enough to extend the query API to allow for a (optional) date / height parameter. If that parameter is not supplied, the most recent / updated value is returned.

What do you think?

contracts/cw3-flex-multisig/src/state.rs
```diff
@@ -91,15 +126,17 @@ impl<'a> IndexList<Proposal> for ProposalIndexes<'a> {
...
+    idx.extend_from_slice(&height.to_be_bytes());
+    idx
 }
 
 // secondary indexes on state for PROPOSALS to find all open proposals efficiently
 pub fn proposals<'a>() -> IndexedMap<'a, U64Key, Proposal, ProposalIndexes<'a>> {
     let indexes = ProposalIndexes {
         status: MultiIndex::new(
-            |p| status_index(p.status),
+            |p| status_height_index(p.status, p.start_height),
```
maurolacy, line 139: If I understand correctly, you want to query all Open proposals after a given height, right? So you will know which proposals need to use the new weight for an addr, vs. the old one.

Maybe I'm missing something, but, why not do it the other way around?

When a group update for an addr comes (which you are subscribed, because of the hook), store the addr (as key) and the current block height, along with the original weight (values) in a map. Something like original_weights or so. All this is O(1).

When voting, check if the voter is in the original_weights map (O(1)), and if it is, check if the stored block height is greater than the start_height of the proposal. If that's the case, use the stored original weight (i.e. ignore the update).
In any other case, query the group contract for the weight.

Makes sense?

maurolacy: What I'm probably missing is: multiple updates over the same addr.

In that case, it would be good to use an index, `original_weights` per `block_height`.
That way, we'll have one index per (updated) addr, where we can efficiently query the original value before a given block height (which would be the `start_height` of each Open proposal).

Given that block heights are monotonically increasing, an array and binary search will do. But an index is probably better / nicer.

contracts/cw3-flex-multisig/src/snapshot.rs
```diff
@@ -0,0 +1,61 @@
+use cosmwasm_std::{CanonicalAddr, Deps, DepsMut, Order, StdResult, Storage};
+use cw3::VoterInfo;
+use cw4::{Cw4Contract, MemberDiff};
+use cw_storage_plus::{Bound, Map, U64Key};
+
+/// SNAPSHOTS pk: (HUmanAddr, height) -> get the weight
```
maurolacy, line 6:
```suggestion
/// SNAPSHOTS pk: (HumanAddr, height) -> get the weight
```

contracts/cw3-flex-multisig/src/snapshot.rs
```diff
@@ -0,0 +1,61 @@
...
+    let start = Bound::inclusive(U64Key::new(height));
+    let first = SNAPSHOTS
+        .prefix(&addr)
+        .range(storage, Some(start), None, Order::Ascending)
+        .next();
+    match first {
+        None => Ok(None),
+        Some(r) => r.map(|(_, v)| Some(v)),
+    }
+}
```
maurolacy, line 61: I see now this is precisely what I'm proposing below. :-)

I started reviewing `state.rs`, and was confused by the block height in the proposals index.

contracts/cw3-flex-multisig/src/contract.rs
```diff
@@ -173,6 +170,15 @@ pub fn handle_vote(
...
     }
 
+    // use a snapshot of "start of proposal" if available, otherwise, current group weight
+    let vote_power = snapshoted_weight(
+        deps.as_ref(),
+        &raw_sender,
+        prop.start_height,
+        &cfg.group_addr,
+    )?
+    .ok_or_else(|| ContractError::Unauthorized {})?;
```
maurolacy, line 180: Yes, this makes sense.

contracts/cw3-flex-multisig/src/state.rs
maurolacy: Update: I see now that this functionality in the index is just an optimization, to avoid creating the snapshots if there is no currently open proposal. Cool.

contracts/cw3-flex-multisig/src/snapshot.rs
```diff
@@ -0,0 +1,61 @@
...
+        None => SNAPSHOTS.save(
+            deps.storage,
+            (&raw_addr, current_height.into()),
+            &VoterInfo { weight: diff.old },
+        ),
+    }
+}
+
+/// this will look for the first snapshot of the given address >= given height
+/// if none, there is no snapshot since that time, and the group query will give the current status
```
maurolacy, line 46:
```suggestion
/// if none, there is no snapshot since that time.
```
Comment is confusing. Group query usage / need better documented elsewere, i.e. outside of this function.

contracts/cw3-flex-multisig/src/snapshot.rs
```diff
@@ -0,0 +1,61 @@
...
+    let snapshot = load_snapshot(deps.storage, &addr, height)?;
+    match snapshot {
+        // use snapshot if available
+        Some(info) => Ok(info.weight),
+        // otherwise load from the group
+        None => group.is_member(&deps.querier, &addr),
+    }
+}
+
+/// saves this diff only if no updates have been saved since the latest snapshot
```
maurolacy, line 27: Why? I don't understand the need for this.

maurolacy (approved):
Looks good to me.

Probably my biggest concern would be for this functionality to be moved to the group contract entirely.

contracts/cw3-flex-multisig/src/state.rs
ethanfrey: Yup, I could write every update, but if we have token weighted voting, a stake may change 30 times during an open proposal,and it will only store the last one

contracts/cw3-flex-multisig/schema/handle_msg.json
ethanfrey: I did this as I plan to make Diff generic over T so it can be used by a generic snapshot provider that will allow this pattern to be easily imported, whenever we want to keep a snapshot

ethanfrey:
> Probably my biggest concern would be for this functionality to be moved to the group contract entirely.

This is a good question. Below is my reasoning (I thought it was more efficient)

I want to pull out the snapshot function in a simple module that can be imported and reused by various contracts. The question is implement in the receiver or in the producer

If producer (group), the receiver (multisig) needs to inform the producer every time a proposal opens or closes, so it tracks needed changes. And the queries could not be raw queries (as it uses iterator to find relevant update). Thus making queries much more expensive that before this pr (raw query) or now (local state + raw query).

It would however, remove the need for the external hook pattern, which would save gas on updates.

With the hooks, you add N messages to every state change, but the queries are cheap. I guess it depends if we assume there are more state changes or more proposals/votes.

ethanfrey:
Let's discuss this design.

BTW, I wanted to make diff+snapshot design an easily importable module to reuse it. Whether that is using hooks, or stored in the producer is an open question.

Let's discuss it a bit more, but my thought is.
1. Merge this pr, do following as separate steps.
2. Refactor this out so core diff / snapshot logic is in a common library 
3. Make a pr not using hooks, but rather snapshots in the producer (ideally minimal code changes needed), then we can compare the two approaches. And also reference that pr and discussion in any design doc - to show the two patterns

ethanfrey:
Also, I think there will be more different cw4 implementations than composable cw3 ones (based on external groups), but I may be wrong.

maurolacy:
> I want to pull out the snapshot function in a simple module that can be imported and reused by various contracts. The question is implement in the receiver or in the producer
> 

That can be done, and is independen of this question.

> If producer (group), the receiver (multisig) needs to inform the producer every time a proposal opens or closes, so it tracks needed changes. And the queries could not be raw queries (as it uses iterator to find relevant update). Thus making queries much more expensive that before this pr (raw query) or now (local state + raw query).

If implemented in producer, I'll do it independently of any receiver. The group contract can have a flag on creation (or even after creation, with an API call); `enable_snapshots` or similar.

> 
> It would however, remove the need for the external hook pattern, which would save gas on updates.

Yes. That, and centralization of impl would be the biggest advantages, as I see it.

> 
> With the hooks, you add N messages to every state change, but the queries are cheap. I guess it depends if we assume there are more state changes or more proposals/votes.

I was supposing updates are relatively rare, and we would in general have more votes than voters's changes.

maurolacy:
> Let's discuss this design.
> 
> BTW, I wanted to make diff+snapshot design an easily importable module to reuse it. Whether that is using hooks, or stored in the producer is an open question.

Yes, exactly.

> Let's discuss it a bit more, but my thought is.
> 
>     1. Merge this pr, do following as separate steps.
> 
>     2. Refactor this out so core diff / snapshot logic is in a common library
> 
>     3. Make a pr not using hooks, but rather snapshots in the producer (ideally minimal code changes needed), then we can compare the two approaches. And also reference that pr and discussion in any design doc - to show the two patterns

Sounds great. Let's do it.

ethanfrey:
Thanks for the review and feedback on the architecture. 

You propose keeping a snapshot of all data, not just just those needed for open proposals?

ethanfrey:
I like the idea of the enable snapshot flag on init

maurolacy:
> You propose keeping a snapshot of all data, not just just those needed for open proposals?

Yes. Mostly for simplicity. It can even support setting an expiration height / time, i. e. "generate snapshot up until this block / date".

contracts/cw3-flex-multisig/src/snapshot.rs
ethanfrey: Optimization

PR #164, opened 2020-12-09
Snapshot module

### LINKED ISSUE #161: Refactor snapshotting into reusable module
Follow up on #160 

Make this a `storage-plus` wrapper around a Map. (Bonus: ideally around indexed map as well)

Configure:
* Checkpoint every block
* Checkpoint only selected blocks (with add/remove checkpoint calls)
* Checkpoint nothing (this is just the wrapped type)

We don't really need to implement the last case, but allows us to use the same type and switch behavior based on InitMsg - as discussed here: https://github.com/CosmWasm/cosmwasm-plus/pull/160#issuecomment-741740324

ethanfrey:
Closes #161 

Make a SnapshotMap in `storage-plus` that handles this automatically for us.

- [x] Implement base
- [x] Implement snapshot every block
- [x] Implement snapshot selected blocks
- [x] Tests

### DIFF
--- a/packages/storage-plus/src/snapshot.rs
+++ b/packages/storage-plus/src/snapshot.rs
@@ -1,0 +1,144 @@
+use serde::de::DeserializeOwned;
+use serde::{Deserialize, Serialize};
+
+use cosmwasm_std::{StdError, StdResult, Storage};
+
+use crate::keys::{PrimaryKey, U64Key};
+use crate::map::Map;
+use crate::path::Path;
+#[cfg(feature = "iterator")]
+use crate::prefix::Prefix;
+
+/// Map that maintains a snapshots of one or more checkpoints
+pub struct SnapshotMap<'a, K, T> {
+    primary: Map<'a, K, T>,
+
+    // maps height to number of checkpoints (only used for selected)
+    checkpoints: Map<'a, U64Key, u32>,
+
+    // this stores all changes (key, height). Must differentiate between no data written,
+    // and explicit None (just inserted)
+    changelog: Map<'a, (K, U64Key), ChangeSet<T>>,
+
+    // TODO: currently only support Never
+    strategy: Strategy,
+}
+
+#[derive(Clone, Copy, PartialEq, Debug, Serialize, Deserialize)]
+pub enum Strategy {
+    EveryBlock,
+    Never,
+    Selected,
+}
+
+impl<'a, K, T> SnapshotMap<'a, K, T> {
+    /// Usage: SnapshotMap::new(snapshot_names!("foobar"), Strategy::EveryBlock)
+    pub const fn new(namespaces: SnapshotNamespaces<'a>, strategy: Strategy) -> Self {
+        SnapshotMap {
+            primary: Map::new(namespaces.pk),
+            checkpoints: Map::new(namespaces.checkpoints),
+            changelog: Map::new(namespaces.changelog),
+            strategy,
+        }
+    }
+}
+
+impl<'a, K, T> SnapshotMap<'a, K, T>
+where
+    T: Serialize + DeserializeOwned,
+    K: PrimaryKey<'a>,
+{
+    pub fn key(&self, k: K) -> Path<T> {
+        self.primary.key(k)
+    }
+
+    #[cfg(feature = "iterator")]
+    pub fn prefix(&self, p: K::Prefix) -> Prefix<T> {
+        self.primary.prefix(p)
+    }
+
+    pub fn save(&self, store: &mut dyn Storage, k: K, data: &T, height: u64) -> StdResult<()> {
+        // TODO: check strategy
+        unimplemented!();
+        // self.key(k).save(store, data)
+    }
+
+    pub fn remove(&self, store: &mut dyn Storage, k: K, height: u64) {
+        // TODO: check strategy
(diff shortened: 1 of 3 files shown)


packages/storage-plus/src/snapshot.rs
```diff
@@ -0,0 +1,371 @@
...
+                }
+                // otherwise, we don't save this
+                Ok(false)
+            }
+        }
+    }
+
+    /// load old value and store changelog
+    fn write_change(&self, store: &mut dyn Storage, k: K, height: u64) -> StdResult<()> {
+        let old = self.may_load(store, k.clone())?;
```
maurolacy, line 120:
```suggestion
        let old = self.primary.may_load(store, k.clone())?;
```
for clarity and consistency.

packages/storage-plus/src/snapshot.rs
```diff
@@ -0,0 +1,371 @@
...
+    pub fn prefix(&self, p: K::Prefix) -> Prefix<T> {
+        self.primary.prefix(p)
+    }
+
+    /// should_checkpoint looks at the strategy and determines if we want to checkpoint
+    fn should_checkpoint(&self, store: &dyn Storage, k: &K) -> StdResult<bool> {
+        match self.strategy {
+            Strategy::EveryBlock => Ok(true),
+            Strategy::Never => Ok(false),
+            Strategy::Selected => {
```
maurolacy, line 91: Interesting. A little bit tricky / delicate to use, but can be useful to save some bytes.

packages/storage-plus/src/snapshot.rs
```diff
@@ -0,0 +1,371 @@
...
+            .prefix(k.clone())
+            .range(store, Some(start), None, Order::Ascending)
+            .next();
+
+        if let Some(r) = first {
+            // if we found a match, return this last one
+            r.map(|(_, v)| v.old)
+        } else {
+            // otherwise, return current value
+            self.may_load(store, k)
```
maurolacy, line 174: What about signalling / exposing this difference? Particularly for `Selected`, it can be useful. At least for debugging.
Maybe by providing an alternative method, `is_correct()` or so, not to spoil the return value here.

packages/storage-plus/src/snapshot.rs
```diff
@@ -0,0 +1,371 @@
...
+        k: K,
+        height: u64,
+        action: A,
+    ) -> Result<T, E>
+    where
+        A: FnOnce(Option<T>) -> Result<T, E>,
+        E: From<StdError>,
+    {
+        let input = self.may_load(store, k.clone())?;
+        let diff = ChangeSet { old: input.clone() };
```
maurolacy, line 196: This one can be computed inside the `should_checkpoint()` block above.

maurolacy (approved):
Looks good to me. Will take a detailed look at the tests during the afternoon.

Next step / variant: timestamp based snapshots.

maurolacy:
> Next step / variant: timestamp based snapshots.

If the sub-key is the same type / size, this is already working, by the way.

ethanfrey:
Thanks for the feedback, I will do some requested cleanup

> Next step / variant: timestamp based snapshots.

Yeah, you have to store both separately, and I find height a much more natural (and guaranteed unique) measure than timestamp. There may be 3 blocks in the same second for example.

I will leave that as an excercise for the reader.

packages/storage-plus/src/snapshot.rs
ethanfrey: When the data is not available, I can either return possibly incorrect data, or return an error.

I think you are correct that an error is better.

That applies to any historical query for `Selected::Never`

packages/storage-plus/src/snapshot.rs
ethanfrey: I have to clone the `input` here (as we consume it in `action`). But you are right, `ChangeSet` can be done below.

packages/storage-plus/src/snapshot.rs
ethanfrey: I addressed this, it down returns `StdError::NotFound` if there is no relevant checkpoint

ethanfrey:
Fixed up change requests. I will merge when CI is happy.

#165 builds on top and is another chance to review the module in practice.

packages/storage-plus/src/snapshot.rs
maurolacy: It is my understanding that it may apply to `Selected` as well: If you're using `Selected`, and didn't create / request a checkpoint for that specific height, you cannot tell for sure if the returned value is correct. It _may_ be correct, but it also be an updated value, which didn't get snapshotted at that height.

maurolacy: What I'm proposing is returning the value nevertheless, but return also a flag (or, maybe better, provide another method) to indicate if the value is guaranteed to be correct for that height, or not.

maurolacy: I've seen you have implemented returning an error when the value is not guaranteed to be correct.

I prefer it, as it's less error prone: if you are using `Selected` and didn't get the guaranteed values for your heights, you're probably using it wrong. Same if you're not using snapshots at all, and pretend to get some historical values.

PR #165, opened 2020-12-10
Use snapshots in cw4-group

### LINKED ISSUE #162: Don't use hooks for snapshotting on cw3-cw4 interface
Builds on #160 and #161

Just add this as a query on the group.

ethanfrey:
Closes #162 

I think this implements the version you preferred. Snapshotting in cw4-group and not relying on the hooks.

It does seem simpler (now that I put all the work into #161). Happy for your feedback or suggestions how to improve it.

I considered a new PR to investigate allowing only selected checkpoints for less writes. But this doesn't work in cw4-group, because who do we trust to control these? By removing old snapshots, you could invalidate a proposal in progress...


maurolacy:
Reviewing this one now.

packages/storage-plus/src/snapshot.rs
```diff
@@ -269,6 +269,7 @@ pub struct SnapshotNamespaces<'a> {
 #[macro_export]
 macro_rules! snapshot_names {
     ($var:expr) => {
+        #[allow(clippy::string_lit_as_bytes)]
```
maurolacy, line 272:
```suggestion
#[allow(clippy::string_lit_as_bytes)]
macro_rules! snapshot_names {
    ($var:expr) => {
```

packages/storage-plus/src/snapshot.rs
```diff
@@ -284,10 +285,15 @@ mod tests {
 
     #[test]
     fn namespace_macro() {
-        let names = snapshot_names!("demo");
-        assert_eq!(names.pk, b"demo");
-        assert_eq!(names.checkpoints, b"demo__checkpoints");
-        assert_eq!(names.changelog, b"demo__changelog");
```
maurolacy, line 290: This works if you put the clippy allow before the macro def (see above).

maurolacy: Better said, this works for me, and without the allow macro. I'm using rust 1.48.0 stable, though.

packages/cw4/src/query.rs
```diff
@@ -18,7 +18,10 @@ pub enum Cw4QueryMsg {
         limit: Option<u32>,
     },
     /// Returns MemberResponse
-    Member { addr: HumanAddr },
+    Member {
+        addr: HumanAddr,
+        height: Option<u64>,
```
maurolacy, line 23: Maybe `at_height`? Not sure about it.

contracts/cw4-group/src/state.rs
```diff
@@ -1,10 +1,14 @@
...
+use cw4::TOTAL_KEY;
+use cw_storage_plus::{snapshot_names, Item, SnapshotMap, SnapshotNamespaces, Strategy};
 
 pub const ADMIN: Item<Option<CanonicalAddr>> = Item::new(b"admin");
 pub const TOTAL: Item<u64> = Item::new(TOTAL_KEY);
-pub const MEMBERS: Map<&[u8], u64> = Map::new(MEMBERS_KEY);
+
+// Note: this must be same as cw4::MEMBERS_KEY but macro needs literal, not const
+pub const MEMBERS: SnapshotMap<&[u8], u64> =
+    SnapshotMap::new(snapshot_names!("members"), Strategy::EveryBlock);
```
maurolacy, line 10: helper `fn` instead of macro? (I hate macros) :-)

maurolacy (commented):
Looks pretty good. Nice how "subclassing" of `SnapshotMap` from `Map` made the code changes straightforward here.

I also think this version is better / simpler, as it requires less interactions.

maurolacy (approved):

packages/storage-plus/src/snapshot.rs
maurolacy: You migh want to take a look at this, as I've now tried with 1.47.0, and it works. After removing the clippy allow, which throws an `error[E0658]: attributes on expressions are experimental`.

contracts/cw4-group/src/state.rs
ethanfrey: Problem is I want `&'static [u8]`
No way I see to do with functions. But if you can write one, I am happy to accept that pr

ethanfrey: I have also avoided macros strongly. This is the first time I wrote one

maurolacy: I see. Not possible without eliminating the references inside `SnapshotNamespaces`.

ethanfrey: I could take it as 3 different arguments instead of the struct, but they end up in Map and have to be `static so we can keep it as a global const

packages/storage-plus/src/snapshot.rs
ethanfrey: I tried with 1.47.0.

The clippy problem never was with this line of code so much, but rather with compiling `cw4-group`. I only get the error there (as it is in non-test code)

maurolacy: Ah, OK. Will take a look tomorrow.

contracts/cw4-group/src/state.rs
maurolacy: Yes... I was playing with this and it's not so much that it can't be done in a function (it can, using `Vec` instead of slices) but that `Map::new()` is marked `const` and so ir requires `const` parameters.

PR #167, opened 2020-12-10
Token weighted group

### LINKED ISSUE #142: Add token-weighted group
Create another implementation of `cw4` (besides `cw4-group`) called `cw4-stake`.

This contract will implement all the same `cw4` queries, both [smart and raw](https://github.com/CosmWasm/cosmwasm-plus/blob/master/packages/cw4/src/helpers.rs#L69-L124). It will error on `UpdateMembers` message, but will send the same "on update" events as defined in #140 (Maybe we should remove `HandleMsg::UpdateMembers` from `cw4` and leave that only in `cw4-group`?).

It works as follows:

* Init defines a native denom, as well as a ratio (tokens to votes - eg. if we want to normalize to 1 ATOM = 1 vote, not 1 uatom, we make this ratio 1.000.000). It also defines an unbonding period (cw0::Duration)
* I can `bond` tokens, sending them with the message. This will update the "bonded tokens" in my account and update my votes based on the ratio
* I can `unbond` tokens, specifying how many and removing that many "bonded tokens" from my account, placing them as "unbonding tokens" with an expiration time equal to current block plus "unbonding period".
* I can `claim` any previously unbonding tokens that have now hit their expiration time, by destroying those claims and releasing the relevant native tokens to my address.

Note: Much of this bonding/unbonding logic has been implemented as [bond/unbond/claim in `cw20-staking`](https://github.com/CosmWasm/cosmwasm-plus/blob/master/contracts/cw20-staking/src/contract.rs#L166-L367). That is a bit different as the ratio per share is dynamic and it does something with the bonded tokens, but you can use a slightly simplified version of that logic.

Bonus: If you pull this whole bonding/unbonding logic and state into one file that could be easily reused by other contracts

Test this group by itself, as well as backing a `cw3-flex-multisig` (before #141 we just test case where membership is static during open proposals)

ethanfrey:
Closes #142


ethanfrey:
This should be complete

maurolacy:
Currently reviewing.

contracts/cw4-group/src/state.rs
```diff
@@ -1,18 +1,13 @@
-use cosmwasm_std::{CanonicalAddr, HumanAddr};
+use cosmwasm_std::CanonicalAddr;
 use cw4::TOTAL_KEY;
 use cw_storage_plus::{Item, SnapshotMap, Strategy};
 
 pub const ADMIN: Item<Option<CanonicalAddr>> = Item::new("admin");
 pub const TOTAL: Item<u64> = Item::new(TOTAL_KEY);
 
-// Note: this must be same as cw4::MEMBERS_KEY but macro needs literal, not const
```
maurolacy, line 8: Nice.

contracts/cw4-group/src/helpers.rs
```diff
@@ -0,0 +1,62 @@
...
+impl Cw4GroupContract {
+    pub fn new(addr: HumanAddr) -> Self {
+        Cw4GroupContract(Cw4Contract(addr))
+    }
+
+    /// Convert this address to a form fit for storage
+    pub fn canonical(&self, api: &dyn Api) -> StdResult<Cw4GroupCanonicalContract> {
+        let canon = api.canonical_address(&self.addr())?;
+        Ok(Cw4GroupCanonicalContract(canon))
+    }
```
maurolacy, line 34: I don't understand why the wrappers are needed. Why not return a `CanonicalAddr` directly?

Why not a `to_canonical()` method in `HumanAddr`?

contracts/cw4-group/src/helpers.rs
```diff
@@ -0,0 +1,62 @@
...
+use cw4::{Cw4Contract, Member};
+
+use crate::msg::HandleMsg;
+
+/// Cw4GroupContract is a wrapper around HumanAddr that provides a lot of helpers
+/// for working with cw4-group contracts.
+///
+/// It extends Cw4Contract to add the extra calls from cw4-group.
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct Cw4GroupContract(pub Cw4Contract);
```
maurolacy, line 15: Is this nesting really needed? / useful?

contracts/cw4-group/src/helpers.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+/// This is a representation of Cw4GroupContract for storage.
+/// Don't use it directly, just translate to the Cw4GroupContract when needed.
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct Cw4GroupCanonicalContract(pub CanonicalAddr);
+
+impl Cw4GroupCanonicalContract {
+    /// Convert this address to a form fit for usage in messages and queries
+    pub fn human(&self, api: &dyn Api) -> StdResult<Cw4GroupContract> {
```
maurolacy, line 58: Why not a `to_human()` in `CanonicalAddr` directly? Why are specific structs / wrappers needed in each case?

contracts/cw4-group/src/error.rs
```diff
@@ -1,17 +1,16 @@
...
+use cw0::hooks::HookError;
+
 #[derive(Error, Debug)]
 pub enum ContractError {
     #[error("{0}")]
     Std(#[from] StdError),
 
+    #[error("{0}")]
+    Hook(#[from] HookError),
+
```
maurolacy, line 13: Nice.

contracts/cw4-group/src/contract.rs
```diff
@@ -98,12 +101,12 @@ pub fn handle_update_members(
     // make the local update
     let diff = update_members(deps.branch(), env.block.height, info.sender, add, remove)?;
     // call all registered hooks
-    let mut ctx = Context::new();
-    for h in HOOKS.may_load(deps.storage)?.unwrap_or_default() {
-        let msg = diff.clone().into_cosmos_msg(h)?;
-        ctx.add_message(msg);
-    }
-    Ok(ctx.into())
+    let messages = prepare_hooks(deps.storage, |h| diff.clone().into_cosmos_msg(h))?;
```
maurolacy, line 104: Clever. I guess `clone()` is cheap here. Plus. a new message must be created nevertheless. Plus, hooks are normaly a few.

contracts/cw3-flex-multisig/src/contract.rs
```diff
@@ -559,7 +560,7 @@ mod tests {
 
         // 3. (Optional) Set the multisig as the group owner
         if multisig_as_group_admin {
-            let update_admin = Cw4HandleMsg::UpdateAdmin {
```
maurolacy, line 562: Beware that this message still exists. Better to use the one froom the spec, I think.

maurolacy (commented):
This is a review of all the changes in the PR, except for the `cw4-stake` ones.

Releasing it now so you can take a look at it. Will specifically review `cw4-stake` in a follow-up.

packages/cw0/src/hooks.rs
```diff
@@ -0,0 +1,51 @@
+use cosmwasm_std::{CosmosMsg, HumanAddr, StdError, StdResult, Storage};
+use thiserror::Error;
+
+use cw_storage_plus::Item;
+
+// store all hook addresses in one item. We cannot have many of them before the contract becomes unusable anyway.
+pub const HOOKS: Item<Vec<HumanAddr>> = Item::new("hooks");
```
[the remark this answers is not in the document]
ethanfrey: Yeah, as a common import. Each contract has it's own names.

I think I want to refactor this to let each contract pick a different string (not just "hooks"), but thought to do that in another PR (or maybe this?)

contracts/cw4-group/README.md
```diff
@@ -39,5 +39,19 @@ decision-making.
...
 
-Update messages and queries are defined by the 
-[cw4 spec](../../packages/cw4/README.md). Please refer to it for more info.
\ No newline at end of file
+Basic update messages, queries, and hooks are defined by the 
+[cw4 spec](../../packages/cw4/README.md). Please refer to it for more info.
+
+`cw4-group` adds two messages:
+
+`UpdateAdmin{admin}` - changes (or clears) the admin for the contract
```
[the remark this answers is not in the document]
ethanfrey: I moved this back and forth a bit. I will adjust. Good eye.

contracts/cw4-group/README.md
```diff
@@ -39,5 +39,19 @@ decision-making.
...
+
+`UpdateMembers{add, remove}` - takes a membership diff and adds/updates the
+members, as well as removing any provided addresses. If an address is on both
+lists, it will be removed. If it appears multiple times in `add`, only the
+last occurance will be used.
```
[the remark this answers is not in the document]
ethanfrey: The cw4 is about a group that can be queried.

We add messages about admin and hook, as other contracts will need to register themselves with those hooks in many cases. But we should not specify how the group itself is managed.

`UpdateMembers` by the admin is just one case. Update on stake is other. I plan a 3rd implementation that merges two groups with some function. Only one implementation supports `UpdateMembers`, so I see no need for it in the generic spec.

contracts/cw4-group/src/helpers.rs
ethanfrey: > Why not a to_canonical() method in HumanAddr?

Because it requires the `deps.api` to do the transformation. We could add it but it would not be any simpler.

 > I don't understand why the wrappers are needed. Why not return a CanonicalAddr directly?

I was storing everything as `CanonicalAddr` in state. I see even the sdk is moving to persist as bech32 strings, and maybe we can start removing this canonical transform at some point.

contracts/cw4-group/src/helpers.rs
ethanfrey: There is a `Deref` so we "inherit" all the query functionality of `Cw4Contract`. That saves us from re-implementing that logic every step.

contracts/cw4-group/src/helpers.rs
ethanfrey: Again, `Cw4GroupCanonicalContract` may be obsolete.

However the human <-> canonical transform is blockchain dependent and based on deps.api.

This came from the initial design when we wanted CosmWasm to support a large variety of blockchains as possible (incl. substrate, avalance, eth2, etc) and saw funding as integrating with blockchains. Turns out, no one wants to pay much, so we turn to focus on one specific runtime - Cosmos SDK Stargate

contracts/cw3-flex-multisig/src/contract.rs
ethanfrey: Good point

ethanfrey:
Thanks for the feedback, I will do the general cleanup. Happy to see your review on `cw4-stake` when you get to it.

contracts/cw4-stake/src/contract.rs
```diff
@@ -0,0 +1,1020 @@
...
+}
+
+pub fn handle_claim(
+    deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+) -> Result<HandleResponse, ContractError> {
+    let sender_raw = deps.api.canonical_address(&info.sender)?;
+    let release = claim_tokens(deps.storage, &sender_raw, &env.block, None)?;
+    if release == Uint128(0) {
```
maurolacy, line 217:
```suggestion
    if release.is_zero() {
```

maurolacy (approved):
Looks good to me. Will take a detailed look of the tests later, and add single comments.

contracts/cw4-stake/README.md
```diff
@@ -0,0 +1,58 @@
...
+Members are defined by an address and a weight. This is transformed
+and stored under their `CanonicalAddr`, in a format defined in
+[cw4 raw queries](../../packages/cw4/README.md#raw).
+
+Note that 0 *is an allowed weight*. This doesn't give any voting rights, but
+it does define this address is part of the group. This could be used in
+eg. a KYC whitelist to say they are allowed, but cannot participate in
+decision-making.
+
+If `min_bond` is higher than `tokens_per_weight`, you cannot have anyone with 0 weight.
```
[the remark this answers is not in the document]
maurolacy: Maybe it's a good idea to explain / detail that.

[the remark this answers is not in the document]
ethanfrey: Exactly.
I should use your wording instead, it sounds clearer

contracts/cw4-group/src/helpers.rs
maurolacy: > There is a `Deref` so we "inherit" all the query functionality of `Cw4Contract`. That saves us from re-implementing that logic every step.

Nice, but the logic is not much. Maybe it can be implemented below directly.

contracts/cw4-stake/src/contract.rs
```diff
@@ -0,0 +1,1020 @@
...
+    deps: DepsMut,
+    info: MessageInfo,
+    new_admin: Option<HumanAddr>,
+) -> Result<HandleResponse, ContractError> {
+    update_admin(deps, info.sender, new_admin)?;
+    Ok(HandleResponse::default())
+}
+
+// the logic from handle_update_admin extracted for easier import
+pub fn update_admin(
```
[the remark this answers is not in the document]
ethanfrey: I hadn't thought of it. The import was mainly for the tests...

But yeah, I could make a general admin controller, this pattern is used in lots of places

packages/cw0/src/hooks.rs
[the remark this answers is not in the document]
maurolacy: I think it's OK as a general name.

contracts/cw4-stake/src/contract.rs
```diff
@@ -0,0 +1,1020 @@
...
+    let cfg = CONFIG.load(deps.storage)?;
+
+    // ensure the sent denom was proper
+    // NOTE: those clones are not needed (if we move denom, we return early),
+    // but the compiler cannot see that
+    let sent = match info.sent_funds.len() {
+        0 => Err(ContractError::MissingDenom(cfg.denom.clone())),
+        1 => {
+            if info.sent_funds[0].denom == cfg.denom {
+                Ok(info.sent_funds[0].amount)
```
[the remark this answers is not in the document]
ethanfrey: good point

contracts/cw4-stake/src/contract.rs
```diff
@@ -0,0 +1,1020 @@
...
+    storage: &mut dyn Storage,
+    sender: HumanAddr,
+    sender_raw: &CanonicalAddr,
+    new_stake: Uint128,
+    cfg: &Config,
+    height: u64,
+) -> StdResult<Vec<CosmosMsg>> {
+    // update their membership weight
+    let new = calc_weight(new_stake, cfg);
+    let old = MEMBERS.may_load(storage, sender_raw)?;
```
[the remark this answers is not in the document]
ethanfrey: You are right, then we short-circuit and do not write the weight nor send a diff :100:

ethanfrey:
Nice comments, thank you for the thorough review

contracts/cw4-group/src/helpers.rs
maurolacy: > > Why not a to_canonical() method in HumanAddr?
> 
> Because it requires the `deps.api` to do the transformation. We could add it but it would not be any simpler.

Yeah, you're right, passing the api to the method doesn't look any better.

> 
> > I don't understand why the wrappers are needed. Why not return a CanonicalAddr directly?
> 
> I was storing everything as `CanonicalAddr` in state. I see even the sdk is moving to persist as bech32 strings, and maybe we can start removing this canonical transform at some point.

That would be a good thing:tm:.

contracts/cw4-group/src/helpers.rs
maurolacy: I like this "inheritance" property of Tuple structs, by the way.

If I'm not mistaken, you still have to wrap the methods in order to expose them. It would be better to expose the public methods of the nested struct directly. And allow for overloading. But, at least traits are inherited, which is better than in the case of plain composition. Slowly moving to real inheritance... :-)

contracts/cw4-stake/README.md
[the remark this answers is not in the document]
maurolacy: I think leaving both remarks would be OK.

contracts/cw4-group/src/helpers.rs
maurolacy: > There is a `Deref` so we "inherit" all the query functionality of `Cw4Contract`. That saves us from re-implementing that logic every step.

I think I misunderstood you before. What you're saying is that, precisely, with that `Deref` trick we don't need to implement wrappers anymore. Cool, I'll try this. Thanks.

contracts/cw4-stake/src/contract.rs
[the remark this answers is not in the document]
ethanfrey: Made issue #181 for this

contracts/cw4-group/src/helpers.rs
maurolacy: Works like a charm. This is an "unofficial" "anti" pattern called [Deref polymorphism](https://github.com/rust-unofficial/patterns/blob/master/anti_patterns/deref.md), by the way.

I wish all unofficial things were as useful and convenient as this. Please, Rust people, more of these "unofficial" stuff, please.

maurolacy: Poor man's inheritance, I'll call it.

PR #168, opened 2020-12-11
Fix macro test / remove clippy directive

maurolacy:
This works for me locally. Using the clippy directive inside the macro makes the test error with some "experimental feature" msg.
Removing the clippy directive doesn't seem to affect anything... plus, original test work.

Let's see if it passes the CI tests.

### DIFF
--- a/packages/storage-plus/src/snapshot.rs
+++ b/packages/storage-plus/src/snapshot.rs
@@ -269,7 +269,6 @@ pub struct SnapshotNamespaces<'a> {
 #[macro_export]
 macro_rules! snapshot_names {
     ($var:expr) => {
-        #[allow(clippy::string_lit_as_bytes)]
         SnapshotNamespaces {
             pk: $var.as_bytes(),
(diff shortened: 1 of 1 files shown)


maurolacy:
You're right, this requires that weird lambda in the tests.

Sorry but I couldn't reproduce all the sequence locally. I'll check a little more to see if there's a way around this, and if it's not, let's disregard this.

ethanfrey:
The CI gets it. As does running clippy inside `cw4-group`.

I'm happy to remove the macro as well... I spent a while trying to do so, but it required me to set 3 strings in the constructor, which was pretty annoying and I wanted to get rid of.

maurolacy:
OK, I've decided to re-open this.

Here's my take at it: in my opinion, avoiding two more (explicit) static strings, simply doesn't justify the burden of the macro, the linter warnings, the missing experimental features, and the added complexity of it all.

Nevertheless, feel free to disregard this, as it's simpler, but certainly more verbose.

PR #169, opened 2020-12-11
Fix namespace macro test

maurolacy:
"Fixing" the namespaces macro by removing it entirely, at the cost of two more (explicit) static strings.

Names can still be improved a little IMO, but I wanted to introduce as little changes as possible first.

- `SnapshotNamespaces` -> `SnapshotNames`?
- Maybe `SnapshotNamespaces` can be entirely removed (and `SnapshotMap::new()` takes the burden).
- `MEMBERS_CHECK` -> `MEMBERS_CHECKPOINTS`?
- `MEMBERS_CHANGE` -> `MEMBERS_CHANGELOG`?
- Using `&[u8]` instead of `&str` for consistency?

Your call, but I think this is simpler, and clearer. At the cost of being a little bit more verbose.

### DIFF
--- a/packages/cw4/src/query.rs
+++ b/packages/cw4/src/query.rs
@@ -53,7 +53,9 @@ pub struct HooksResponse {
 
 /// TOTAL_KEY is meant for raw queries
 pub const TOTAL_KEY: &[u8] = b"total";
-pub const MEMBERS_KEY: &[u8] = b"members";
+pub const MEMBERS_KEY: &str = "members";
+pub const MEMBERS_CHECK: &str = "members__checkpoints";
+pub const MEMBERS_CHANGE: &str = "members__changelog";
 
 /// member_key is meant for raw queries for one member, given canonical address
 pub fn member_key(address: &[u8]) -> Vec<u8> {
--- a/packages/cw4/src/query.rs
+++ b/packages/cw4/src/query.rs
@@ -54,8 +54,8 @@ pub struct HooksResponse {
 /// TOTAL_KEY is meant for raw queries
 pub const TOTAL_KEY: &[u8] = b"total";
 pub const MEMBERS_KEY: &str = "members";
-pub const MEMBERS_CHECK: &str = "members__checkpoints";
-pub const MEMBERS_CHANGE: &str = "members__changelog";
+pub const MEMBERS_CHECKPOINTS: &str = "members__checkpoints";
+pub const MEMBERS_CHANGELOG: &str = "members__changelog";
 
 /// member_key is meant for raw queries for one member, given canonical address
 pub fn member_key(address: &[u8]) -> Vec<u8> {
--- a/packages/storage-plus/src/snapshot.rs
+++ b/packages/storage-plus/src/snapshot.rs
@@ -42,11 +42,16 @@ pub enum Strategy {
 
 impl<'a, K, T> SnapshotMap<'a, K, T> {
     /// Usage: SnapshotMap::new(snapshot_names!("foobar"), Strategy::EveryBlock)
-    pub const fn new(namespaces: SnapshotNamespaces<'a>, strategy: Strategy) -> Self {
+    pub const fn new(
+        pk: &'a str,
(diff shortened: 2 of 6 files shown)


maurolacy:
Probably for another time and place, but: do we really want / need all those statically referenced slices? They are a pain to deal with. By example here, when you naïvely want to derive some names.

I understand there are issues with not using references: Size not known at compile time, extra space, mutability. Maybe others?. But still.

I know you've fought with this for a much longer time than me, and probably have a much clearer picture. So, I would very much like to hear a harangue supporting (const) statically referenced slices everywhere. :-)

And, complementary to this, what are the disadvantages of using `Vec<u8>` (or even, `String` values) for names / stuff?

ethanfrey:
The reason I use such static slices is that I want to have a const constructor.

This allows:

```rust
pub const MEMBERS: SnapshotMap<&[u8], u64> =
    SnapshotMap::new(snapshot_names!("members"), Strategy::EveryBlock);
```

Before we had to use function constructors and they were a lot more verbose:

```rust
const PREFIX_PERMISSIONS: &[u8] = b"permissions";

/// returns a bucket with all permissions (query by subkey)
pub fn permissions(storage: &mut dyn Storage) -> Bucket<Permissions> {
    bucket(storage, PREFIX_PERMISSIONS)
}

/// returns a bucket with all permissionsk (query by subkey)
/// (read-only version for queries)
pub fn permissions_read(storage: &dyn Storage) -> ReadonlyBucket<Permissions> {
    bucket_read(storage, PREFIX_PERMISSIONS)
}
```

I really want to make the constructors super simple. 

And yes, we could use `&str` rather than `&[u8]` for namespaces (In general in storage-plus). And the name changes make sense

packages/storage-plus/src/snapshot.rs
```diff
@@ -266,40 +266,47 @@ pub struct SnapshotNamespaces<'a> {
...
+        assert_eq!(names.checkpoints, b"demo__checkpoints");
+        assert_eq!(names.changelog, b"demo__changelog");
     }
 
     type TestMap = SnapshotMap<'static, &'static [u8], u64>;
-    const NEVER: TestMap = SnapshotMap::new(snapshot_names!("never"), Strategy::Never);
-    const EVERY: TestMap = SnapshotMap::new(snapshot_names!("every"), Strategy::EveryBlock);
-    const SELECT: TestMap = SnapshotMap::new(snapshot_names!("select"), Strategy::Selected);
+    const NEVER: TestMap = SnapshotMap::new(
+        SnapshotNamespaces::new("never", "never__check", "never__change"),
```
ethanfrey, line 299: Yeah, this is rather more verbose, but I guess the simplicity of not needing a macros in the constructor is helpful

packages/cw4/src/query.rs
```diff
@@ -53,7 +53,9 @@ pub struct HooksResponse {
 
 /// TOTAL_KEY is meant for raw queries
 pub const TOTAL_KEY: &[u8] = b"total";
-pub const MEMBERS_KEY: &[u8] = b"members";
+pub const MEMBERS_KEY: &str = "members";
```
ethanfrey, line 56: Nice to add all keys there

ethanfrey (approved):
This is more verbose, but you are correct, less confusing.

Can you remove the `SnapshotNamespaces` entirely and just use three string arguments in new? Then I would call this good.

Also, I think updating all the namespaces in all the constructors in `storage-plus` to accept `&str` would be nice. We can still store `&[u8]` internally (maybe change that later), but at least the constructor is simpler. (And `&str.as_bytes()` is a `const fn` so we can do this in a const constructor)

maurolacy:
> Also, I think updating all the namespaces in all the constructors in `storage-plus` to accept `&str` would be nice. We can still store `&[u8]` internally (maybe change that later), but at least the constructor is simpler. (And `&str.as_bytes()` is a `const fn` so we can do this in a const constructor)

I'll create an issue for this, and do it after this one.

PR #173, opened 2020-12-11
Replace byte slices by string slices in names and constructors

### LINKED ISSUE #170: Update namespaces / constructors to accept &str
Update all the namespaces in all the constructors in storage-plus to accept `&str`.

We can still store `&[u8]` internally (maybe change that later), but at least the constructor is simpler.
As `&str.as_bytes()` is a `const fn`, we can do this in a `const` constructor.

maurolacy:
Closes #170.

I focused in names & ctors. Possibly other things can be changed as well, in other PR / PRs:

- key(), prefix()
- to_length_prefixed()
- nested_namespace_with_keys()
- load(), update(), save(), may_load(), ...
- Bound
- ...

### DIFF
--- a/contracts/cw721-base/src/state.rs
+++ b/contracts/cw721-base/src/state.rs
@@ -28,12 +28,12 @@ pub struct Approval {
     pub expires: Expiration,
 }
 
-pub const CONTRACT_INFO: Item<ContractInfoResponse> = Item::new(b"nft_info");
-pub const MINTER: Item<CanonicalAddr> = Item::new(b"minter");
-pub const TOKEN_COUNT: Item<u64> = Item::new(b"num_tokens");
+pub const CONTRACT_INFO: Item<ContractInfoResponse> = Item::new("nft_info");
+pub const MINTER: Item<CanonicalAddr> = Item::new("minter");
+pub const TOKEN_COUNT: Item<u64> = Item::new("num_tokens");
 
-// pub const TOKENS: Map<&str, TokenInfo> = Map::new(b"tokens");
-pub const OPERATORS: Map<(&[u8], &[u8]), Expiration> = Map::new(b"operators");
+// pub const TOKENS: Map<&str, TokenInfo> = Map::new("tokens");
+pub const OPERATORS: Map<(&[u8], &[u8]), Expiration> = Map::new("operators");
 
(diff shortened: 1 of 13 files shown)


packages/cw4/src/query.rs
```diff
@@ -52,16 +52,18 @@ pub struct HooksResponse {
...
 pub const MEMBERS_CHECKPOINTS: &str = "members__checkpoints";
 pub const MEMBERS_CHANGELOG: &str = "members__changelog";
 
 /// member_key is meant for raw queries for one member, given canonical address
 pub fn member_key(address: &[u8]) -> Vec<u8> {
-    // length encoded members key (update if you change MEMBERS_KEY)
-    // inlined here to avoid storage-plus import
-    let mut key = b"\x00\x07members".to_vec();
+    // FIXME?: Inlined here to avoid storage-plus import
+    if MEMBERS_KEY.len() > 0xFF {
```
ethanfrey, line 63: nice

contracts/cw721-base/src/state.rs
```diff
@@ -28,12 +28,12 @@ pub struct Approval {
     pub expires: Expiration,
 }
 
-pub const CONTRACT_INFO: Item<ContractInfoResponse> = Item::new(b"nft_info");
-pub const MINTER: Item<CanonicalAddr> = Item::new(b"minter");
-pub const TOKEN_COUNT: Item<u64> = Item::new(b"num_tokens");
+pub const CONTRACT_INFO: Item<ContractInfoResponse> = Item::new("nft_info");
+pub const MINTER: Item<CanonicalAddr> = Item::new("minter");
```
ethanfrey, line 32: So nice to see all these binaries turn into strings

packages/storage-plus/src/item.rs
```diff
@@ -17,9 +17,9 @@ pub struct Item<'a, T> {
 }
 
 impl<'a, T> Item<'a, T> {
-    pub const fn new(storage_key: &'a [u8]) -> Self {
+    pub const fn new(storage_key: &'a str) -> Self {
         Item {
-            storage_key,
+            storage_key: storage_key.as_bytes(),
```
ethanfrey, line 22: Great to keep the internals the same. We will have some binary data (length prefixed stuff), so easiest just to change the constructors for nicer public API

ethanfrey (approved):
Very nice

PR #180, opened 2020-12-14
Add threshold to cw3 flex

### LINKED ISSUE #139: Extend threshold types for multisig
Builds on #80 

Currently the `cw3-fixed-multsig` only supports `AbsoluteCount` threshold, and the initial version of `cw3-flex-multisig` as well. Let us extend `cw3-flex-multisig` to support [all three threshold types defined in cw3](https://github.com/CosmWasm/cosmwasm-plus/blob/master/packages/cw3/src/query.rs#L49-L79).

This should snapshot total_weight at the beginning of the vote, and calculate acceptance criteria based on that. (Note: there may be a subtle difference between the snapshotted member weights - beginning of the block when proposal was created - and the total weight - beginning of the tx when the proposal was created. We can make a follow up issue to see how to resolve that).

These votes (like percentage and quorum) must work properly with significant membership changes during the election.

ethanfrey:
Closes #139 

This allows us to set any of the 3 defined threshold types for passing proposals.
Handler code is working and tested for the original case (`AbsoluteCount`).
Low level code (calculate passing, etc) is tested for all cases I could think of.

TODO:
- [x] unit test all helper functions in msg for the other threshold types
- [x] unit test all helper functions in state for the other threshold types
- [x] revisit rules for passing with quorum (hard to pass early)
- [x] provide high-level scenario tests with other threshold types (esp with changing groups)
- [x] test coverage for proposal lists, vote details


ethanfrey:
This should be ready to review. Covered with lots of unit tests. I will provide a few high-level scenario tests for the more dynamic variants, but the logic should all be covered with tests now.

ethanfrey:
Finished added all tests

contracts/cw3-flex-multisig/src/state.rs
```diff
@@ -45,6 +86,53 @@ impl Proposal {
...
+fn votes_needed(weight: u64, percentage: Decimal) -> u64 {
+    // we multiply by 1million to detect rounding issues
+    const FACTOR: u128 = 1_000_000;
+    let applied = percentage * Uint128(FACTOR * weight as u128);
+    let rounded = (applied.u128() / FACTOR) as u64;
+    if applied.u128() % FACTOR > 0 {
+        rounded + 1
+    } else {
+        rounded
+    }
```
maurolacy, line 135:
```suggestion
    ((applied.u128() + FACTOR - 1) / FACTOR) as u64
```

contracts/cw3-flex-multisig/src/state.rs
```diff
@@ -45,6 +86,53 @@ impl Proposal {
...
+        match self.threshold {
+            Threshold::AbsoluteCount { weight_needed } => self.votes.yes >= weight_needed,
+            Threshold::AbsolutePercentage { percentage_needed } => {
+                self.votes.yes >= votes_needed(self.total_weight, percentage_needed)
+            }
+            Threshold::ThresholdQuora { threshold, quroum } => {
+                // this one is tricky, as we have two compares:
+                if self.expires.is_expired(block) {
+                    // * if we have closed yet, we need quorum% of total votes to have voted (counting abstain)
+                    //   and threshold% of yes votes from those who voted (ignoring abstain)
```
maurolacy, line 108: Why ignoring abstain here, and not also above in `AbsolutePercentage`?

contracts/cw3-flex-multisig/src/state.rs
```diff
@@ -45,6 +86,53 @@ impl Proposal {
...
+                    let total = self.votes.total();
+                    let opinions = total - self.votes.abstain;
+                    total >= votes_needed(self.total_weight, quroum)
+                        && self.votes.yes >= votes_needed(opinions, threshold)
+                } else {
+                    // * if we have not closed yet, we need threshold% of yes votes (from 100% voters - abstain)
+                    //   as we are sure this cannot change with any possible sequence of future votes
```
maurolacy, line 115: Yes, but, we might still miss the quorum, right?

contracts/cw3-flex-multisig/src/state.rs
```diff
@@ -45,6 +86,53 @@ impl Proposal {
...
+                    self.votes.yes
+                        >= votes_needed(self.total_weight - self.votes.abstain, threshold)
```
maurolacy, line 117: Again, discarding abstain here (or not discarding it above) seems inconsistent to me.

maurolacy: I have to think better about this (and probably write some code), but there doesn't need to be a necessity to differentiate between expired or not expired, for knowing if a proposal passed.

If the snapshotted yes votes are greater than a percentage of the initial total weight, and the number of people who voted is greater than a percentage (quorum) of the initial total weight, proposal is passed.

Maybe the distinction is made because of the abstained voters, but, why discount them from the total initial weight?

Too many questions... I propose we discuss tomorrow about this, and then I complete the review.

maurolacy (commented):
Incomplete review. Posting anyway for your consideration.

I propose we have a chat tomorrow to discuss some details I'm certainly missing.

contracts/cw3-flex-multisig/src/state.rs
```diff
@@ -45,6 +86,53 @@ impl Proposal {
...
+            }
+            Threshold::ThresholdQuora { threshold, quroum } => {
+                // this one is tricky, as we have two compares:
+                if self.expires.is_expired(block) {
+                    // * if we have closed yet, we need quorum% of total votes to have voted (counting abstain)
+                    //   and threshold% of yes votes from those who voted (ignoring abstain)
+                    let total = self.votes.total();
+                    let opinions = total - self.votes.abstain;
+                    total >= votes_needed(self.total_weight, quroum)
+                        && self.votes.yes >= votes_needed(opinions, threshold)
```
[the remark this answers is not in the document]
ethanfrey: I thought so too, but then what are abstains?

If we need 50% and 3 yes and 3 no and 1 abstain, that should pass, right?

I will add some docs here. I think I wrote them in the docs but they were a bit scattered and didn't make sense to you

contracts/cw3-flex-multisig/src/state.rs
ethanfrey: Yours is shorter but more confusing as to the purpose - here you definitely need a comment ;)

contracts/cw3-flex-multisig/src/state.rs
ethanfrey: AbsolutePercentage is 50% of the total weight of the group.

In quorum, we care about the ones who vote, as long as enough vote.

contracts/cw3-flex-multisig/src/state.rs
ethanfrey: Good point. I assumed threshold > quorum.
You are right, with 50% threshold and 80% quroum, we may have eg 60% yes votes and not pass...

I will add a check and test for that

contracts/cw3-flex-multisig/src/msg.rs
```diff
@@ -1,19 +1,100 @@
...
+    /// as the first will pass with 60% yes votes and 10% no votes, while the second
+    /// will require the others to vote anything (no, abstain...) to pass
+    AbsolutePercentage { percentage_needed: Decimal },
+    /// Declares a threshold (minimum percentage of votes that must approve)
+    /// and a quorum (minimum percentage of voter weight that must vote).
+    /// This allows eg. 25% of total weight YES to pass, if we have quorum of 40%
+    /// and threshold of 51% and most of the people sit out the election.
+    /// This is more common in general elections where participation is expected
+    /// to be low.
+    ThresholdQuora { threshold: Decimal, quroum: Decimal },
```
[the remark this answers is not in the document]
ethanfrey: Renamed

ethanfrey:
Please review https://github.com/CosmWasm/cosmwasm-plus/pull/188 and give feedback there, then we can see what discussion points are still open.

contracts/cw3-flex-multisig/src/state.rs
ethanfrey: Point here is you can only close early if you assume all missing votes are "no" and you would still pass.

50% threshold and 40% quorum with 20 total weight:

5 yes and 4 no would pass when the proposal is closed (hits 50% threshold of existing voters), but would not be sufficient to close early (what if 3 more nos came in... we cannot call this until we are sure it cannot fail).

We would need 10 yes (and any number of other votes) to call this early.

contracts/cw3-flex-multisig/src/state.rs
```diff
@@ -25,18 +27,57 @@ pub struct Proposal {
...
     pub msgs: Vec<CosmosMsg<Empty>>,
     pub status: Status,
-    /// how many votes have already said yes
-    pub yes_weight: u64,
-    /// how many votes needed to pass
-    pub required_weight: u64,
+    /// pass requirements
+    pub threshold: Threshold,
+    // the total weight when the proposal started (used to calculate percentages)
+    pub total_weight: u64,
```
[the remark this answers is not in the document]
ethanfrey: I think I will skip this one

ethanfrey:
Also I addressed the edge case on hitting absolute threshold, but not quorum that you pointed out. Good catch there.

contracts/cw3-flex-multisig/src/state.rs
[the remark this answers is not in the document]
maurolacy: Makes sense to me. Differentiating people who explicitly abstained, from the ones that didn't vote (yet).
Let's de the same in the other threshold (`AbsolutePercentage`).

contracts/cw3-flex-multisig/src/state.rs
maurolacy: You're right. Done.

contracts/cw3-flex-multisig/src/state.rs
maurolacy: Same rules apply, IMO. Suppose there are 7 voters, absolute percentage is 50%, 3 vote yes, and one abstains. Proposal should pass, right?

OK, I'll read the document, and continue from here.

contracts/cw3-flex-multisig/src/state.rs
```diff
@@ -45,6 +86,57 @@ impl Proposal {
...
+            }
+        }
+    }
+}
+
+// this is a helper function so Decimal works with u64 rather than Uint128
+// also, we must *round up* here, as we need 8, not 7 votes to reach 50% of 15 total
+fn votes_needed(weight: u64, percentage: Decimal) -> u64 {
+    // we multiply by 1million to detect rounding issues
+    const FACTOR: u128 = 1_000_000;
```
maurolacy, line 132: Better moved to a global (config) section? Better naming would be `PRECISION_FACTOR` or so.

I don't want to introduce unnecessary complexity, but this can be probably transformed into an `u32` (or lower) `PRECISION_DIGITS`, and used as a power of 10 exponent.

contracts/cw3-flex-multisig/src/state.rs
```diff
@@ -45,6 +86,57 @@ impl Proposal {
...
+                    total >= votes_needed(self.total_weight, quorum)
+                        && self.votes.yes >= votes_needed(opinions, threshold)
+                } else {
+                    // * if we have not closed yet, we need threshold% of yes votes (from 100% voters - abstain)
+                    //   as we are sure this cannot change with any possible sequence of future votes
+                    // * we also need quorum (which may not always be the case above)
+                    self.votes.total() >= votes_needed(self.total_weight, quorum)
+                        && self.votes.yes
+                            >= votes_needed(self.total_weight - self.votes.abstain, threshold)
+                }
```
maurolacy, line 122: Both branches are the same. We can remove the expiration condition entirely.

maurolacy (approved):
Looks good.

Only remaining (minor) concern I have is about `Abstain` subtraction when compute passing weight in `AbsoluteThreshold`.

contracts/cw3-flex-multisig/src/msg.rs
```diff
@@ -1,19 +1,137 @@
...
 
+/// This defines the different ways tallies can happen.
+///
+/// The total_weight used for calculating success as well as the weights of each
+/// individual voter used in tallying should be snapshotted at the beginning of
+/// the block at which the proposal starts (this is likely the responsibility of a
+/// correct cw4 implementation).
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
+#[serde(rename_all = "snake_case")]
+pub enum Threshold {
```
maurolacy, line 26: I've just noticed a there's duplicity between the descriptions in here, versus the ones in `ThresholdResponse` (part of `packages/cw3`).

Not wrong per se, but maybe a good idea to unify / centralize the docs in the spec. We can alsway use "See also" or simialr  to point to the main docs, by example.

ethanfrey: It is not used in `packages/cw3` explicitly, but rather when configuring an Init message in an implementation.

I guess we could move it there as a standard type to configure in `init`. I wanted to make sure the docs were visible, and if there was some way to link/import them that would be better

contracts/cw3-flex-multisig/src/state.rs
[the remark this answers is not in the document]
ethanfrey: Oh, so you are saying an abstain is counted as the average of what everyone else voted?

Meaning for `AbsolutePercentage`, we calculate 

`self.votes.yes >= votes_needed(self.total_weight - self.votes.abstain, percentage)`

Makes sense to me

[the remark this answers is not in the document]
ethanfrey: Do you want to add that as a small PR on top of this? (And also update the relevant docs)

contracts/cw3-flex-multisig/src/state.rs
ethanfrey: > Better moved to a global (config) section? Better naming would be PRECISION_FACTOR or so.

Agree with pull it out to a global

> I don't want to introduce unnecessary complexity, but this can be probably transformed into an u32 (or lower) PRECISION_DIGITS, and used as a power of 10 exponent.

It could, but why add an unneeded `.pow()` everytime we calculate a division. Maybe we can do that one time at init

contracts/cw3-flex-multisig/src/state.rs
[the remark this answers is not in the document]
maurolacy: Sure.

contracts/cw3-flex-multisig/src/state.rs
maurolacy: Maybe it can be done at init, or even at global scope, and with keeping the `const`.

I'll try it if you don't mind.

contracts/cw3-flex-multisig/src/msg.rs
maurolacy: I'll do it. And unify the docs.

contracts/cw3-flex-multisig/src/state.rs
ethanfrey: We can pull the quorum check out of this, you are correct there.

What is different is the total we compare against. Look closely at the total: unexpired is against the total number of votes (minus abstain), expired is against total weight (minus abstain)

maurolacy: No, they are not, and this is an important difference.

Once the voting is over (expired), the passing threshold must be computed over the total weight of those **who voted**.

ethanfrey: Pushing a commit that makes this much more clear

contracts/cw3-flex-multisig/src/state.rs
ethanfrey: done

contracts/cw3-flex-multisig/src/state.rs
```diff
@@ -259,13 +262,13 @@ mod test {
         let mut votes = Votes::new(7);
         votes.add_vote(Vote::No, 4);
         votes.add_vote(Vote::Abstain, 2);
-        // same expired or not, if total > 2 * yes
+        // same expired or not, if yes >= ceiling(0.5 * (total - abstained))
```
ethanfrey, line 265: we should add another test with a larger total_weight that shows this failing

ethanfrey: I just did that

ethanfrey:
I just added a few smaller commits for some of the last outstanding points.

If you can unify the docs a bit more, that would be great. Then feel free to merge when done with docs update.

maurolacy:
> I just added a few smaller commits for some of the last outstanding points.

This is looking great.

> If you can unify the docs a bit more, that would be great. Then feel free to merge when done with docs update.

OK. One more thing:
I think the different checks between expired and not expired voring must be applied to `AbsolutePercentage` too. I mean, after voting is ended, you would want the passing percentage to be computed over the weights of the ones who effectively voted (minus Abstain), not over the total weight.

Or is it called `Absolute` because of that? Then we could have a fourth version: `RelativePercentage`

maurolacy:
I'll add an issue to consider implementing `RelativePercentage` in another PR.

Merging now.

ethanfrey:
> I think the different checks between expired and not expired voting must be applied to AbsolutePercentage too. I mean, after the voting has ended, you would want the passing percentage to be computed over the weights of the ones who effectively voted (minus Abstain), not over the total weight.
>
> Or is it called Absolute because of that? Then we could have a fourth version: RelativePercentage (or plain Percentage).

`AbsolutePercentage` is basically a way of a dynamic M of N multisig. 51% automatically shifts from "3 of 5" to "5 of 9" to "6 of 10". It doesn't matter show does or does not vote, just how many sigs they are. This, like AbsoluteCount, behave very much like off-chain signature aggregation, but allow the signing group to evolve.

Your idea of `RelativePercentage` is like `ThresholdQuorum` without a quorum. That means if only one person voted on a quick election, they could get it passed. I would prefer to use a `ThresholdQuorum` with a low (eg. 10-20% quorum) instead of one with no checks at all there.

Anyway, feel free to add an issue for `RelativePercentage` and we can discuss it there. I am also happy for other types that do something not covered by these three. (note these are just binary elections, so most of the cool findings, like "ranked choice" do not apply here... we can work on multiple choice elections in another spec)

ethanfrey (commented):
I'd love hyperlinks or other way to connect these two descriptions better

contracts/cw3-flex-multisig/src/state.rs
maurolacy: Crossed posts here. Good commit separating threshold from quorum checks.

contracts/cw3-flex-multisig/src/msg.rs
```diff
@@ -21,55 +21,22 @@ pub struct InitMsg {
...
-    ///
-    /// Example: we set `percentage` to 51%. Proposal 1 starts when there is a `total_weight` of 5.
-    /// This will require 3 weight of yes votes in order to pass. Later, the Proposal 2 starts but the
-    /// `total_weight` of the group has increased to 9. That proposal will then automatically
-    /// require 5 yes of 9 to pass, rather than 3 yes of 9 as would be the case with `AbsoluteCount`.
-    ///
-    /// A proposal of this type can pass early as soon as the needed weight of yes votes has been cast.
+    /// Declares a percentage of the total weight that must cast Yes votes in order for
+    /// a proposal to pass.
+    /// See `ThresholdResponse.AbsolutePercentage` in the cw3 spec for details.
```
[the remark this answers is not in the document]
maurolacy: Yes, they are possible, and are used extensively in the crates documentation. Here in particular, I first added and afterwards then removed the hyperlink formatting, because these comments go to the json files.

We definitely should start using these links, and confirm that they are working correctly in the crates docs.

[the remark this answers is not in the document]
maurolacy: > (Side-note, we really need to review the RustDoc output once we push 0.4.0)

I propose we create an issue in cosmwasm for that, and consider / mention all the projects.

contracts/cw3-flex-multisig/src/state.rs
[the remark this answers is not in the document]
maurolacy: Sorry, here I thought you indicated a commit, not a PR. I can always revert these and re-introduce them in a different PR for documentation, if you want.

contracts/cw3-flex-multisig/src/msg.rs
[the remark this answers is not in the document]
ethanfrey: Oh, there are issues... for quite some time:

https://github.com/CosmWasm/cosmwasm/issues/436

https://github.com/CosmWasm/cosmwasm/issues/196

Just seems that all of us prefer to code than tidy up RustDoc

Issue #143, opened 2020-11-13
Add cw20 support to token-weighted group


ethanfrey:
Builds on #142 

The previous issue only supported native tokens. Extend this to support either one native denom or one cw20 address - set in init. Try for yourself which is the best code reuse - if it is easier to make 2 different contracts, one native-only and one cw20-only and share some group logic between them, that is also fine. Any given contract instance should only take either native coins or cw20 coins, so they don't need both code paths in the contract - think about making this maintainable and bug-free.

----

Here is a breakdown of the steps needed:

- [x] Use [`cw20::Balance`](https://github.com/CosmWasm/cosmwasm-plus/blob/666f265a8bdc4c9c6d47c3094a5edfa320a91bc7/packages/cw20/src/balance.rs#L11-L14) instead of `String` for [`InitMsg::denom`](https://github.com/CosmWasm/cosmwasm-plus/blob/ec45a73aad77f36d74ca23aeef77b472fe7de7e2/contracts/cw4-stake/src/msg.rs#L11) and [`Config::denom`](https://github.com/CosmWasm/cosmwasm-plus/blob/ec45a73aad77f36d74ca23aeef77b472fe7de7e2/contracts/cw4-stake/src/state.rs#L15)
- [x] [Change `handle_bond`](https://github.com/CosmWasm/cosmwasm-plus/blob/f9b41f8fbe5a0125dec1b07a53852b439edef801/contracts/cw4-stake/src/contract.rs#L67-L71) to accept native of cw20 balance. That means accepting `sender: HumanAddr, amount: cw20::Balance` instead of `info: MessageInfo`. This conversion should be done in the [`handle` switch](https://github.com/CosmWasm/cosmwasm-plus/blob/f9b41f8fbe5a0125dec1b07a53852b439edef801/contracts/cw4-stake/src/contract.rs#L61) just like in [`cw20-escrow`](https://github.com/CosmWasm/cosmwasm-plus/blob/34c26d42c2958e770c78b3e091cdf6a0503f0235/contracts/cw20-escrow/src/contract.rs#L38)
[...]

PR #185, opened 2020-12-15
Cw20 bonding curve

### LINKED ISSUE #5: Build bonding curve with cw20-base

ethanfrey:
Closes #5 

- [x] Define curve trait
- [x] Implement Constant
- [x] Implement Linear
- [x] Implement Square Root
- [x] Import default cw20-base handlers
- [x] Add custom handlers for buy and sell/burn
- [x] Store DecimalPlaces in contract storage
- [x] Configure Curve in contract (storage? external argument?)
- [x] Test buy/sell functionality
- [x] Sanity check it works with cw20-base functionality (like supply and allowances)

There were some comments that various of these decimal math operations are very slow, let's benchmark them and we can use that to see if we do a follow-up optimization PR

Follow-Up PRs:
* Benchmark Curves (native)
* Test handler gas cost (singlepass vm integration tests) - also that there are no floats in these libs
* Allow cw20 token as reserve currency (similar to #143)
* Optional exit tax (part of reserve stays in contract) along with admin who can spend that


ethanfrey:
@orkunkl you can take a look at this if you want, especially the README, `curves.rs` and the `handle_sell` and `handle_buy` functions. Most of this is stubs now with TODOs and `unimplemented!()` but happy for any input you have from working with other bonding curve packages

Issue #163, opened 2020-12-09
Support composite keys on secondary indexes (multi-index)


ethanfrey:
Right now, keys is always `Vec<u8>`.

If I combine say (status, height) as one vector, I can not easily search over all heights for a given status (which I can on primary indices). This is due to combining them, then length-prefixing.

If we return say `(U8Key, U64Key)`, then we could get a prefix on the first key (status) and range within it, just like on `Map`.

This example was the first time I came across this need (which I hacked by knowing the internals of `storage-plus`), but we will want this in the future as well as app complexity grow: https://github.com/CosmWasm/cosmwasm-plus/blob/7975aef08a8d44444742d7dc187d341f839a6039/contracts/cw3-flex-multisig/src/state.rs#L53-L83

PR #194, opened 2020-12-18
Create Admin controller

### LINKED ISSUE #181: Pull out admin helper/controller into cw0
We have this same pattern of optional admin address in lots of places.
We can make this dry.

ethanfrey:
Closes #181 

- [x] Create new package for controllers, outside of cw0
- [x] Add Admin controller here
- [x] Integrate it into `cw4-group`
- [x] Integrate it into `cw4-stake`

Follow-Up:

* Port Hooks/Claim controllers to the same pattern (and move packages) #195 
* More some structs into cw0, so they can be imported by eg. cw4?


ethanfrey:
@maurolacy curious what you think of this (just focus on `cw4-group/srccontracts.rs` and `controllers/src/admin.rs`.

Admin is a trivial example and maybe easy to cut and paste, but I was looking for a format to move these controllers, for maximum composability. So I can port hooks, claims, etc into this format.

maurolacy:
Hi, OK, will take a look tomorrow.

Making slow progress with #163 ... let's talk tomorrow, as I'm almost there with the impl, and have some questions for you. Good thing is that I now have a much better grasp at storage-plus (`Map`, `Prefix` and the like).

ethanfrey:
Nice. Yeah, storage-plus is a good piece to master and review by usage.

Good to review the api first with real usages. Then the implementation. I am sure there are ways to clean it up. Especially all the length prefixing calls. But best to actually work on the code.

I hope the Key trait makes sense (and you see how this allows us to naively construct indexes)

PR #196, opened 2020-12-21
Secondary index composite key

ethanfrey:
From @maurolacy so I can comment

After discussion, this is either very difficult or impossible with current API and encoding schemes.
We may need to differentiate VarLength indexes and FixedLength indexes

### DIFF
--- a/contracts/cw721-base/src/state.rs
+++ b/contracts/cw721-base/src/state.rs
@@ -46,7 +46,9 @@ pub fn increment_tokens(storage: &mut dyn Storage) -> StdResult<u64> {
 }
 
 pub struct TokenIndexes<'a> {
-    pub owner: MultiIndex<'a, &'a [u8], TokenInfo>,
+    // Use tuple here, so pk goes to second tuple element, and range() works,
+    // provided that index_fn result is length prefixed
+    pub owner: MultiIndex<'a, (&'a [u8], &'a [u8]), TokenInfo>,
 }
 
 impl<'a> IndexList<TokenInfo> for TokenIndexes<'a> {
@@ -58,7 +60,12 @@ impl<'a> IndexList<TokenInfo> for TokenIndexes<'a> {
 
 pub fn tokens<'a>() -> IndexedMap<'a, &'a str, TokenInfo, TokenIndexes<'a>> {
     let indexes = TokenIndexes {
-        owner: MultiIndex::new(|d| d.owner.to_vec(), "tokens", "tokens__owner"),
+        // FIXME: Publish and use to_length_prefixed
+        owner: MultiIndex::new(
+            |d| [b"\x00", &[d.owner.len() as u8], &*d.owner].concat(),
+            "tokens",
+            "tokens__owner",
+        ),
     };
     IndexedMap::new("tokens", indexes)
 }
--- a/contracts/cw721-base/src/state.rs
+++ b/contracts/cw721-base/src/state.rs
@@ -46,7 +46,7 @@ pub fn increment_tokens(storage: &mut dyn Storage) -> StdResult<u64> {
 }
 
 pub struct TokenIndexes<'a> {
-    pub owner: MultiIndex<'a, TokenInfo>,
+    pub owner: MultiIndex<'a, &'a [u8], TokenInfo>,
 }
 
 impl<'a> IndexList<TokenInfo> for TokenIndexes<'a> {
--- a/packages/storage-plus/src/indexed_map.rs
+++ b/packages/storage-plus/src/indexed_map.rs
@@ -168,7 +168,7 @@ mod test {
     }
 
     struct DataIndexes<'a> {
-        pub name: MultiIndex<'a, Data>,
+        pub name: MultiIndex<'a, &'a [u8], Data>,
         pub age: UniqueIndex<'a, Data>,
     }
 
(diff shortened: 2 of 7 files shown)


contracts/cw721-base/src/state.rs
```diff
@@ -46,7 +46,7 @@ pub fn increment_tokens(storage: &mut dyn Storage) -> StdResult<u64> {
 }
 
 pub struct TokenIndexes<'a> {
-    pub owner: MultiIndex<'a, TokenInfo>,
+    pub owner: MultiIndex<'a, &'a [u8], TokenInfo>,
```
ethanfrey, line 49: The index function (line 61) returns a Vec<u8>. You can use PkOwned type here

packages/storage-plus/src/indexed_map.rs
```diff
@@ -258,15 +294,123 @@ mod test {
...
         let aged = map.idx.age.item(&store, &too_old).unwrap();
         assert_eq!(None, aged);
     }
 
+    #[test]
+    fn range_simple_key_by_multi_index() {
+        let mut store = MockStorage::new();
+        let mut map = build_map();
+
+        // save data
```
ethanfrey, line 311: I would pull this out into a setup function as I see this in 3 tests or so.

packages/storage-plus/src/indexed_map.rs
```diff
@@ -258,15 +294,123 @@ mod test {
...
+        };
+        let pk: &[u8] = b"5630";
+        map.save(&mut store, pk, &data4).unwrap();
+
+        let marias: StdResult<Vec<_>> = map
+            .idx
+            .name
+            .range(
+                &store,
+                Some(Bound::Inclusive(to_length_prefixed(b"Maria"))),
```
ethanfrey, line 345: :+1:

packages/storage-plus/src/indexed_map.rs
```diff
@@ -258,15 +294,123 @@ mod test {
...
+        let pk: &[u8] = b"5630";
+        map.save(&mut store, pk, &data4).unwrap();
+
+        // a duplication
+        let pk: &[u8] = b"5631";
+        map.save(&mut store, pk, &data3).unwrap();
+
+        let marias: StdResult<Vec<_>> = map
+            .idx
+            .name_age
```
ethanfrey, line 398: Okay, this is where you are having trouble with the API

packages/storage-plus/src/indexes.rs
```diff
@@ -6,17 +6,26 @@ use serde::{Deserialize, Serialize};
...
 use crate::map::Map;
 use crate::prefix::range_with_prefix;
-use crate::{Bound, Endian};
+use crate::{Bound, Endian, Prefix, PrimaryKey};
+use std::marker::PhantomData;
 
 /// MARKER is stored in the multi-index as value, but we only look at the key (which is pk)
-const MARKER: u32 = 1;
+// FIXME: Re-introduce this for MultiIndex
+// const MARKER: bool = true;
```
ethanfrey, line 17: Where did this go?
btw, I used 1u32 as it has the smallest json representation (`1` vs `true`)

packages/storage-plus/src/indexes.rs
```diff
@@ -39,31 +48,36 @@ where
...
-impl<'a, T> Index<T> for MultiIndex<'a, T>
+impl<'a, K, T> Index<T> for MultiIndex<'a, K, T>
 where
     T: Serialize + DeserializeOwned + Clone,
+    K: PrimaryKey<'a>,
 {
     fn save(&self, store: &mut dyn Storage, pk: &[u8], data: &T) -> StdResult<()> {
         let idx = (self.index)(data);
-        self.idx_map.save(store, (&idx, &pk), &MARKER)
+        self.idx_map.save(store, (&idx, &pk), &data)
```
ethanfrey, line 80: We don't want to store all the data everywhere, as I need to update all indexes on any change, even when the index doesn't change. I put it in UniqueIndex as an optimization, but I think better to use MARKER there as well

packages/storage-plus/src/indexes.rs
```diff
@@ -106,6 +121,10 @@ where
         Box::new(mapped)
     }
 
+    pub fn prefix(&self, p: K::Prefix) -> Prefix<T> {
```
ethanfrey, line 124: Okay, this may be the issue - we need to see the remainder of the key used in the Prefix. Does this work in primary keys?

ethanfrey (commented):
First review on WIP

packages/storage-plus/src/indexes.rs
maurolacy: I did this to make `prefix()` work. `prefix()` returns `Prefix<T>`, so I needed the value type to match.

That said, this is just temporary, and towards a solution.

packages/storage-plus/src/indexes.rs
maurolacy: O know. See comment above. This is WIP, and I just did it to make things simpler at the beginning.

packages/storage-plus/src/indexed_map.rs
maurolacy: Yes.

packages/storage-plus/src/indexes.rs
maurolacy: This only works when you pass the complete index_function result to `prefix`.

See commented [line 399](https://github.com/CosmWasm/cosmwasm-plus/blob/7f8c315bd38f9726e4d19181844bb3a802dfde47/packages/storage-plus/src/indexed_map.rs#L399)

maurolacy:
Closing in favour of #211.

PR #201, opened 2020-12-21
Fix cw20-base helpers.ts and point to heldernet

orkunkl:
This change is required for coralnet to heldernet update in docs.

### DIFF
--- a/contracts/cw20-base/helpers.ts
+++ b/contracts/cw20-base/helpers.ts
@@ -2,7 +2,7 @@
  * This is a set of helpers meant for use with @cosmjs/cli
  * With these you can easily use the cw20 contract without worrying about forming messages and parsing queries.
  * 
- * Usage: npx @cosmjs/cli@^0.22 --init https://raw.githubusercontent.com/CosmWasm/cosmwasm-plus/master/contracts/cw20-base/helpers.ts
+ * Usage: npx @cosmjs/cli@^0.23 --init https://raw.githubusercontent.com/CosmWasm/cosmwasm-plus/master/contracts/cw20-base/helpers.ts
  * 
  * Create a client:
  *   const client = await useOptions(hackatomOptions).setup(password);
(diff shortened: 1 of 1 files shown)


contracts/cw20-base/helpers.ts
```diff
@@ -20,24 +20,29 @@ interface Options {
   readonly httpUrl: string
   readonly networkId: string
   readonly feeToken: string
-  readonly gasPrice: number
+  readonly gasPrice: GasPrice
```
ethanfrey, line 23: This changed. Is this using 0.22 still? or 0.23?

contracts/cw20-base/helpers.ts
```diff
@@ -202,7 +181,7 @@ interface CW20Contract {
   // codeId must come from a previous deploy
   // label is the public name of the contract in listing
   // if you set admin, you can run migrations on this contract (likely client.senderAddress)
-  instantiate: (codeId: number, initMsg: InitMsg, label: string, admin?: string) => Promise<CW20Instance>
+  instantiate: (codeId: number, initMsg: Record<string, unknown>, label: string, admin?: string) => Promise<CW20Instance>
```
ethanfrey, line 184: The TS compiler forces this on you?

Ideally it would auto downcast InitMsg to this more general `Record<string, unknown>`. And nice to have the clarity in the contract-specific API. If you have TS issues, leave it and I can look into it later.

ethanfrey (approved):
Looks good. Please double-check the cosmjs version and update the header if needed.

Otherwise, feel free to merge and I will look into InitMsg later

ethanfrey (commented):
Oh, and can you update `https://github.com/CosmWasm/cosmwasm-plus/tree/v0.2.1/contracts/cw20-base` to `v0.3.2`?

I think it is the same public interface, but good to keep using the most recent versions. I have done some minor bug fixes on that contract

contracts/cw20-base/helpers.ts
orkunkl: This is for v0.23.0

contracts/cw20-base/helpers.ts
orkunkl: This is the error I get with v0.23:
```
../../../../../.npm/_npx/10245/lib/node_modules/@cosmjs/cli/[eval].ts:307:53 - error TS2345: Argument of type 'InitMsg' is not assignable to parameter of type 'Record<string, unknown>'.
  Index signature is missing in type 'InitMsg'.

307     const result = await client.instantiate(codeId, initMsg, label, { memo: `Init ${label}`, admin});
                                                        ~~~~~~~
```

orkunkl:
@ethanfrey I don't know which version of `rust-optimizer` is used to build v0.3.2. Maybe we could do a minor release with v0.10.7 to make sure it is updated?

ethanfrey:
I'll merge this (doesn't have to be correct) and update shortly with the v0.4.0 release

PR #202, opened 2020-12-21
Fix cw1-subkeys helper.ts and point to heldernet

orkunkl:
This change is required for coralnet to heldernet update in docs.

### DIFF
--- a/contracts/cw1-subkeys/helpers.ts
+++ b/contracts/cw1-subkeys/helpers.ts
@@ -2,7 +2,7 @@
  * This is a set of helpers meant for use with @cosmjs/cli
  * With these you can easily use the cw20 contract without worrying about forming messages and parsing queries.
  * 
- * Usage: npx @cosmjs/cli --init https://github.com/CosmWasm/cosmwasm-plus/blob/master/contracts/cw1-subkeys/helpers.ts
+ * Usage: npx @cosmjs/cli@^0.23 --init https://github.com/CosmWasm/cosmwasm-plus/blob/master/contracts/cw1-subkeys/helpers.ts
  * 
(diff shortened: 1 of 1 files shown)


contracts/cw1-subkeys/helpers.ts
```diff
@@ -20,24 +20,29 @@ interface Options {
...
+  gasPrice:  GasPrice.fromString("0.025ucosm"),
+  bech32prefix: 'cosmos',
+  feeToken: 'ucosm',
+  faucetUrl: 'https://faucet.heldernet.cosmwasm.com/credit',
   hdPath: makeCosmoshubPath(0),
-  defaultKeyFile: path.join(process.env.HOME, ".coral.key"),
+  defaultKeyFile: path.join(process.env.HOME, ".heldernet.key"),
+  gasLimits: {
+    upload: 1500000,
+    init: 600000,
```
ethanfrey, line 42: I think these can be brought down a lot. We know what the contract does.

Anyway, low priortiy, you can merge without this. Just want to mark it as a TODO

ethanfrey (approved):
Looks good. Same general comment as #201 

Only major change needed is setting a  cosmjs version in the header.

orkunkl:
Same question goes for this PR: https://github.com/CosmWasm/cosmwasm-plus/pull/201#issuecomment-749413463

ethanfrey:
We used v0.10.4, [see here](https://github.com/CosmWasm/cosmwasm-plus/blob/fd542d229c17b21079ce2c325ab92cc159d312f9/.circleci/config.yml#L867)

Nice to point this out. v0.4.0 will be built with v0.10.7

PR #207, opened 2020-12-22
Generalize UniqueIndex keys

### LINKED ISSUE #209: Support composite keys on secondary indexes (unique-index)
Similar to #163, but just for `UniqueIndex`, which is simpler.

As part of this, first generalize index keys by using `PrimaryKey`, and then add `prefix()` and `range()` methods, to select and iterate over composite keys.

maurolacy:
Closes #209.

Use generic index keys in `UniqueIndex`, so we can `prefix()` and `range()` over them.

### DIFF
--- a/packages/storage-plus/src/indexed_map.rs
+++ b/packages/storage-plus/src/indexed_map.rs
@@ -157,6 +157,7 @@ mod test {
     use super::*;
 
     use crate::indexes::{index_int, index_string, MultiIndex, UniqueIndex};
+    use crate::U32Key;
     use cosmwasm_std::testing::MockStorage;
     use cosmwasm_std::{MemoryStorage, Order};
     use serde::{Deserialize, Serialize};
@@ -169,7 +170,7 @@ mod test {
 
     struct DataIndexes<'a> {
         pub name: MultiIndex<'a, Data>,
-        pub age: UniqueIndex<'a, Data>,
+        pub age: UniqueIndex<'a, U32Key, Data>,
     }
 
     // Future Note: this can likely be macro-derived
@@ -184,7 +185,7 @@ mod test {
     fn build_map<'a>() -> IndexedMap<'a, &'a [u8], Data, DataIndexes<'a>> {
         let indexes = DataIndexes {
             name: MultiIndex::new(|d| index_string(&d.name), "data", "data__name"),
-            age: UniqueIndex::new(|d| index_int(d.age), "data__age"),
+            age: UniqueIndex::new(|d| index_int(d.age as u32), "data__age"),
         };
         IndexedMap::new("data", indexes)
     }
@@ -257,13 +258,13 @@ mod test {
 
         // match on proper age
         let proper = index_int(42);
-        let aged = map.idx.age.item(&store, &proper).unwrap().unwrap();
+        let aged = map.idx.age.item(&store, proper).unwrap().unwrap();
         assert_eq!(pk.to_vec(), aged.0);
         assert_eq!(data, aged.1);
 
         // no match on wrong age
         let too_old = index_int(43);
-        let aged = map.idx.age.item(&store, &too_old).unwrap();
+        let aged = map.idx.age.item(&store, too_old).unwrap();
         assert_eq!(None, aged);
     }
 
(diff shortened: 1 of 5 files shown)


packages/storage-plus/src/indexes.rs
```diff
@@ -19,8 +20,8 @@ pub fn index_string(data: &str) -> Vec<u8> {
 
 // Look at https://docs.rs/endiannezz/0.4.1/endiannezz/trait.Primitive.html
 // if you want to make this generic over all ints
-pub fn index_int<T: Endian>(data: T) -> Vec<u8> {
-    data.to_be_bytes().into()
+pub fn index_int<T: Endian>(data: T) -> IntKey<T> {
+    IntKey::<T>::new(data)
```
ethanfrey, line 24: This is fine, but why even use this now that we have K
`U32Key::new(data.age)` should be clear enough (no one needs to think about endian-ness or whatever

packages/storage-plus/src/indexed_map.rs
```diff
@@ -257,13 +258,13 @@ mod test {
 
         // match on proper age
         let proper = index_int(42);
```
ethanfrey, line 260: If you are using a named key, why use index_int (which hides all this stuff)?
We can just return `U32Key`

ethanfrey (commented):
Looks good. I would make the key less opaque now (no need for `index_int` hidden away... at least for UniqueIndex). 

Otherwise, I would add a testcase for composite index. I don't know, maybe another type... where `(first_name, last_name)` must be unique, but each alone can exist.

John Doe
John Wayne
Maria Doe

are all fine, but adding a second John Doe will return an error.

ethanfrey:
Also note these storage-plus changes may be api breaking... If any contracts have to be modified when doing so, please label that api breaking in the title (meaning it must be v0.5.0 not v0.4.1)

packages/storage-plus/src/indexed_map.rs
maurolacy: All right.

maurolacy:
> Also note these storage-plus changes may be api breaking... If any contracts have to be modified when doing so, please label that api breaking in the title (meaning it must be v0.5.0 not v0.4.1)

Good point. This doesn't happen here AFAIK, but it can happen with changes to `MultiIndex`. Will take it into account.

maurolacy:
`prefix()` and `range()` now work as expected. But, they are the "canonical" impl, that is, they require `Prefix` to be over the value type, i.e. `Prefix<T>`.

Given that `range()` is a method of `Prefix<T>`, there isn't much else that can be done here. That means that, on one side, I had to make the value type used in the `UniqueIndex` map (i. e. `UniqueRef`) public. On the other, I cannot change these returned values; because `range()` is outside of my control.

I'm thinking that a good way forward would be to create a `MappedPrefix` type, which can build prefixes over arbitrary value types, but then accepts a mapping function as a parameter, in order to transform these arbitrary values to our desired types during ranging / iteration. That way, we can specify generic types (i.e. our desired `<T>`) as outputs, and use indirections internally; whose values are then "magically" mapped to the generic output type.

Something like you're already doing in `pks()`, `items()`, etc. , but generic / parametric. And, contained in its own type.

maurolacy:
> I'm thinking that a good way forward could be to create a `MappedPrefix` type

Done in the latest commits. Not with a new `MappedPrefix`, but by simply generalizing the deserialization function of `Prefix`.

maurolacy:
These commits are mostly a rebase from master.

I would like to merge this, as I think it's stable.

packages/storage-plus/src/prefix.rs
```diff
@@ -69,11 +70,20 @@ where
     T: Serialize + DeserializeOwned,
 {
     pub fn new(top_name: &[u8], sub_names: &[&[u8]]) -> Self {
+        Prefix::new_de_fn(top_name, sub_names, deserialize_kv)
+    }
+
+    pub fn new_de_fn(
```
ethanfrey, line 76: I would use a different name, but I see this extension, and good to keep the default behavior the same.

Definitely need more time to review this

maurolacy: Yes, I was struggling when naming it.

Maybe `new_custom_de`?
Another option would be to keep only `new`, and add a helper method `set_de_fn` or so. But I think it's more confusing / unclear.

maurolacy: Take your time to review this. It's a more complex variant of the unique index case, but in the end it's the same mechanism: when creating the `Prefix`, a custom deserialization function is passed, which handles the indirection. I couldn't make it simpler.

maurolacy: I meant, the related PR for the multi index.

PR #210, opened 2020-12-28
Implement PrimaryKey for generic (T, U, V) triplet

### LINKED ISSUE #197: Add (T, U, V) triple primary key
This makes very clear the distinction between prefix and suffix.

We should be able to range by fixing (T, U) and iterating over V

Let's test both the following: `((T, U), V)` and `(T, U, V)`. For both, we should be able to give (T, U) and iterate over V.

Example: `(&[u8], U32Key, U64Key)`

maurolacy:
Closes #197.

`PrimaryKey` for nested `((T, U), V)` tuples cannot at the moment be defined univocally. They can be defined, but only in a way that deserializes to `(T, U, V)`, i.e. the nesting is lost when serializing.

They are not needed / required at this point, so I think it's OK the leave them undefined for the moment.

### DIFF
--- a/packages/storage-plus/src/keys.rs
+++ b/packages/storage-plus/src/keys.rs
@@ -26,6 +26,7 @@ pub trait PrimaryKey<'a>: Clone {
 type Pk0 = ();
 type Pk1<'a> = &'a [u8];
 type Pk2<'a, T = &'a [u8], U = &'a [u8]> = (T, U);
+type Pk3<'a, T = &'a [u8], U = &'a [u8], V = &'a [u8]> = (T, U, V);
 
 type PkStr<'a> = &'a str;
 
@@ -74,6 +75,33 @@ impl<'a, T: PrimaryKey<'a> + Prefixer<'a>, U: PrimaryKey<'a>> PrimaryKey<'a> for
     }
 }
 
+// use generics for combining there - so we can use &[u8], PkOwned, or IntKey
+impl<'a, T: PrimaryKey<'a> + Prefixer<'a>, U: PrimaryKey<'a> + Prefixer<'a>, V: PrimaryKey<'a>>
+    PrimaryKey<'a> for (T, U, V)
+{
+    type Prefix = T;
+
+    fn key(&self) -> Vec<&[u8]> {
+        let mut keys = self.0.key();
+        keys.extend(&self.1.key());
+        keys.extend(&self.2.key());
+        keys
+    }
+
+    fn parse_key(serialized: &'a [u8]) -> Self {
+        let l1 = decode_length(&serialized[0..2]);
+        let first = &serialized[2..2 + l1];
+        let l2 = decode_length(&serialized[2 + l1..2 + l1 + 2]);
+        let second = &serialized[2 + l1 + 2..2 + l1 + 2 + l2];
+        let third = &serialized[2 + l1 + 2 + l2..];
+        (
+            T::parse_key(first),
+            U::parse_key(second),
+            V::parse_key(third),
+        )
+    }
+}
+
 // pub trait Prefixer<'a>: Copy {
 pub trait Prefixer<'a> {
     /// returns 0 or more namespaces that should length-prefixed and concatenated for range searches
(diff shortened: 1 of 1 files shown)


packages/storage-plus/src/keys.rs
```diff
@@ -312,6 +346,15 @@ mod test {
         assert_eq!(key, parsed);
     }
 
+    #[test]
+    fn parse_joined_keys_pk3() {
```
ethanfrey, line 350: Good test, I was just going to ask this one

packages/storage-plus/src/keys.rs
```diff
@@ -74,6 +75,33 @@ impl<'a, T: PrimaryKey<'a> + Prefixer<'a>, U: PrimaryKey<'a>> PrimaryKey<'a> for
     }
 }
 
+// use generics for combining there - so we can use &[u8], PkOwned, or IntKey
+impl<'a, T: PrimaryKey<'a> + Prefixer<'a>, U: PrimaryKey<'a> + Prefixer<'a>, V: PrimaryKey<'a>>
```
ethanfrey, line 79: This looks good

packages/storage-plus/src/keys.rs
```diff
@@ -74,6 +75,33 @@ impl<'a, T: PrimaryKey<'a> + Prefixer<'a>, U: PrimaryKey<'a>> PrimaryKey<'a> for
...
+    PrimaryKey<'a> for (T, U, V)
+{
+    type Prefix = T;
+
+    fn key(&self) -> Vec<&[u8]> {
+        let mut keys = self.0.key();
```
ethanfrey, line 85: Huh, how does this work in the end? Isn't `self.0.key()` a `Vec<u8>`?

packages/storage-plus/src/keys.rs
```diff
@@ -74,6 +75,33 @@ impl<'a, T: PrimaryKey<'a> + Prefixer<'a>, U: PrimaryKey<'a>> PrimaryKey<'a> for
     }
 }
 
+// use generics for combining there - so we can use &[u8], PkOwned, or IntKey
+impl<'a, T: PrimaryKey<'a> + Prefixer<'a>, U: PrimaryKey<'a> + Prefixer<'a>, V: PrimaryKey<'a>>
+    PrimaryKey<'a> for (T, U, V)
+{
+    type Prefix = T;
```
ethanfrey, line 82: Prefix should be `(T, U)`

packages/storage-plus/src/keys.rs
ethanfrey: Can you add a test for the prefix method?

ethanfrey (commented):
Shame you merged so fast. The code looks good but a few points it would be good to address on a follow-up PR

ethanfrey:
I re-read the issue and saw:

> Let's test both the following: ((T, U), V) and (T, U, V). For both, we should be able to give (T, U) and iterate over V.
> Example: (&[u8], U32Key, U64Key)

I understand only supporting `(T, U, V)`, but it would be good to see a test using that in a range search - a higher level test with such a key (in map or such)

packages/storage-plus/src/keys.rs
maurolacy: Yes. This is just a generalization of the `(T, U)` case above. Same code.

packages/storage-plus/src/keys.rs
maurolacy: Good catch. Wouldn't we want to consider both prefixes as valid? I mean, `(T, U)` is a prefix, but wouldn't `T` be another valid one?

maurolacy:
> I understand only supporting `(T, U, V)`, but it would be good to see a test using that in a range search - a higher level test with such a key (in map.rs or such)

OK. I'll create a small PR with the tests.

maurolacy:
> Shame you merged so fast. The code looks good but a few points it would be good to address on a follow-up PR

Merged because this was based on the original code / logic for `(T, U)`, and it was more or less straightforward. Will follow-up with more tests, and the fix for `Prefix`, if needed.

packages/storage-plus/src/keys.rs
ethanfrey: Sure, but look how this is used.
Maybe we need to go through this all and check the usage and document it better.

I don't have the head space to go deep in here now. Best is to leave these PRs open and I will review and merge them (or comment on them) when I have the time to check all the details.

Issue #199, opened 2020-12-21
Storage-plus: See if we can simplify PkOwned to Vec<u8>


ethanfrey:
Is there a reason why we don't use that directly like `&[u8]`. Maybe an oversight when I wrote this API?

Then we can just return `Vec<u8>` than forcing wrapping.

maurolacy:
Using `&[u8]` or `&str` directly _might_ be possible, but will require to carefully track and fix lifetime conficts first.

A simpler way seems to be implementing needed traits (`PrimaryKey`, `Prefixer`, probably others) directly for `Vec[u8]`, and remove `PkOwned` entirely.

maurolacy:
Created #233 to track / fix the lifetimes issue.

PR #211, opened 2020-12-28
Multi index generic key

### LINKED ISSUE #163: Support composite keys on secondary indexes (multi-index)

maurolacy:
Now closes #163.

I was finally able to implement this, after some wresting with the Rust compiler, and a lot of learning.

Basically, following the same strategy as in #207. A custom deserialization function; which in this case is more complex, as it has to resolve the value indirection by going to the store.

This implementation imposes one restriction, or better said, has one requirement from the user: The `MultiIndex` definition must include a field for the primary key.

By example, if you want to (multi) index by a given byte field, instead of defining the `MultiIndex` just for `&'a [u8]` (or `PkOwned`), you must instead define it for `(PkOwned, PkOwned)`; the last `PkOwned` is the placeholder for the primary key.

I think this restriction or requirement is reasonable, and I'll document this in the `MultiIndex`. It seems moving beyond this would require something like nested closures (using `(K, PkOwned)` internally), and this implies resolving their deserialization ambiguity (see #210).

Besides that, implementing this required passing a reference to `Storage` to the custom deserialization function; along with the pk namespace. I managed to resolve it by using function pointers, plus a couple "gluing" closures (and some parameter passing / carrying).

It would be nice to have something like currying implemented, in order to avoid having to pass around "fixed" (early defined) parameters (namely, the pk namespace), but my efforts in that direction didn't materialize. To the best of my knowledge, doing so would require using (capturing, therefore unsized) closures instead of function pointers, and wrestling either with references to opaque types (`impl Fn` and friends), or boxed (due to unsized) capturing closures.

~~**Update**: This requires #197, which was already locally merged. Better to review that first.~~

~~**Update 2**: Changed the base to `unique-index-generic-key`, as this draws on top of it.~~

**Update 3**: Only missing piece (if we agree on this, of course) would documenting the functionality / form of use in `MultiIndex`. Will be doing that soon.

### DIFF
--- a/contracts/cw1-subkeys/Cargo.toml
+++ b/contracts/cw1-subkeys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cw1-subkeys"
-version = "0.4.0"
+version = "0.4.1"
 authors = ["Ethan Frey <ethanfrey@users.noreply.github.com>"]
 edition = "2018"
 description = "Implement subkeys for authorizing native tokens as a cw1 proxy contract"
@@ -18,10 +18,10 @@ backtraces = ["cosmwasm-std/backtraces"]
 library = []
 
 [dependencies]
-cw0 = { path = "../../packages/cw0", version = "0.4.0" }
-cw1 = { path = "../../packages/cw1", version = "0.4.0" }
-cw2 = { path = "../../packages/cw2", version = "0.4.0" }
-cw1-whitelist = { path = "../cw1-whitelist", version = "0.4.0", features = ["library"] }
+cw0 = { path = "../../packages/cw0", version = "0.4.1" }
+cw1 = { path = "../../packages/cw1", version = "0.4.1" }
+cw2 = { path = "../../packages/cw2", version = "0.4.1" }
+cw1-whitelist = { path = "../cw1-whitelist", version = "0.4.1", features = ["library"] }
 cosmwasm-std = { version = "0.12.2", features = ["iterator", "staking"] }
 cosmwasm-storage = { version = "0.12.2", features = ["iterator"] }
 schemars = "0.7"
--- a/contracts/cw1-subkeys/Cargo.toml
+++ b/contracts/cw1-subkeys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cw1-subkeys"
-version = "0.3.2"
+version = "0.4.0"
 authors = ["Ethan Frey <ethanfrey@users.noreply.github.com>"]
 edition = "2018"
 description = "Implement subkeys for authorizing native tokens as a cw1 proxy contract"
@@ -18,10 +18,10 @@ backtraces = ["cosmwasm-std/backtraces"]
 library = []
 
 [dependencies]
-cw0 = { path = "../../packages/cw0", version = "0.3.2" }
-cw1 = { path = "../../packages/cw1", version = "0.3.2" }
-cw2 = { path = "../../packages/cw2", version = "0.3.2" }
-cw1-whitelist = { path = "../cw1-whitelist", version = "0.3.2", features = ["library"] }
+cw0 = { path = "../../packages/cw0", version = "0.4.0" }
+cw1 = { path = "../../packages/cw1", version = "0.4.0" }
+cw2 = { path = "../../packages/cw2", version = "0.4.0" }
+cw1-whitelist = { path = "../cw1-whitelist", version = "0.4.0", features = ["library"] }
 cosmwasm-std = { version = "0.12.2", features = ["iterator", "staking"] }
 cosmwasm-storage = { version = "0.12.2", features = ["iterator"] }
 schemars = "0.7"
--- a/contracts/cw1-subkeys/Cargo.toml
+++ b/contracts/cw1-subkeys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cw1-subkeys"
-version = "0.4.1"
+version = "0.5.0"
 authors = ["Ethan Frey <ethanfrey@users.noreply.github.com>"]
 edition = "2018"
 description = "Implement subkeys for authorizing native tokens as a cw1 proxy contract"
@@ -18,15 +18,15 @@ backtraces = ["cosmwasm-std/backtraces"]
 library = []
 
 [dependencies]
-cw0 = { path = "../../packages/cw0", version = "0.4.1" }
-cw1 = { path = "../../packages/cw1", version = "0.4.1" }
-cw2 = { path = "../../packages/cw2", version = "0.4.1" }
-cw1-whitelist = { path = "../cw1-whitelist", version = "0.4.1", features = ["library"] }
-cosmwasm-std = { version = "0.12.2", features = ["iterator", "staking"] }
-cosmwasm-storage = { version = "0.12.2", features = ["iterator"] }
+cw0 = { path = "../../packages/cw0", version = "0.5.0" }
+cw1 = { path = "../../packages/cw1", version = "0.5.0" }
+cw2 = { path = "../../packages/cw2", version = "0.5.0" }
+cw1-whitelist = { path = "../cw1-whitelist", version = "0.5.0", features = ["library"] }
+cosmwasm-std = { version = "0.13.2", features = ["iterator", "staking"] }
+cosmwasm-storage = { version = "0.13.2", features = ["iterator"] }
 schemars = "0.7"
 serde = { version = "1.0.103", default-features = false, features = ["derive"] }
 thiserror = { version = "1.0.20" }
 
 [dev-dependencies]
-cosmwasm-schema = { version = "0.12.2" }
+cosmwasm-schema = { version = "0.13.2" }
--- a/contracts/cw721-base/src/state.rs
+++ b/contracts/cw721-base/src/state.rs
@@ -3,7 +3,7 @@ use serde::{Deserialize, Serialize};
 
 use cosmwasm_std::{CanonicalAddr, StdResult, Storage};
 use cw721::{ContractInfoResponse, Expiration};
-use cw_storage_plus::{Index, IndexList, IndexedMap, Item, Map, MultiIndex};
+use cw_storage_plus::{Index, IndexList, IndexedMap, Item, Map, MultiIndex, PkOwned};
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
 pub struct TokenInfo {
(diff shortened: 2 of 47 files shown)


ethanfrey:
Sorry for not reviewing this yet.

I'm going to cut another (very minor) version with 0.13 of the sdk, then will review this more and we can include it in the larger release with cosmwasm 0.14 support.

When you have a chance, can you rebase and resolve conflicts (almost all in the Cargo.toml files, should be easy)

maurolacy:
> When you have a chance, can you rebase and resolve conflicts (almost all in the Cargo.toml files, should be easy)

Yes. In fact I have to rebase and use the new sub-prefix functionality here (if I recall correctly). Allow a few days for me to properly fininsh / test this.

ethanfrey:
Sure. No rush. I won't be doing much more work in this repo until cosmwasm 0.14-alpha (or better) is out

maurolacy:
Okay, merge was straightforward, and integrating the new functionality, too.
@ethanfrey, I think this is ready for review.

contracts/cw721-base/src/state.rs
```diff
@@ -46,7 +46,8 @@ pub fn increment_tokens(storage: &mut dyn Storage) -> StdResult<u64> {
 }
 
 pub struct TokenIndexes<'a> {
-    pub owner: MultiIndex<'a, TokenInfo>,
+    // pk goes to second tuple element
+    pub owner: MultiIndex<'a, (PkOwned, PkOwned), TokenInfo>,
```
ethanfrey, line 50: Do we have to use PkOwned and not &str?

(That can be a future optimization, but curious)

packages/storage-plus/src/indexed_map.rs
```diff
@@ -175,7 +175,8 @@ mod test {
     }
 
     struct DataIndexes<'a> {
-        pub name: MultiIndex<'a, Data>,
+        // Second arg is for storing pk
+        pub name: MultiIndex<'a, (PkOwned, PkOwned), Data>,
```
ethanfrey, line 179: It would be a good test to see if it is possible (or not) to use &str or &[u8] in secondary indexes.
Even a failure and then a commented line saying why it is not possible is a clear guide.

ethanfrey (commented):
Nice work and sorry it took so long to review.

I'd love to see if you could use &str or &[u8] for the user-provided data in these secondary indexes, so we can query with `"Maria"`, not `PkOwned("Maria".to_bytes().to_vec())`

contracts/cw721-base/src/state.rs
maurolacy: We're using `PkOwned` just because `PrimaryKey` is implemented for it. See https://github.com/CosmWasm/cosmwasm-plus/blob/4bbb2050c288bf11b4fe2933cc1f4975d144b94a/packages/storage-plus/src/keys.rs#L153L157.

It would be a matter of implementing `PrimaryKey` for `Vec[u8]` and friends. I can do it, but maybe in a different PR would be a good idea.

maurolacy:
> I'd love to see if you could use &str or &[u8] for the user-provided data in these secondary indexes, so we can query with `"Maria"`, not `PkOwned("Maria".to_bytes().to_vec())`

We can implement `Into` to ease that. Or better, get rid of `PkOwned` entirely. But I would do any of these in a different PR, as this is already crowded with stuff.

packages/storage-plus/src/indexed_map.rs
maurolacy: I think it should be possible. Maybe not `&[u8]` but `[u8]` (i.e. an owned slice). Related to the other comments. If you want me to address this here, I can give it a try during the week, or better, next week.

contracts/cw721-base/src/state.rs
ethanfrey: We also implement PrimaryKey for &[u8] and &str with some lifetime requirements. I *think* this index usage will satisfy the lifetime requirements: https://github.com/CosmWasm/cosmwasm-plus/blob/4bbb2050c288bf11b4fe2933cc1f4975d144b94a/packages/storage-plus/src/keys.rs#L26-L53

The point is not just to avoid a copy, but also to make the usage much simpler. You can just do `&canonical_addr` if the key is `&[u8]` for example

ethanfrey: Supporting Vec<u8> directly not the PkOwned wrapper is definitely another PR

packages/storage-plus/src/indexed_map.rs
ethanfrey: If you could just try it and see compiler errors that would be nice.
If it is non-trivial, just comment it out and make an issue for later

packages/storage-plus/src/indexes.rs
```diff
@@ -38,97 +37,183 @@ where
...
+    pub fn sub_prefix(&self, p: K::SubPrefix) -> Prefix<T> {
+        Prefix::with_deserialization_function(
+            self.idx_namespace,
+            &p.prefix(),
+            self.pk_namespace,
+            deserialize_multi_kv,
+        )
+    }
+
+    // FIXME?: Move to Prefix<T> for ergonomics
```
ethanfrey, line 141: Is this comment outdated? You use K::Prefix

Or you mean to move the method? If so, that can wait til later.

ethanfrey (approved):
Minor comments. Happy to see an attempt at using &[u8] in one just so you can create a proper issue for what compile errors we get.

Otherwise, good to merge

contracts/cw1-subkeys/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "cw1-subkeys"
-version = "0.4.1"
+version = "0.5.0"
```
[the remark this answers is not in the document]
ethanfrey: Oh, this apparently was just due to the filters github applied, and not part of the PR.

contracts/cw721-base/src/state.rs
maurolacy: You're right. I'll give it a try.

packages/storage-plus/src/indexes.rs
maurolacy: I mean that, yes (move the method), will clarify the comment.

contracts/cw721-base/src/state.rs
maurolacy: I now know why I used `PkOwned`: To avoid running into lifetime conflicts.

Slices cannot be used, because they don't have size known at compile time. It's the same with `str`.

That leave us with:

- `PkOwned`
- `Vec[u8]`, but it lacks trait impls. So, `PkOwned`.
- `String`. Again, it lacks needed trait impls, and probably for a good reason: unsupported bytes, utf-8 compliance, etc.

So, to move forward, we either:
- Stick to `PkOwned`, and improve usability through `Into` and `From`.
- Implement traits for `Vec[u8]` directly, and remove `PkOwned`. There's [a ticket for that already](#199).
- Fix the lifetime conflicts of `MultiIndex` and related methods, which receive a lifetime as a parameter. This is probably doable, but will require some time / effort.

ethanfrey: Sounds good.

> * Stick to PkOwned, and improve usability through Into and From.
> * Implement traits for Vec[u8] directly, and remove PkOwned. There's a ticket for that already (#199).

Of these two, I prefer the second. It's in a reasonable place in the backlog. We'll get to it after the major work ahead.

> * Fix the lifetime conflicts of UniqueIndex, MultiIndex and their methods (those which receive a lifetime as a parameter). This is probably doable, but will require some time, and effort.

I made a ticket for this one: #232

ethanfrey:
Looks good. I agree with your comments. There are 2 follow up issues further down the bakclog

packages/storage-plus/src/indexed_map.rs
maurolacy: Here are the errors, when changing the `name` `MultiIndex` to use `&[u8]` in the `indexed_map` tests:
```
   Compiling cw-storage-plus v0.5.0 (/usr2/home/mauro/work/cosmwasm-plus/packages/storage-plus)
error[E0495]: cannot infer an appropriate lifetime for autoref due to conflicting requirements
   --> packages/storage-plus/src/indexed_map.rs:209:50
    |
209 |             name: MultiIndex::new(|d, k| (d.name.as_ref(), &*k), "data", "data__name"),
    |                                                  ^^^^^^
    |
note: first, the lifetime cannot outlive the anonymous lifetime #1 defined on the body at 209:35...
   --> packages/storage-plus/src/indexed_map.rs:209:35
    |
209 |             name: MultiIndex::new(|d, k| (d.name.as_ref(), &*k), "data", "data__name"),
    |                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: ...so that reference does not outlive borrowed content
   --> packages/storage-plus/src/indexed_map.rs:209:43
    |
209 |             name: MultiIndex::new(|d, k| (d.name.as_ref(), &*k), "data", "data__name"),
    |                                           ^^^^^^
note: but, the lifetime must be valid for the lifetime `'a` as defined on the function body at 207:18...
   --> packages/storage-plus/src/indexed_map.rs:207:18
    |
207 |     fn build_map<'a>() -> IndexedMap<'a, &'a [u8], Data, DataIndexes<'a>> {
    |                  ^^
note: ...so that the expression is assignable
   --> packages/storage-plus/src/indexed_map.rs:216:9
    |
216 |         IndexedMap::new("data", indexes)
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: expected `indexed_map::IndexedMap<'a, &'a [u8], _, DataIndexes<'a>>`
               found `indexed_map::IndexedMap<'_, &[u8], _, DataIndexes<'_>>`

error: aborting due to previous error
```

maurolacy:
Happy to see this merged! 🎉 

Created #233 and #234 as follow-up.

maurolacy:
You beat me to it, closed #233 as a duplicate of #232.

PR #213, opened 2021-01-05
Triple primary key 2

maurolacy:
Follow-up of #210.

Fixes triple key `Prefix` type. Add `Map` triple keys tests.

### DIFF
--- a/packages/storage-plus/src/keys.rs
+++ b/packages/storage-plus/src/keys.rs
@@ -370,6 +370,17 @@ mod test {
     }
 
     #[test]
+    fn parse_joined_keys_pk3_alt() {
+        type K<'a> = (&'a str, U64Key, &'a str);
+
+        let key: K = ("one", 222.into(), "three");
+        let joined = key.joined_key();
+        assert_eq!(3 + 8 + 5 + 2 * 2, joined.len());
+        let parsed = K::parse_key(&joined);
+        assert_eq!(key, parsed);
+    }
+
+    #[test]
     fn parse_joined_keys_int() {
         let key: U64Key = 12345678.into();
         let joined = key.joined_key();
@@ -380,11 +391,33 @@ mod test {
 
     #[test]
     fn parse_joined_keys_string_int() {
-        let key: (U32Key, &str) = (54321.into(), "random");
+        type K<'a> = (U32Key, &'a str);
+
+        let key: K = (54321.into(), "random");
         let joined = key.joined_key();
         assert_eq!(2 + 4 + 6, joined.len());
-        let parsed = <(U32Key, &str)>::parse_key(&joined);
+        let parsed = K::parse_key(&joined);
         assert_eq!(key, parsed);
         assert_eq!("random", parsed.1);
+    }
+
+    #[test]
+    fn proper_prefixes() {
+        let simple: &str = "hello";
+        assert_eq!(simple.prefix(), vec![b"hello"]);
+
+        let pair: (U32Key, &[u8]) = (12345.into(), b"random");
+        let one: Vec<u8> = vec![0, 0, 48, 57];
+        let two: Vec<u8> = b"random".to_vec();
+        assert_eq!(pair.prefix(), vec![one.as_slice(), two.as_slice()]);
+
+        let triple: (&str, U32Key, &[u8]) = ("begin", 12345.into(), b"end");
+        let one: Vec<u8> = b"begin".to_vec();
+        let two: Vec<u8> = vec![0, 0, 48, 57];
+        let three: Vec<u8> = b"end".to_vec();
+        assert_eq!(
+            triple.prefix(),
+            vec![one.as_slice(), two.as_slice(), three.as_slice()]
+        );
     }
 }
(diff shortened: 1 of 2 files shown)


maurolacy:
I ran into problems when trying to use `(T, U)`:
```
error[E0277]: the trait bound `(T, U): std::convert::AsRef<keys::PkOwned>` is not satisfied
  --> packages/storage-plus/src/keys.rs:83:19
   |
8  | pub trait PrimaryKey<'a>: Clone {
   |           ---------- required by a bound in this
9  |     type Prefix: Prefixer<'a>;
   |                  ------------ required by this bound in `keys::PrimaryKey`
...
83 |     type Prefix = (T, U);
   |                   ^^^^^^ the trait `std::convert::AsRef<keys::PkOwned>` is not implemented for `(T, U)`
   |
   = note: required because of the requirements on the impl of `keys::Prefixer<'a>` for `(T, U)`
```

Tried to implement `AsRef<PkOwned>` explicitly for `(T, U)` but started to get collissions with other implementations... notably, of `Prefixer`... not sure how to do it properly.
So, I "cheated" and fallback to using `Pk2`. If you know how to solve the previous error, I'll be glad to generalize / fix this.

packages/storage-plus/src/keys.rs
```diff
@@ -79,7 +79,9 @@ impl<'a, T: PrimaryKey<'a> + Prefixer<'a>, U: PrimaryKey<'a>> PrimaryKey<'a> for
 impl<'a, T: PrimaryKey<'a> + Prefixer<'a>, U: PrimaryKey<'a> + Prefixer<'a>, V: PrimaryKey<'a>>
     PrimaryKey<'a> for (T, U, V)
 {
-    type Prefix = T;
+    // FIXME: Generalize
+    // type Prefix = (T, U);
+    type Prefix = Pk2<'a>;
```
ethanfrey, line 84: This is wrong - we change types magically.

You could do `Pk2<'a, T, U>`, but we need to make sure the T + U are preserved.
This likely means updating `impl Prefixer for Pk2` definition

packages/storage-plus/src/map.rs
```diff
@@ -118,6 +118,8 @@ mod test {
 
     const ALLOWANCE: Map<(&[u8], &[u8]), u64> = Map::new("allow");
 
+    const TRIPLE: Map<(&[u8], &[u8], &[u8]), u64> = Map::new("triple");
```
ethanfrey, line 121: It would be good to test with non `&[u8]` items, as in the issue.
But that will likely fail to compile/pass due to the issue mentioned above. Let's give a harder test case and fix the issues

packages/storage-plus/src/map.rs
```diff
@@ -237,6 +277,38 @@ mod test {
         );
     }
 
+    #[test]
+    #[cfg(feature = "iterator")]
+    fn range_triple_key() {
```
ethanfrey, line 282: Good test. This is what I want to see for the full-stack usage.

Can you update the TRIPLE to `(&[u8], Int32Key, Int64Key)` and verify this works just as well?

ethanfrey (commented):
Please test and update for tripple keys that don't only use `&[u8]`

packages/storage-plus/src/keys.rs
maurolacy: Yes, I know. I couldn't come up with a proper impl. Will revisit.

packages/storage-plus/src/map.rs
maurolacy: OK.

packages/storage-plus/src/map.rs
maurolacy: Sure.

ethanfrey: I fixed the prefix implemention in my last commit.
You should be able to use it in the new code.

ethanfrey: I need to reflect why we cannot satisfy Prefix automatically for anything that implements PrimaryKey - I assume this is a subset of the work.

packages/storage-plus/src/map.rs
```diff
@@ -140,17 +141,17 @@ mod test {
         assert_eq!(b"john".to_vec().as_slice(), &key[9..13]);
         assert_eq!(b"maria".to_vec().as_slice(), &key[13..]);
 
-        let path = TRIPLE.key((b"john", b"maria", b"pedro"));
+        let path = TRIPLE.key((b"john", 8u8.into(), "pedro"));
         let key = path.deref();
         // this should be prefixed(allow) || prefixed(john) || maria
         assert_eq!(
-            "triple".len() + "john".len() + "maria".len() + b"pedro".len() + 2 * 3,
+            "triple".len() + "john".len() + 1 + "pedro".len() + 2 * 3,
```
ethanfrey, line 148: :+1:

ethanfrey (approved):
Looking better. Thank you for finding these issues

maurolacy:
> Looking better. Thank you for finding these issues

Thank you for the review, and for the fixes.

Issue #214, opened 2021-01-06
Add Prefixer sub-prefixes support


maurolacy:
The current `Prefixer` trait has only a `prefix()` method that returns a "full prefix", i. e., for a given composite `PrimaryKey`, a prefix that is composed by all but the last element of the composite key.

When ranging over composite keys, we would like to be able to select the keys that are part of the (fixed) prefix. That way, we can range over, by example, all the ages related to a fixed name. As there can be multiple names with the same age, we would typically use a `MultiIndex`, disambiguating entries by (unique) primary key, plus, using a prefix that fixes only the name. That second part (using a prefix for only the first element of the key, in a triple key) is currently unsupported.

For some context, see #213, and https://github.com/CosmWasm/cosmwasm-plus/pull/213#discussion_r551976227 in particular.

Creating this issue to track possible implementations for this.

maurolacy:
First idea that comes to mind is to add a `prefix1()` method to the `Prefixer` trait, and implement it for all the currently supported types.

Advantages:

- Simple. Minimal modifications are required.

Disadvantages:

- Maybe a little unclear (though that can be fixed by documentation).
- Still requires passing the full `K::Prefix` argument to the `prefix1()` wrapper.

Variants / Improvements:

- Add a `subprefix()` method, with an int value indicating desired number of prefix elements. May be difficult to implement in a generic way.
- Add a `PrefixSub1` type, and implement a `prefix1()` wrapper for it. This adds more clutter, but may be a good balance between usability and simplicity.

maurolacy:
A possible alternative would be to use (abuse) tuples notation, using nesting for indicating the number of desired (sub-)prefixes. By example, `(T, (U, V))` would be a triple key, but with a one element prefix, whereas `((T, U), V)` (or directly `(T, U, V)`) will still be a triple key, but with a 2-element prefix.

Advantages:

- No need for extra methods.

Disadvantages:

- May be difficult to implement / generalize.
- Syntax is a little obscure / indirect.

maurolacy:
@ethanfrey, what do you think? I can present a couple impls for this, and we can discuss them over the code if you like.

By the way, this is required for the `MultiIndex` impl (see #211), but only if we ever want to `range()` over sub-prefixes.
That was the original use case / motivation, but that use case has since been deprecated (when we moved group updates to the `cw4-group` contract).

So, this is low priority, and more of a learning exercise / for the sake of completeness stuff for me at this point.

[...]

PR #215, opened 2021-01-06
Prefixer sub prefix

maurolacy:
Implemented first variant for #214.

**Update**: This turned out to be simpler, as the same `prefix()` method can be used for both, prefixes and sub-prefixes, after signalling which one through its related type alias / helper function.

Only missing piece is de-serialization of remaining keys, as they are being returned encoded. This can be tricky, as we don't have a regular format for keys, i. e. the first ones are length prefixed, but the last one isn't.
I think we can live with this: it's an issue only for `Map`, and probably `IndexedMap`. For `UniqueIndex`, and upcoming `MultiIndex` with sub-prefixes, this is already taken care of by the custom de-serialization function.

### DIFF
--- a/packages/storage-plus/src/keys.rs
+++ b/packages/storage-plus/src/keys.rs
@@ -7,9 +7,10 @@ use crate::Endian;
 // pub trait PrimaryKey<'a>: Copy {
 pub trait PrimaryKey<'a>: Clone {
     type Prefix: Prefixer<'a>;
+    type SubPrefix: Prefixer<'a>;
 
     /// returns a slice of key steps, which can be optionally combined
-    fn key<'b>(&'b self) -> Vec<&'b [u8]>;
+    fn key(&self) -> Vec<&[u8]>;
 
     fn joined_key(&self) -> Vec<u8> {
         let keys = self.key();
@@ -24,8 +25,9 @@ pub trait PrimaryKey<'a>: Clone {
 
 impl<'a> PrimaryKey<'a> for &'a [u8] {
     type Prefix = ();
+    type SubPrefix = ();
 
-    fn key<'b>(&'b self) -> Vec<&'b [u8]> {
+    fn key(&self) -> Vec<&[u8]> {
         // this is simple, we don't add more prefixes
         vec![self]
     }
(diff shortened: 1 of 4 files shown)


packages/storage-plus/src/keys.rs
```diff
@@ -38,8 +40,9 @@ impl<'a> PrimaryKey<'a> for &'a [u8] {
 // Provide a string version of this to raw encode strings
 impl<'a> PrimaryKey<'a> for &'a str {
     type Prefix = ();
+    type SubPrefix = ();
```
ethanfrey, line 43: Simple. Just add the 2-step variant rather than some generic N-step construction. Should be plenty for any case I can think of.

packages/storage-plus/src/keys.rs
```diff
@@ -52,8 +55,9 @@ impl<'a> PrimaryKey<'a> for &'a str {
 // use generics for combining there - so we can use &[u8], PkOwned, or IntKey
 impl<'a, T: PrimaryKey<'a> + Prefixer<'a>, U: PrimaryKey<'a>> PrimaryKey<'a> for (T, U) {
     type Prefix = T;
+    type SubPrefix = ();
 
-    fn key<'b>(&'b self) -> Vec<&'b [u8]> {
+    fn key(&self) -> Vec<&[u8]> {
```
ethanfrey, line 60: And all that was unneeded?
Looks like a better clippy/linter got this working well

packages/storage-plus/src/map.rs
```diff
@@ -308,6 +318,32 @@ mod test {
...
+            .range(&store, None, None, Order::Ascending)
+            .collect();
+        let all = all.unwrap();
+        assert_eq!(3, all.len());
+        // FIXME: range() works, but remaining keys are still encoded
+        assert_eq!(
+            all,
+            vec![
+                (
+                    [to_length_prefixed(b"\x09"), b"recipient".to_vec()].concat(),
```
ethanfrey, line 334: Ah, the big-endian encoding of U8key manually...
This is fine for now. It would be interesting to see how to auto-parse these keys, but the caller should know how in the end

ethanfrey (approved):
Looks good. Thanks

packages/storage-plus/src/keys.rs
maurolacy: This is something the idea built-in linter recommends. These are called "elided lifetimes", and the linter's recommendation is to remove them.

packages/storage-plus/src/map.rs
maurolacy: Yes. It could be handled automatically by the deserialization function (returning a vec! of (properly deserialized) tuples). But, for that we need to know if the sub-prefix is composed by one or more keys... if all the keys where length-prefixed that would be straightforward: we just honour the length-prefixed information. But currently, the last key is not length-prefixed. So, this introduces ambiguity.

I think an easy fix would be to make all keys length-prefixed, and for the deserilization fn to handle that directly.

maurolacy: We could also consider reducing the max length of length-prefixed keys to one byte (255), to save some bytes / processing.

PR #218, opened 2021-01-10
Make self constant in Item::update

### LINKED ISSUE #217: Fix const_item_mutation warnings
Starting [with Rust 1.48.0 there is](https://github.com/rust-lang/rust/pull/75573) a new `const_item_mutation` warning, which is triggered many times in this repo. I think the warning is right, and calling any mut metod of a `const`ant like `TOTAL` is dangerous. I did not look into this in detail, but if global state is required, `const` need to change to `static`. However, maybe `update` should not take a mutable `self` reference.

```
warning: taking a mutable reference to a `const` item
   --> contracts/cw4-stake/src/contract.rs:184:5
    |
184 | /     TOTAL.update(storage, |total| -> StdResult<_> {
185 | |         Ok(total + new.unwrap_or_default() - old.unwrap_or_default())
186 | |     })?;
    | |______^
    |
    = note: `#[warn(const_item_mutation)]` on by default
    = note: each usage of a `const` item creates a new temporary
    = note: the mutable reference will refer to this temporary, not the original `const` item
note: mutable reference created due to call to this method
   --> /projects/CosmWasm/cosmwasm-plus/packages/storage-plus/src/item.rs:64:5
    |
64  | /     pub fn update<A, E>(&mut self, store: &mut dyn Storage, action: A) -> Result<T, E>
65  | |     where
66  | |         A: FnOnce(T) -> Result<T, E>,
67  | |         E: From<StdError>,
...   |
72  | |         Ok(output)
73  | |     }
    | |_____^
note: `const` item defined here
   --> contracts/cw4-stake/src/state.rs:24:1
    |
24  | pub const TOTAL: Item<u64> = Item::new(TOTAL_KEY);
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: 1 warning emitted
```

webmaster128:
Closes #217

### DIFF
--- a/packages/storage-plus/src/item.rs
+++ b/packages/storage-plus/src/item.rs
@@ -61,7 +61,7 @@ where
     /// in the database. This is shorthand for some common sequences, which may be useful.
     ///
     /// If the data exists, `action(Some(value))` is called. Otherwise `action(None)` is called.
-    pub fn update<A, E>(&mut self, store: &mut dyn Storage, action: A) -> Result<T, E>
+    pub fn update<A, E>(&self, store: &mut dyn Storage, action: A) -> Result<T, E>
     where
         A: FnOnce(T) -> Result<T, E>,
         E: From<StdError>,


ethanfrey (approved):
Good call, we no longer need &mut self anywhere in the storage plus storage helpers

webmaster128:
Yeah, storage is mutable but not the config

PR #220, opened 2021-01-18
Update contracts and packages to cw 0.13.2

orkunkl:


contracts/cw4-stake/src/contract.rs
```diff
@@ -158,6 +158,7 @@ pub fn handle_unbond(
     })
 }
 
+#[allow(const_item_mutation)]
```
orkunkl, line 161: Without this linter config the following error is thrown:
```
warning: taking a mutable reference to a `const` item
   --> contracts/cw4-stake/src/contract.rs:184:5
    |
184 | /     TOTAL.update(storage, |total| -> StdResult<_> {
185 | |         let res = total + new.unwrap_or_default() - old.unwrap_or_default();
186 | |         Ok(res)
187 | |     })?;
    | |______^
    |
    = note: `#[warn(const_item_mutation)]` on by default
    = note: each usage of a `const` item creates a new temporary
    = note: the mutable reference will refer to this temporary, not the original `const` item
note: mutable reference created due to call to this method
   --> /code/packages/storage-plus/src/item.rs:64:5
    |
64  | /     pub fn update<A, E>(&mut self, store: &mut dyn Storage, action: A) -> Result<T, E>
65  | |     where
66  | |         A: FnOnce(T) -> Result<T, E>,
67  | |         E: From<StdError>,
...   |
72  | |         Ok(output)
73  | |     }
    | |_____^
note: `const` item defined here
   --> contracts/cw4-stake/src/state.rs:25:1
    |
25  | pub const TOTAL: Item<u64> = Item::new(TOTAL_KEY);
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: 1 warning emitted
```
Compiler assumes the `TOTAL` value is being mutated.

maurolacy: Make sure you're using the same linter version than the CI:
```
export CLIP_VERSION="+1.47.0"
cargo $CLIP_VERSION clippy -- -D warnings
```

When doing that, this warning is not reported.

maurolacy (commented):
lgtm.

contracts/cw4-stake/src/contract.rs
maurolacy: The clippy version mismatch is why you're hitting errors in the CI, by the way.

orkunkl: Oh yeah you are right, I was running on rust version 1.49

ethanfrey: The bug is the `&mut self` in update.
I think this was fixed (or at least reported).
I will check, we should definitely modify that - not disable the warning, but make it non mut

ethanfrey (approved):
I removed some hanging `&mut self` which were rightly causing linter errors.
Otherwise, looks good.

PR #221, opened 2021-02-05
Juggernaut/add cw20 support

juggernaut09:
This pull Request is regarding the issue #143

- used `cw20::Denom` instead of String in `InitMsg::denom` and `Config::denom`
-  Changed `handle_bond` to accept native of cw20 balance. That means accepting `sender: HumanAddr, amount: cw20::Balance` instead of `info: MessageInfo`.
-  `handle_bond` is now accepting incoming tokens using a switch on `config.denom`. Kept the current logic for native case.
-  Update `handle_claim` to release either native token (as now) or cw20-token (used unimplemented!() now, implemented using a match case.
- @ethanfrey please review these changes.


contracts/cw4-stake/src/contract.rs
```diff
@@ -67,37 +65,88 @@ pub fn handle(
...
-        0 => Err(ContractError::NoFunds {}),
-        1 => {
-            if info.sent_funds[0].denom == cfg.denom {
-                Ok(info.sent_funds[0].amount)
-            } else {
-                Err(ContractError::MissingDenom(cfg.denom.clone()))
+    let sent_funds = match amount {
+        Balance::Native(balance) => {
+            GenericBalance {
+                native: balance.0,
```
ethanfrey, line 79: I think this matching can be done a lot simpler.

Here is the check for a cw20 whitelist from escrow: https://github.com/CosmWasm/cosmwasm-plus/blob/master/contracts/cw20-escrow/src/contract.rs#L127-L132 (this is a bit easier as we only have one denom to check here)

I would suggest this code for checking the balance: https://github.com/CosmWasm/cosmwasm-plus/blob/master/packages/cw0/src/payment.rs#L13-L36 but I see it needs MessageInfo (which we don't have here). You could make and export another function `must_pay_fund()` that does the body, such that we have:

```rust
pub fn must_pay(info: &MessageInfo, denom: &str) -> Result<Uint128, PaymentError> {
    must_pay_funds(&info.sent_funds, denom)
}
```

contracts/cw4-stake/src/state.rs
```diff
@@ -31,3 +33,44 @@ pub const MEMBERS: SnapshotMap<&[u8], u64> = SnapshotMap::new(
 );
 
 pub const STAKE: Map<&[u8], Uint128> = Map::new("stake");
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug, Default)]
+pub struct GenericBalance {
```
ethanfrey, line 38: This makes sense for another usecase with multiple possibilities.

Here the know we have a balance of exactly one token type. This could be a simpler enum (and never Vec<Cw20Coin>, but just one). In the end, you could likely just use Balance where you want to use this.

contracts/cw4-stake/src/state.rs
```diff
@@ -6,13 +6,15 @@ use cw0::Duration;
...
+use cw20::{Balance, Cw20Coin};
+use cosmwasm_std::Coin;
 
 pub const CLAIMS: Claims = Claims::new("claims");
 
 #[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
 pub struct Config {
     /// denom of the token to stake
-    pub denom: String,
+    pub denom: Balance,
```
ethanfrey, line 17: Why a balance here? This is a different object, with no amount:

```rust
pub enum Denom {
  Native(String),
  Cw20(HumanAddress),
}
```

This will just have one object of one type and be much simpler to compare with.
I think with proper data structures, the logic in the contract can become much simpler

contracts/cw4-stake/src/contract.rs
```diff
@@ -350,7 +427,14 @@ mod tests {
         unbonding_period: Duration,
     ) {
         let msg = InitMsg {
-            denom: DENOM.to_string(),
+            denom: Balance::Native(NativeBalance(vec![Coin {
+                denom: String::from("stake"),
```
ethanfrey, line 431: How about `denom: Denom::Native("stake".to_string()),` ?

contracts/cw4-stake/src/msg.rs
```diff
@@ -3,12 +3,13 @@ use serde::{Deserialize, Serialize};
...
 use cosmwasm_std::{Coin, HumanAddr, Uint128};
 use cw0::Duration;
+use cw20::Balance;
 pub use cw_controllers::ClaimsResponse;
 
 #[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
 pub struct InitMsg {
     /// denom of the token to stake
-    pub denom: String,
+    pub denom: Balance,
```
ethanfrey, line 12: And same denom struct here

contracts/cw4-stake/src/contract.rs
```diff
@@ -67,37 +65,88 @@ pub fn handle(
...
+        },
+        Balance::Cw20(token) => {
+            GenericBalance {
+                native: vec![],
+                cw20: vec![token],
             }
+        },
+    };
+    let sent;
+    if cfg_denom.native.len() > 0 && sent_funds.native.len() > 0{
```
ethanfrey, line 106: With proper data structs, you could even do a double match:

```rust
let amount = match (amount, cfg_denom) {
    (Denom::Native(want), Balance::Native(have)) => must_pay_funds(have, want),
    (Denom::Cw20(want), Denom::Cw20(have)) => if want == have.address {
        Ok(have.amount)
      } else {
        Err(ContractError::InvalidDenom(...))
      }
    _ => Err(ContractError::MixedNativeAndCw20(...)),
}?;
```

ethanfrey: This should replace 50-60 lines of code currently

ethanfrey (changes requested):
Nice start. The wiring up of the objects looks good, but I think you made it much too complex for checking the input denom.

Hope these suggestions of simpler data structs help.

juggernaut09:
Thank you for reviewing. I will come up with the optimized code in the next commit.

contracts/cw4-stake/src/state.rs
juggernaut09: Do you want me to declare this new enum in CW4 package?

ethanfrey:
I got a notification, but the last commit was 12 days ago.
Happy to review if there is some new code pushed.

contracts/cw4-stake/src/contract.rs
```diff
@@ -213,9 +228,17 @@ pub fn handle_claim(
     }
 
     let config = CONFIG.load(deps.storage)?;
-    let amount = coins(release.u128(), config.denom);
+    let amount;
+    if let Denom::Native(denom) = &config.denom {
```
ethanfrey, line 232: You can use a match here, then you don't need the final else (match can prove it covers all enum variants)

contracts/cw4-stake/src/contract.rs
```diff
@@ -67,37 +70,33 @@ pub fn handle(
...
-
-    // ensure the sent denom was proper
-    // NOTE: those clones are not needed (if we move denom, we return early),
-    // but the compiler cannot see that (yet...)
-    let sent = match info.sent_funds.len() {
-        0 => Err(ContractError::NoFunds {}),
-        1 => {
-            if info.sent_funds[0].denom == cfg.denom {
-                Ok(info.sent_funds[0].amount)
+    let amount = match (&cfg.denom, &amount) {
```
ethanfrey, line 77: Nice

contracts/cw4-stake/src/state.rs
```diff
@@ -31,3 +33,44 @@ pub const MEMBERS: SnapshotMap<&[u8], u64> = SnapshotMap::new(
 );
 
 pub const STAKE: Map<&[u8], Uint128> = Map::new("stake");
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug, Default)]
```
ethanfrey, line 37: This is no longer used anywhere, correct? Can you delete it?

packages/cw0/src/lib.rs
```diff
@@ -1,11 +1,12 @@
...
-pub use crate::expiration::{Duration, Expiration, DAY, HOUR, WEEK};
 pub use pagination::{
     calc_range_end_human, calc_range_start_human, calc_range_start_string, maybe_canonical,
 };
 pub use payment::{may_pay, must_pay, nonpayable, PaymentError};
+
+pub use crate::balance::NativeBalance;
+pub use crate::expiration::{Duration, Expiration, DAY, HOUR, WEEK};
+
+mod balance;
```
ethanfrey, line 9: General style is to put mods on top and use/pub use after that.

packages/cw20/src/denom.rs
```diff
@@ -0,0 +1,25 @@
+use cosmwasm_std::CanonicalAddr;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
```
ethanfrey, line 5: Very nice

ethanfrey (commented):
Much nicer.

We will be doing a big overhaul to v0.14 cosmwasm soon. 
Can you address the few PR comments I added and we can merge this in (half done), so it gets updated along with the rest of the contracts?

ethanfrey (commented):
Thanks for the update. The CosmWasm v0.14.0 upgrade happened faster than expected.

Can you rebase and resolve conflicts. Then I am happy to merge. You can look at the migration PR to see some changes needed (in particular, handle -> execute as entry point) and returning a plain `Response` object (same fields, just a rename)

ethanfrey (approved):
Thanks for fixing this up.
I'm going to merge this in now. I will be doing some refactoring in another PR, then happy for you to complete the support.

PR #223, opened 2021-02-12
cw3-fixed-multisig: write cw20 multi-contract mint test

orkunkl:
This PR is for multi-contract testing framework workshop.
Also will increase coverage.

### DIFF
--- a/contracts/cw3-fixed-multisig/src/integration_tests.rs
+++ b/contracts/cw3-fixed-multisig/src/integration_tests.rs
@@ -1,3 +1,5 @@
+#![cfg(test)]
+
 use crate::contract::{handle, init, query};
 use crate::msg::{HandleMsg, InitMsg, Voter};
 use cosmwasm_std::testing::{mock_env, MockApi, MockStorage};
--- a/contracts/cw3-fixed-multisig/src/integration_tests.rs
+++ b/contracts/cw3-fixed-multisig/src/integration_tests.rs
@@ -1,0 +1,142 @@
+use crate::contract::{handle, init, query};
+use crate::msg::{HandleMsg, InitMsg, Voter};
+use cosmwasm_std::testing::{mock_env, MockApi, MockStorage};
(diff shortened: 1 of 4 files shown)


maurolacy:
Hi,

Can you add some context / description?
Also, this mentions `flex` in the title and branch, but the changes are in the `cw3-fixed-multisig` contract.

contracts/cw3-fixed-multisig/src/integration_tests.rs
```diff
@@ -0,0 +1,144 @@
+#![cfg(test)]
```
maurolacy, line 1: If this is meant as a "classic" integration test (i. e. to be compiled to wasm, and run in the VM), it must go in `tests/` directory, and called `integration.rs`. Then you can do `cargo wasm && cargo integration-test` to run it.

Besides, you'll have to modify it to run in the VM context. That's relatively straightforward to do. Take a look at the header comments in any of the integration tests in the `cosmwasm` repo.

orkunkl:
@maurolacy You are right about the naming, I got distracted when creating the PR. I will fix the commit when squashing the commits.

contracts/cw3-fixed-multisig/src/integration_tests.rs
ethanfrey: Another name would be better, but it does belong here. 
It runs in a (mock) multi-contract environment. Actually more useful feedback than vm integration tests in may cases as you capture contract-contract interactions.

No need to move it, maybe just rename it

ethanfrey:
@orkunkl Can you polish this off so we can merge. I want to do a big cosmwasm-plus upgrade to v0.14.0 soon, and this should be in master first. (Well, I don't really want to do the upgrade but need it done so I can write some contracts for hackatom ru... any volunteers to bump cosmwasm?)

ethanfrey:
Hmm... I'll just merge this in cuz Mauro started the upgrade. But please feel free to polish it off later.

PR #235, opened 2021-03-04
Upgrade to 0.14.0 alpha2

### LINKED ISSUE #225: Upgrade to CosmWasm 0.14
When released

### LINKED ISSUE #229: Upgrade to CosmWasm v0.14.0

maurolacy:
Closes #229 .

A couple of tests in multi-test are failing, but publishing anyway as they are interesting cases to look at (~~wrong~~ sender can now send bank messages in the multi-test context).

**Update**: This makes sense, because `BankMsg` doesn't have a `from_address` anymore. Will just modify / remove these tests.

Still needs some styling, and closely following `MIGRATING.md` guidelines, but this is compiling, passes linting AFAIK, and passes ~~almost~~ all tests.

### DIFF
--- a/contracts/cw20-base/src/allowances.rs
+++ b/contracts/cw20-base/src/allowances.rs
@@ -1,6 +1,6 @@
 use cosmwasm_std::{
-    attr, Binary, BlockInfo, CanonicalAddr, Deps, DepsMut, Env, HandleResponse, HumanAddr,
-    MessageInfo, StdResult, Storage, Uint128,
+    attr, Binary, BlockInfo, CanonicalAddr, Deps, DepsMut, Env, HumanAddr, MessageInfo, Response,
+    StdResult, Storage, Uint128,
 };
 use cw20::{AllowanceResponse, Cw20ReceiveMsg, Expiration};
 
@@ -14,7 +14,7 @@ pub fn handle_increase_allowance(
     spender: HumanAddr,
     amount: Uint128,
     expires: Option<Expiration>,
-) -> Result<HandleResponse, ContractError> {
+) -> Result<Response, ContractError> {
     let spender_raw = &deps.api.canonical_address(&spender)?;
     let owner_raw = &deps.api.canonical_address(&info.sender)?;
 
@@ -34,7 +34,8 @@ pub fn handle_increase_allowance(
         },
     )?;
 
-    let res = HandleResponse {
+    let res = Response {
+        submessages: vec![],
         messages: vec![],
         attributes: vec![
             attr("action", "increase_allowance"),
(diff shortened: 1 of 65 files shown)


ethanfrey:
I did some cleanup to see if this passes CI now.

I wonder if we should rename `HandleMsg` to `ExecuteMsg` across the board as well? This will change all the schema file output as well.

contracts/cw20-base/src/allowances.rs
```diff
@@ -176,7 +179,8 @@ pub fn handle_burn_from(
         Ok(meta)
     })?;
 
-    let res = HandleResponse {
+    let res = Response {
```
ethanfrey, line 182: Would it make sense to extend `Response::default()`?

Or do something like:

```
let mut res = Response::default();
res.add_attribute("action", "burn_from");
res.add_attribute("from", owner);
res.add_attribute("by", deps.api.human_address(&spender_raw));
res.add_attribute("amount", amount);
Ok(res)
```

We could do this consistently if you like (at least on a few contracts as demo and make an issue to finish it).

NB: I would prefer the shorter (and more Ethereum like) `res.emit("from", owner)` but Simon didn't like that as it wasn't clear. Curious as to your opinion.

ethanfrey (commented):
Looking good. If there are some particular tests you want me to look at closer, just add a comment there.

I am curious about renaming `HandleMsg` -> `ExecuteMsg` and anything else we need to change like that for consistency

contracts/cw20-base/src/allowances.rs
maurolacy: I prefer it as it is now, for clarity.

A `default()` and overwriting is discouraged for style, by the way. See https://rust-lang.github.io/rust-clippy/master/#field_reassign_with_default 
 
A default can still be useful, though now that addresses don't have defaults, a "full default" is not possible anymore.

maurolacy:
> I did some cleanup to see if this passes CI now.
> 
> I wonder if we should rename `HandleMsg` to `ExecuteMsg` across the board as well? This will change all the schema file output as well.

I think the sensible thing is to rename it. Particularly when writing new contracts, it would be confusing to have to write a `HandleMsg` for `execute`.

Maybe we can leave this change for a future version, though, as not to introduce so many changes at once.

ethanfrey:
> > I did some cleanup to see if this passes CI now.
> > I wonder if we should rename `HandleMsg` to `ExecuteMsg` across the board as well? This will change all the schema file output as well.
> 
> I think the sensible thing would be to rename it. Particularly when writing new contracts, it would be confusing to have to write a `HandleMsg` for `execute`.
> 
> Maybe we can leave this change for a future version, though, as not to introduce so many changes at once.

We can do that in another PR, but let's do this by the next v0.6.0 cosmwasm-plus release, so that is a stable point.

maurolacy:
Only missing piece AFAIK, was migrating to the new entry point system. Do the old system still works? If not, we should open a follow-up issue and finish that.

ethanfrey:
Yeah, they both work and there is a new issue for that one. It's a bit down the backlog https://github.com/CosmWasm/cosmwasm-plus/issues/230 But if you want to, you can pull that next (it shouldn't be a big deal)

maurolacy:
Just for reference / documentation, addressed pending changes in #249 and #250.

Issue #228, opened 2021-03-02
Add contract sanity checking during build / release


maurolacy:
Let's add sanity checking (check exports, imports, deterministic middleware, ...) to wasm-compiled (.wasm) smart contracts.

An easy way to do this would be to add a simple / dummy integration test for the contract. Given that compiling and running integration tests is expensive, we could run these tests in the CI only during the contract building / release step.

An alternative or complement to this would be to use a small utility to just compile the contract. Something similar to [`module_size`](https://github.com/CosmWasm/cosmwasm/blob/main/packages/vm/examples/module_size.rs). Perhaps with some additional checks / validations.

Let's do this for cosmwasm-plus contracts, and also cosmwasm and cosmwasm-examples contracts.

PR #237, opened 2021-03-05
Cw20 bonding deterministic

### LINKED ISSUE #227: Fix / remove cw20-bonding floating point instructions
`cw20-bonding` is failing to compile with: `CompileErr { msg: "Could not compile: WebAssembly translation error: Error in middleware Deterministic: Non-determinstic operator detected: I64ReinterpretF64" }'`

It seems there are some floating point instructions still hanging around in there. Let's investigate why this is happening.

Related issue: [add sanity checks during build / release](https://github.com/CosmWasm/cosmwasm-plus/issues/228).

maurolacy:
Closes #227.

Non-determinism (`I64ReinterpretF64` / `F64ConvertI64U` operators) was caused by the use of
the `num-integer` crate.

Replaced it by a couple of crates for integer sqrt and integer cbrt.

- [x] Add a simple integration test.
- [ ] Replace integration test with #228, when done.


ethanfrey:
Nice find with the crates. Not just where it came from (me blindly trusting crates with name "integer" not to include floating point ops), but finding alternatives.

Looks like it is just a CI bug. Happy if you want to do #228 first and then rebase this using that CI when it is ready.

It makes me happy to know you have a solution you tested locally :relieved:

ethanfrey (approved):
Looks good as soon as it passes CI.

Feel free to fix it here, or base it on #228

PR #238, opened 2021-03-06
Create Contract to send cw20 tokens over ics20 

### LINKED ISSUE #226: Add cw20-ics20 contract
Requires #225 

Use the new IBC entry points to create a contract that accepts cw20 tokens and sends them to a remote chain via ics20. Must accept the tokens it issued and return them to the original contract. 

Accepting remote tokens (issued on other chain) and creating a new cw20-base contract to handle them is out-of-scope here and can be done in a future issue if desired (but maybe they should just connect to `transfer`?

ethanfrey:
Closes #226


contracts/cw20-ics20/src/ibc.rs
```diff
@@ -0,0 +1,248 @@
...
+    Ok(res)
+}
+
+fn enforce_order_and_version(channel: &IbcChannel) -> Result<(), ContractError> {
+    if channel.version != ICS20_VERSION {
+        return Err(ContractError::InvalidIbcVersion {
+            version: channel.version.clone(),
+        });
+    }
+    if let Some(version) = &channel.counterparty_version {
```
alpe, line 90: 👍  not always set

contracts/cw20-ics20/src/ibc.rs
```diff
@@ -0,0 +1,248 @@
...
+        if version != ICS20_VERSION {
+            return Err(ContractError::InvalidIbcVersion {
+                version: version.clone(),
+            });
+        }
+    }
+    if channel.order != ICS20_ORDERING {
```
alpe, line 97: 👍  important

contracts/cw20-ics20/src/ibc.rs
```diff
@@ -0,0 +1,248 @@
...
+    enforce_order_and_version(&channel)?;
+
+    let info = ChannelInfo {
+        id: channel.endpoint.channel_id,
+        counterparty_endpoint: channel.counterparty_endpoint,
+        connection_id: channel.connection_id,
+    };
+    CHANNEL_INFO.save(deps.storage, &info.id, &info)?;
+
+    // TODO: add events/attributes here?
```
alpe, line 79: none emitted for this in the transfer module

contracts/cw20-ics20/src/ibc.rs
```diff
@@ -0,0 +1,248 @@
...
+        },
+    )?;
+
+    // if we have funds, now send the tokens to the requested recipient
+    let to_send = Amount::from_parts(denom, amount);
+    let msg = send_amount(to_send, HumanAddr::from(msg.receiver))?;
+    let res = IbcReceiveResponse {
+        acknowledgement: ack_success()?,
+        messages: vec![msg],
+        // TODO: similar event messages like ibctransfer module
```
alpe, line 144: Following event is emitted:
```go
			types.EventTypePacket,
			sdk.NewAttribute(sdk.AttributeKeyModule, types.ModuleName),
			sdk.NewAttribute(types.AttributeKeyReceiver, data.Receiver),
			sdk.NewAttribute(types.AttributeKeyDenom, data.Denom),
			sdk.NewAttribute(types.AttributeKeyAmount, fmt.Sprintf("%d", data.Amount)),
			sdk.NewAttribute(types.AttributeKeyAckSuccess, fmt.Sprintf("%t", err != nil)),
```

contracts/cw20-ics20/src/ibc.rs
```diff
@@ -0,0 +1,248 @@
...
+    CHANNEL_STATE.update(
+        deps.storage,
+        (&channel, &denom),
+        |orig| -> Result<_, ContractError> {
+            // this will return error if we don't have the funds there to cover the request (or no denom registered)
+            let mut cur = orig.ok_or(ContractError::InsufficientFunds {})?;
+            cur.outstanding = (cur.outstanding - amount)?;
+            Ok(cur)
+        },
+    )?;
```
alpe, line 136: I don't fully understand this block so I may go this wrong. The "no funds" is an IBC application level error that is part of the payload and should not result in an error that reverts the TX in the module.
See https://github.com/cosmos/cosmos-sdk/blob/v0.41.4/proto/ibc/core/channel/v1/channel.proto#L145

alpe (commented):
No time left for a proper require but I peeked and ❤️  it

contracts/cw20-ics20/src/ibc.rs
ethanfrey: You are very correct here.
I made a separate function in https://github.com/CosmWasm/cosmwasm-plus/pull/238/commits/0844e34447f03ea464f281e66bb1b0816d8c7aa7 so that I can catch the errors and turn them into "acknowledgement messages containing errors".

contracts/cw20-ics20/src/ibc.rs
ethanfrey: ohh.... this is a very important and very major issue!

packet and acknowledgements are never stored in the iavl tree, but rather their hashes are there and you need to get the preimage from the events.

When we submit `SendPacket`, this is handled in wasmd and *should* emit such a packet (need to check). Since the contract cannot create custom event types, we need to also auto-emit such an event in wasmd for acknowledgements. Otherwise we cannot relay such packets.

ethanfrey: Okay, investigated a bit more. The ChannelKeeper manages all the essential events.

On [`SendPacket`](https://github.com/cosmos/cosmos-sdk/blob/v0.42.0/x/ibc/core/04-channel/keeper/packet.go#L120-L135) and [`WriteAcknowledgement`](https://github.com/cosmos/cosmos-sdk/blob/v0.42.0/x/ibc/core/04-channel/keeper/packet.go#L367-L382)

ethanfrey: Update: We return the acknowledgement from `IbcModule.OnRecvPacket` in `x/wasm`.

That is the [same pattern that is used in `ibctransfer`](https://github.com/cosmos/cosmos-sdk/blob/v0.42.0/x/ibc/applications/transfer/module.go#L322-L354) and when digging deeper, I see that the acknowledgement is [written by the ibc msg server](https://github.com/cosmos/cosmos-sdk/blob/v0.42.0/x/ibc/core/keeper/msg_server.go#L445-L458).

So, it looks like we don't need to do anything here, but add some app-specific attributes

.circleci/config.yml
```diff
@@ -390,6 +391,41 @@ jobs:
...
+          path: ~/project
+      - run:
+          name: Version information
+          command: rustc --version; cargo --version; rustup --version
+      - restore_cache:
+          keys:
+            - cargocache-cw20-ics20-rust:1.50.0-{{ checksum "~/project/Cargo.lock" }}
+      - run:
+          name: Unit Tests
+          env: RUST_BACKTRACE=1
```
webmaster128, line 409: This does not work. It must be

```
    environment:
      RUST_BACKTRACE: 1
```

See https://circleci.com/docs/2.0/configuration-reference/

The `env` key is copied over from GotHub Action long time ago and does nothing here.

contracts/cw20-ics20/src/contract.rs
```diff
@@ -0,0 +1,343 @@
...
+};
+use crate::state::{Config, CHANNEL_INFO, CHANNEL_STATE, CONFIG};
+use cw0::{nonpayable, one_coin};
+
+// version info for migration info
+const CONTRACT_NAME: &str = "crates.io:cw20-ics20";
+const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn init(
```
webmaster128, line 23: Note this will conflict with https://github.com/CosmWasm/cosmwasm-plus/pull/243

ethanfrey: I need that merged in first and a rebase. The contract_checker failed in CI as I didn't have instantiate

ethanfrey:
Rebased with `v0.14.0-beta1`. Let's see if it passes CI now.

I'd like a review, but I guess I merge if no blockers and test with the relayer.

PR #240, opened 2021-03-08
Contract sanity checking

### LINKED ISSUE #228: Add contract sanity checking during build / release

maurolacy:
Closes #228.


.circleci/config.yml
```diff
@@ -780,11 +775,25 @@ jobs:
...
             - /usr/local/cargo/registry
             - target
           key: cargocache-wasm-rust:1.50.0-{{ checksum "~/project/Cargo.lock" }}
+      - run:
+          name: Compile wasm files
+          command: |
+            for W in ./target/wasm32-unknown-unknown/release/*.wasm
+            do
+              echo -n "Compiling `basename $W`... "
+              compile_module $W || exit $?
```
ethanfrey, line 795: Nice line.

And I like where this is placed. Doesn't add much more delay to the CI and ensures compatibility

.circleci/config.yml
```diff
@@ -340,12 +340,6 @@ jobs:
       - run:
           name: Build and run schema generator
           command: cargo schema --locked
-      - run:
```
ethanfrey, line 343: Should we remove the integration/test.rs file as well?
I would suggest that (and remove devDeps to keep CI fast)

ethanfrey (approved):
Looks good to merge once cosmwasm pr is merged and CI is happy

.circleci/config.yml
maurolacy: Yes, this is optimal in that sense. Also compiling this utility for debug, which is twice as fast.

.circleci/config.yml
maurolacy: I already removed the integ tests, but forgot about the dev deps. Thanks.

PR #243, opened 2021-03-10
Update to 0.14.0 beta1

### LINKED ISSUE #242: Update to CosmWasm v0.14.0-beta1
Let's get the latest release in here.

I think the only major change is the rename of init -> instantiate.

I also realized the wasm contract checker returns errors on my `-alpha2` formats (it is too advanced): https://app.circleci.com/pipelines/github/CosmWasm/cosmwasm-plus/955/workflows/9db7be4b-7003-46e5-89ff-67e698a3fa90/jobs/15471/parallel-runs/0/steps/0-108?invite=true

So this becomes a bit more urgent, so I can merge in other PRs.

maurolacy:
Closes #242.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -85,9 +85,9 @@ checksum = "9f6b64db6932c7e49332728e3a6bd82c6b7e16016607d20923b537c3bc4c0d5f"
 
 [[package]]
 name = "cosmwasm-crypto"
-version = "0.14.0-alpha2"
+version = "0.14.0-beta1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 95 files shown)


maurolacy:
Do we increase contract versions to 0.6.0, right?

ethanfrey:
> Do we increase contract / package versions to 0.6.0, right?

I only bump this when we do a release. It stays 0.5.0 until the next tag. Not sure if that is "proper' but that is how I've been doing it with the scripts and lets us pick the version number right before tagging.

I would call a v0.6.0-alpha1 once we got the cw20-ics20 contract merged. I want some real world tests on that and a bit more cleanup before v0.6.0

ethanfrey (approved):
Looks like a very comprehensive rename. Thanks for taking care of this.

PR #244, opened 2021-03-11
Fix ics20 denom

ethanfrey:
Properly handle parsing the voucher's denom when it is returned to us.

Thanks @alpe for finding this `port/channel/orig_denom`

Also add a migrate entry point that accepts any older cw20-ics20 contract as a migration target.


alpe (approved):
Very nice. Please see the comment about protocol/app level errors

PR #245, opened 2021-03-15
port erc-1155 to cosmwasm

yihuang:
https://github.com/CosmWasm/cosmwasm-plus/blob/8adba30729f0c8da16bdb1c834991380d111d365/packages/cw1155/README.md

Open issues:
- Should contract store metadata and token set (for enumeration) on chain, or it should be done by off-chain indexing service?


orkunkl:
Thank you I will review when I get the chance.

contracts/cw1155-base/src/contract.rs
```diff
@@ -0,0 +1,614 @@
...
+
+    let messages = match msg {
+        Some(msg) => vec![Cw1155ReceiveMsg {
+            operator: info.sender.clone(),
+            from: from.clone(),
+            amount,
+            token_id: token_id.clone(),
+            msg,
+        }
+        .into_cosmos_msg(to.clone().unwrap())?], // `to` in Send messages must not be None
```
tomtau, line 102: not sure if [MIRAI contracts](https://github.com/facebookexperimental/MIRAI#using-mirai-together-with-the-rust-design-by-contracts-crate) can be used to express this?

packages/cw1155/src/lib.rs
```diff
@@ -0,0 +1,17 @@
...
+mod query;
+mod receiver;
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn it_works() {
+        assert_eq!(2 + 2, 4);
+    }
+}
```
tomtau, line 17:
```suggestion
```

packages/cw1155/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+Any contract wish to receive CW1155 tokens must implement `Cw1155ReceiveMsg` and `Cw1155BatchReceiveMsg`.
+
+`Cw1155ReceiveMsg { operator, from, token_id, amount, msg}` - 
+
+`Cw1155BatchReceiveMsg { operator, from, batch, msg}` - 
+
+## Metadata and Enumerable
+
+[TODO] ERC1155 suggests that metadata and enumerable should be indexed from events log, to save some on-chain storage. Should we define standard events like ERC1155?
```
tomtau, line 48: probably ok, but not sure if in line with the rest of cosmwasm-plus, as a kind of the goal seemed to have been to allow direct querying in cosmjs https://github.com/cosmos/cosmjs/tree/main/packages/cosmwasm-stargate/src/queries etc.

tomtau (commented):
👍

packages/cw1155/README.md
ethanfrey: We do emit events (well attributes) on most actions. 
It would be good to properly define them to allow easy querying.

If you want to define the Events in the cw1155 spec, then we would be happy to review. It would set a good example and we could update the other modules to follow it.

packages/cw1155/src/lib.rs
ethanfrey: You need at least one test so `cargo test` doesn't error on this package.

Good to remove this as soon as you write any reasonable test of the real code.

packages/cw1155/src/msg.rs
```diff
@@ -0,0 +1,43 @@
...
+use serde::{Deserialize, Serialize};
+
+use cosmwasm_std::{Binary, HumanAddr, Uint128};
+
+pub type TokenId = String;
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
+#[serde(rename_all = "snake_case")]
+pub enum Cw1155HandleMsg {
+    /// TransferFrom is a base message to move tokens
```
ethanfrey, line 11: Interesting, there is no specific `Transfer` message, but if from is None, then we send from the `info.sender` account? What happens if the `to` field is `None`? I assume that is required, like `SendFrom.contract`

I'm a bit surprised to see Options here, can you just add a line of comment explaining what they do

packages/cw1155/src/receiver.rs
```diff
@@ -0,0 +1,73 @@
...
+use cosmwasm_std::{to_binary, Binary, CosmosMsg, HumanAddr, StdResult, Uint128, WasmMsg};
+
+use crate::msg::TokenId;
+
+/// Cw1155ReceiveMsg should be de/serialized under `Receive()` variant in a HandleMsg
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
+#[serde(rename_all = "snake_case")]
+pub struct Cw1155ReceiveMsg {
+    pub operator: HumanAddr,
+    pub from: Option<HumanAddr>,
```
ethanfrey, line 13: From should always be set, right? Otherwise you don't know who is calling this.

What is the different between from and operator?

ethanfrey (commented):
First pass, just looking at the spec.

Thanks for making this PR.

packages/cw1155/src/msg.rs
yihuang: > Both from and to are Option, if from is None, it means mint new tokens, if to is None, it means burn tokens. They must not both be None at the same time.

In the spec there's a line to explain, `Option` is used to mean minting(from is `None`) and burning(to is `None`).

packages/cw1155/src/receiver.rs
yihuang: When from is `None`, it means new tokens is minted directly to the receiver by operator.

packages/cw1155/README.md
```diff
@@ -0,0 +1,48 @@
...
+The specification is split into multiple sections, a contract may only
+implement some of this functionality, but must implement the base.
+
+## Base
+
+### Messages
+
+`TransferFrom{from, to, token_id, value}` - This transfers some amount of tokens between two accounts. The operator should have approval from the source account.
+
+Both `from` and `to` are `Option`, if `from` is `None`, it means mint new tokens, if `to` is `None`, it means burn tokens. They must not both be `None` at the same time.
```
ethanfrey, line 15: I now see the description here (I asked for it below). It would be good to have such essential docs also in the rustdoc on the types.

Having "transfer my tokens", "transfer your tokens", "mint new tokens", "burn my tokens" and "burn your tokens" all in the same message type does reduce some boilerplate, but would make it very easy to make mistakes. Especially since JSON ignores extra fields. (And the client sneds JSON)

If they accidentally use `rcpt` or `too` instead of `to`, this will burn the tokens. :scream: 

I see the idea of joining "send mine" and "send yours" as one type, but I would keep Mint and Burn as separate message types, so it is clear what is needed - and very unlikley to accidentally type "burn_from" instead of "transfer_from"

packages/cw1155/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+`SendFrom{from, to, token_id, value, msg}` - This transfers some amount of tokens between two accounts. `to`
```
ethanfrey, line 17: Two questions:

* There is no mention of allowances - for SendFrom to work from other accounts, we need those, right?
* This seems to support many different cw20-style tokens in one contract. Where is NFT support done?

ethanfrey (commented):
A few more comments.

I think it would be best to break this into 2 PRs.
Can you make one for the `packages/cw1155` first? Then we can discuss the design issues?

It doesn't make sense to review the code until the interface is clear. This way we could merge the package/spec quickly and then you work on updating the implementation to match in a separate PR

yihuang:
Thanks for reviewing, I'll make another PR soon:
- only specification
- more separated message branches for different cases, mint/burn/transfer/send, etc.
- data types for events(attributes).

Closing this one for now.

ethanfrey:
Thanks for the PRs. I will review the spec shortly.

yihuang:
> Thanks for the PRs. I will review the spec shortly.

Still choose to mix burn/mint with transfer messages, because otherwise, there'll need much more message types to be complete.

PR #247, opened 2021-03-17
Add cw1155 specification

### LINKED ISSUE #246: No equivalent of ERC1155 standard
- ERC1155 treats fungible and non-fungible tokens equally, non-fungible tokens just have one max supply.
- Metadata and token enumerations can be handled off-chain, by indexing attributes in block results.
- Approval is set to some operator for all tokens at once, no fine-granularity control over allowance.
  Implementation might assume approval for self automatically.
- Mint/burn/transfer are mixed in one set of messages, otherwise, there'll be too many message types.

yihuang:
retry of https://github.com/CosmWasm/cosmwasm-plus/pull/245
fixes #246 

Design decisions:
- Fungible tokens and non-fungible tokens are treated equally, non-fungible tokens just have one max supply.
- Transfer permission is approved or disapproved to some operator over entire set of tokens. (Scoped control is defined in [ERC1761](https://eips.ethereum.org/EIPS/eip-1761), we can port that separately)
- Metadata query and token enumeration should be done by indexing attributes in block-results off-chain.
- Mint and burn are mixed with transfer/send messages to keep number of message types small, otherwise we'll need much more message types like `Mint`/`MintToContract`/`BatchMint`/`BatchMintToContract`, etc.
  In transfer/send messges, `from`/`to` are optional, a `None` `from` means minting, a `None` `to` means burning, they must not both be `None` at the same time.


packages/cw0/src/event.rs
```diff
@@ -0,0 +1,5 @@
+use cosmwasm_std::Attribute;
+
+pub trait Event {
```
ethanfrey, line 3: Can you add rustdoc on this?

I am sure I will see how it is used a few files down, but if someone comes to it for the first time, nice to explain.

packages/cw1155/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+homepage = "https://cosmwasm.com"
+documentation = "https://docs.cosmwasm.com"
+
+[dependencies]
+cosmwasm-std = { version = "0.14.0-beta1" }
+schemars = "0.7"
+serde = { version = "1.0.103", default-features = false, features = ["derive"] }
+
+[dev-dependencies]
+cosmwasm-schema = { version = "0.14.0-alpha2" }
```
ethanfrey, line 18: Probably should be -beta1 as well for consistency. (Not that the schema package changed)

packages/cw1155/README.md
```diff
@@ -0,0 +1,63 @@
...
+CW1155 is a specification for managing multiple tokens based on CosmWasm.
+The name and design is based on Ethereum's ERC1155 standard.
+
+Design decisions:
+
+- Fungible tokens and non-fungible tokens are treated equally, non-fungible tokens just have one max supply.
+
+- Approval is set or unset to some operator over entire set of tokens. (More nuanced control is defined in [ERC1761](https://eips.ethereum.org/EIPS/eip-1761), do we want to merge them together?)
+
+- Metadata and token enumeration should be done by indexing events off-chain.
```
ethanfrey, line 12: 1. token enumeration should be done by indexing events off-chain

This is because it is impossible in Ethereum. We can do better. How about using enumeration like in cw721 on tokens or tokens/owner? or cw20 on allowances?

ethanfrey: 2. Off-chain metadata via events is okay. Adding one URL there or such would be a simple addition that seems like it is used various places.

Also, the "indexing events off-chain" should reference an example and some such indexer (I think you guys have built one, so maybe a nice place to link to it and say you could build on this)

packages/cw1155/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+- Mint and burn are mixed with transfer/send messages, otherwise, we'll have much more message types, e.g. `Mint`/`MintToContract`/`BatchMint`/`BatchMintToContract`, etc.
+
+  In transfer/send messges, `from`/`to` are optional, a `None` `from` means minting, a `None` `to` means burning, they must not both be `None` at the same time.
```
ethanfrey, line 16: This seems dangerous to me, as the input is JSON, and extra fields are ignored. So changing `from` to `form` turns a transfer to a mint (likely to fail anyway), but using `too` vs `to` will make you burn your tokens, rather than have the tx rejected by the chain.

@willclarktech and @maurolacy I think we discussed the issue of nullable fields changing functionality to be dangerous. Will especially in the context of CosmJS, some case when we said omitting a field from "ChangeOwner" would make it "ClearOwner" (making no one owner) could be an expensive mistake with a typo.

Can you please comment on this design here (or better look at the [corresponding type definitions](https://github.com/CosmWasm/cosmwasm-plus/pull/247/files#diff-07f199dd55ccb4e6b043c50dded9da214c860437339f9a73e0bea9ed89a85d72R12-R29)) and say if you approve or suggest ways to improve.

packages/cw1155/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Base
+
+### Messages
+
+`TransferFrom{from, to, token_id, value}` - This transfers some amount of tokens between two accounts. The operator should have approval from the source account.
+
+`SendFrom{from, to, token_id, value, msg}` - This transfers some amount of tokens between two accounts. `to`
```
ethanfrey, line 24: The fact that you combine `Send` and `SendFrom` is nice (just provide your own address as from to make it a normal send). Just not sure to throw in Mint/MintFrom/Burn/BurnFrom here

packages/cw1155/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Receiver
+
+Any contract wish to receive CW1155 tokens must implement `Cw1155ReceiveMsg` and `Cw1155BatchReceiveMsg`.
+
+`Cw1155ReceiveMsg { operator, from, token_id, amount, msg}` - 
+
+`Cw1155BatchReceiveMsg { operator, from, batch, msg}` - 
+
+### Events
```
ethanfrey, line 51: Very cool to specify the emitted events. We should do that on all contracts for the key events.

packages/cw1155/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+- `transfer(from, to, token_id, value)`
+
+  `from`/`to` are optional, no `from` attribute means minting, no `to` attribute means burning.
```
ethanfrey, line 55: Hmm.. I was going to suggest different events for mint and burn, but on reflection agree with you here. If you are searching for transfer events but the mint events are missing, you will get the wrong balances. 

This is a clever design to enable better indexing.

packages/cw1155/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+- `metadata(url, token_id)`
+
+  Metadata url of `token_id` is changed, `url` should point to a json file.
```
ethanfrey, line 59: Nice to emit this event. I would also store the current value of the URL in state. It is not that large. (You can max size it to 200 bytes or so to keep people from base64 encoding images into it)

packages/cw1155/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Metadata and Enumerable
+
+[TODO] ERC1155 suggests that metadata and enumerable should be indexed from events log, to save some on-chain storage. Should we define standard events like ERC1155?
```
ethanfrey, line 63: I think metadata is defined above, correct? So you can remove that part.

For enumerable, I would suggest just implementing it making use of the store iterators in the sdk. It is rather efficient.

packages/cw1155/src/event.rs
```diff
@@ -0,0 +1,42 @@
...
+use crate::msg::TokenId;
+
+pub struct TransferEvent<'a> {
+    pub from: Option<&'a HumanAddr>,
+    pub to: Option<&'a HumanAddr>,
+    pub token_id: TokenId,
+    pub amount: Uint128,
+}
+
+impl Event for TransferEvent {
```
ethanfrey, line 13: Very nice idea. I like it.

So, the standard defines some Event types and the contracts can just instantiate these types and "serialize" them into the response. After this PR is merged, I will look to see how to adapt other contracts to this pattern.

packages/cw1155/src/event.rs
```diff
@@ -0,0 +1,42 @@
...
+    fn write_attributes(&self, attributes: &mut Vec<Attribute>) {
```
ethanfrey, line 14: Response object now has an [`add_attribute` method](https://github.com/CosmWasm/cosmwasm/blob/main/packages/std/src/results/response.rs#L112-L117). Maybe this could make use of it?

```rust
fn write_attributes(&self, resp: &mut Response) {
  resp.add_attributute("action", "transfer");
  resp.add_attributute("token_id", self.token_id);
  // ...
}
```

Not that it does anything different, but seems a bit cleaner to read and sets a nice pattern for other devs. (And I would like your feedback on the method name `add_attribute`, we are still in beta, when it can be changed)

packages/cw1155/src/msg.rs
```diff
@@ -0,0 +1,48 @@
...
+    TransferFrom {
+        // `None` means minting
+        from: Option<HumanAddr>,
+        // `None` means burning
+        to: Option<HumanAddr>,
+        token_id: TokenId,
+        value: Uint128,
+    },
+    /// SendFrom is a base message to move tokens to contract
+    /// if `env.sender` has sufficient pre-approval.
```
ethanfrey, line 21: -> if `env.sender` is the owner or has sufficient pre-approval.

packages/cw1155/src/query.rs
```diff
@@ -0,0 +1,41 @@
...
+    /// Return type: BalanceResponse.
+    Balance { owner: HumanAddr, token_id: TokenId },
+    /// Returns the current balance of the given address for a batch of tokens, 0 if unset.
+    /// Return type: BatchBalanceResponse.
+    BatchBalance {
+        owner: HumanAddr,
+        token_ids: Vec<TokenId>,
+    },
+    /// Queries the approval status of an operator for a given owner
+    /// Return type: ApprovedForAllResponse.
```
ethanfrey, line 21: Again from cw721, how about iterating (page-wise) over all approvals. Here is a template for one way (and you can look at the implementation in cw721-base): https://github.com/CosmWasm/cosmwasm-plus/blob/master/packages/cw721/src/query.rs#L17-L25

packages/cw1155/src/msg.rs
```diff
@@ -0,0 +1,48 @@
...
+    /// BatchSendFrom is a base message to move tokens to another to without triggering actions
+    BatchSendFrom {
+        // `None` means minting
+        from: Option<HumanAddr>,
+        contract: HumanAddr,
+        batch: Vec<(TokenId, Uint128)>,
+        msg: Option<Binary>,
+    },
+    /// Enable or disable approval for a third party ("operator") to manage all of the caller's tokens.
+    SetApprovalForAll { operator: HumanAddr, approved: bool },
```
ethanfrey, line 47: In cw721 we took a few liberties to "improve" the Ethereum design, even with the "all or nothing" approval. How about adding an [optional expiration date](https://github.com/CosmWasm/cosmwasm-plus/blob/master/packages/cw721/src/msg.rs#L38)?

Adding approvals on token by token basis would be nice as well, even if it is full control of the balance on that token_id. Are you concerned about performance? Or prefer the API this way? Especially iteration is MUCH more efficient in CosmWasm/Cosmos SDK than Ethereum and it is nice to embrace it.

packages/cw1155/src/receiver.rs
```diff
@@ -0,0 +1,73 @@
...
+use cosmwasm_std::{to_binary, Binary, CosmosMsg, HumanAddr, StdResult, Uint128, WasmMsg};
+
+use crate::msg::TokenId;
+
+/// Cw1155ReceiveMsg should be de/serialized under `Receive()` variant in a HandleMsg
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
+#[serde(rename_all = "snake_case")]
+pub struct Cw1155ReceiveMsg {
+    pub operator: HumanAddr,
+    pub from: Option<HumanAddr>,
```
ethanfrey, line 13: Since this is an interface other devs will have to support, good to be explicit.

In this case `operator` was the one who actually executed the message and `from` is the account the funds were sent from, correct? If so, please just add some RustDoc strings here.

ethanfrey: Otherwise, nice job in this file.

ethanfrey (commented):
Very nice and clean implementation and I really like some of the ideas. Especially events, but also merging eg. Send/SendFrom into one message type (avoids lots of duplication).

Most of the comments are minor suggestions to polish off the spec. The only real issue is combining send/burn/mint into one message. I would like a 2nd (or 3rd) opinion here. This is really about API design not coding.

Also, what do you think in general of extending the Ethereum spec to match our framework a bit I mean, you use Uint128 rather than Uint256 (128 bit is plenty large enough, and they just have 256 cuz everything is 256bit in Ethereum). It is a chance to take these same ideas and enhance them with some of the capabilities of CosmWasm (especially cheap iteration over parts of a set)

packages/cw1155/README.md
yihuang: - Added the `Metadata` extension query apis, similar to CW721.
- Changed event name from `metadata` to `token_info`, to keep it consistent with CW721 naming.

packages/cw1155/README.md
yihuang: I added metadata and enumerable extension similar to the ones in CW721.

packages/cw1155/src/event.rs
yihuang: I'm not very good at picking english names, I think `add_attribute` is good, I changed the method name in `Event` trait to `add_attribututes` too.
Another possible choice here is `fn iter_attributes(&self, resp: &mut Response) -> impl Iterator<Item=Attribute>`.

packages/cw1155/src/msg.rs
yihuang: done

packages/cw1155/src/query.rs
yihuang: Looks good, copied those from CW721 spec.

packages/cw1155/src/receiver.rs
yihuang: done

packages/cw1155/src/msg.rs
yihuang: I think the only trade-off is use cases vs simplicity and on-chain resource usage.
For example, for the [example cw721 contract](https://github.com/CosmWasm/cosmwasm-plus/blob/master/contracts/cw721-base/src/state.rs#L13) to support approval by token, it needs to store a `Vec<Approval>` in `TokenInfo`, and need to do `O(n)` search when checking permission.

packages/cw1155/README.md
yihuang: I added metadata and enumerable extensions like cw721, it's indeed more convenient for client I think.

packages/cw1155/Cargo.toml
yihuang: done

packages/cw0/src/event.rs
yihuang: done

yihuang:
> Very nice and clean implementation and I really like some of the ideas. Especially events, but also merging eg. Send/SendFrom into one message type (avoids lots of duplication).
> 
> Most of the comments are minor suggestions to polish off the spec. The only real issue is combining send/burn/mint into one message. I would like a 2nd (or 3rd) opinion here. This is really about API design not coding.
> 
> Also, what do you think in general of extending the Ethereum spec to match our framework a bit I mean, you use Uint128 rather than Uint256 (128 bit is plenty large enough, and they just have 256 cuz everything is 256bit in Ethereum). It is a chance to take these same ideas and enhance them with some of the capabilities of CosmWasm (especially cheap iteration over parts of a set)

- Add metadata and enumeable query extension similar to cw721
- Modified ApprovalForAll similar to cw721
- Add some comments

Open issues:

- Should we mix mint/burn messages with transfer/send
- Should we support approval by token, like cw721

packages/cw1155/src/msg.rs
ethanfrey: Fair enough. This does add more runtime complexity. We can leave it as is.

ethanfrey:
Thanks for the updates, looking very nice.

> Should we support approval by token, like cw721?

I don't know if this use case will be used much or at all and it does indeed add runtime overhead to every check. I am fine to leave it off.

> Should we mix mint/burn messages with transfer/send? 

This is my only outstanding concern, and I would like the comment of another developer with CosmWasm experience here. Otherwise, it looks good.

packages/cw1155/README.md
```diff
@@ -3,42 +3,47 @@
...
 
 ### Messages
 
-`TransferFrom{from, to, token_id, value}` - This transfers some amount of tokens between two accounts. The operator should have approval from the source account.
+`TransferFrom{from, to, token_id, value}` - This transfers some amount of tokens between two accounts. The operator should either be the `from` account or have approval from it.
 
 `SendFrom{from, to, token_id, value, msg}` - This transfers some amount of tokens between two accounts. `to` 
 must be an address controlled by a smart contract, which implements
-the `CW1155Receiver` interface. The operator should have approval from the source account. The `msg` will be passed to the recipient contract, along with the other fields.
+the `CW1155Receiver` interface. The operator should eitherbe the `from` account or have approval from it. The `msg` will be passed to the recipient contract, along with the other fields.
```
ethanfrey, line 27: either be (missing space)

packages/cw1155/README.md
```diff
@@ -3,42 +3,47 @@
...
 `Balance { owner, token_id }` - Query the balance of `owner` on perticular type of token, default to `0` when record not exist.
 
-`BatchBalance { owner, token_ids }` - Query the balance of `owner` on multiple types of tokens, batched version of `Balance`
+`BatchBalance { owner, token_ids }` - Query the balance of `owner` on multiple types of tokens, batched version of `Balance`.
 
-`ApprovedForAll{ owner, spender }` - Query if `spender` has the permission to transfer or send tokens owned by `msg.sender`.
+`ApprovedForAll{owner, include_expired, start_after, limit}` - List all operators that can
+access all of the owner's tokens. Return type is `ApprovedForAllResponse`.
+If `include_expired` is set, show expired owners in the results, otherwise,
+ignore them. If query for some specific operator, set `start_after` to the operator, and set limit to 1.
```
ethanfrey, line 46: Actually interesting point - start_after does not include that value itself. It is great for pagination, but not good for finding one in particular. Maybe you do want the other query as well to just get one.

ethanfrey (commented):
Looks good minus the discussion on separating Transfer/Mint/Burn.

If you wish to reduce types, another thought would be to combine TransferFrom and SendFrom.

eg.

```rust
/// This handles contract call or user account
SendFrom {
  from: HumanAddr, // either env.sender self or another account sender is approved on
  to: HumanAddr, // an external account or contract 
  token_id: TokenId,
  value: Uint128,
  // if msg is None, just transfer the tokens.
  // if msg is Some(m), then call a Cw1155 receiver on a contract with the `to` address (after moving the token)
  msg: Option<Binary>,
}
```

And then add other types for minting and burning (or at least burning).
This would allow you to reduce the number of types and address my concern about possibly mixing burning and send.

If you want to still keep `from: None` to mean `mint` I have less issues than with burn. In the end maybe you can end up with the followng types:

* `SendFrom` - mints or transfers from my account or transfers from other account. If there is a message, it will assume the target is a contract and invoke the cw1155 receiver on it.
* `BatchSendFrom` - like SendFrom but with a list of tokens
* `BurnFrom` - burns one token either from my account or an account i was authorized on
* `ApproveAll`
* `RevokeAll`

packages/cw1155/README.md
willclarktech: What about minting/burning if the null address is provided, like Ethereum's `0x0000000000000000000000000000000000000000`? Then at least you have to specify it. Otherwise I'd suggest `mint`/`burn` boolean options to be explicit.

yihuang:
> ```
>   // if msg is None, just transfer the tokens.
>  // if msg is Some(m), then call a Cw1155 receiver on a contract with the `to` address (after moving the token)
>  msg: Option<Binary>,
> ```

I think it needs to be `Option<Option<Binary>>`:
- `None`, normal recipient address
- `Some(None)`, recipient address is contract, but send `None` as extra msg.
- `Some(Some(msg))`, recipient address is contract, send extra msg.

> `BurnFrom` - burns one token either from my account or an account i was authorized on

I guess we probably also want `BatchBurnFrom`?

packages/cw1155/README.md
yihuang: added `ApprovedForAllItem`.

ethanfrey:
> I think it needs to be Option<Option<Binary>>:

`Option<Option<Binary>>` and `Option<Binary>` have the same JSON representation.

There is no real use case when we send a message to a contract with data `None`. You can always send `Some(b"{}")` for the cases when it doesn't matter (and ensure you have a valid JSON object as body)

packages/cw1155/src/query.rs
```diff
@@ -27,6 +27,12 @@ pub enum Cw1155QueryMsg {
         start_after: Option<HumanAddr>,
         limit: Option<u32>,
     },
+    /// Query approved status `owner` granted to `operator`.
+    /// Return type: ApprovedForAllItemResponse
+    ApprovedForAllItem {
```
ethanfrey, line 32: I would rename to `IsApprovedForAll` that would suggest a boolean as a response.

The other one is a bit vague (calling it `ListApprovedForAll` would be more explicit), but I think clear enough.

yihuang:
> > I think it needs to be Option<Option>:
> 
> `Option<Option<Binary>>` and `Option<Binary>` have the same JSON representation.
> 
> There is no real use case when we send a message to a contract with data `None`. You can always send `Some(b"{}")` for the cases when it doesn't matter (and ensure you have a valid JSON object as body)

But the receiver msg contains `Option<Binary>`, do you think we should remove the `Option` here too?

packages/cw1155/src/query.rs
yihuang: > I would rename to `IsApprovedForAll` that would suggest a boolean as a response.

that's better, thanks.

> The other one is a bit vague (calling it `ListApprovedForAll` would be more explicit), but I think clear enough.

I think current overall naming conversion is using nouns for message branch names, so I guess we can hold this one for now?

ethanfrey:
> But the receiver msg contains Option<Binary>, do you think we should remove the Option here too?

You are right. I designed the cw20 receiver interface before implementing any contracts. It would be good to remove Option there.

packages/cw1155/README.md
yihuang: done

packages/cw1155/README.md
yihuang: Now, mint/burn messages are separated.

yihuang:
- Merged TransferFrom into SendFrom
- Separated out Mint/Burn messages

ethanfrey (approved):
Thank you for the changes

packages/cw1155/README.md
```diff
@@ -54,10 +54,10 @@ Any contract wish to receive CW1155 tokens must implement `Cw1155ReceiveMsg` and
 
   `from`/`to` are optional, no `from` attribute means minting, no `to` attribute means burning.
 
-- `metadata(url, token_id)`
+- `token_info(url, token_id)`
```
ThienLK1: In event.rs i only see MetadataEvent, not found token_info.
And i dont see any api using this event.
I just want add uri for token, but dont find any api do this?

yihuang: metadata modification is not part of the standard API, different contracts can implement their own APIs to do that I think.

ThienLK1: Do you mean base on ERC1155Metadata?
Do you think it can fulfill what's missing?

PR #248, opened 2021-03-18
cw20-base: upgrade helper.ts to cosmjs 0.24.1

orkunkl:
WIP

### DIFF
--- a/contracts/cw20-base/helpers.ts
+++ b/contracts/cw20-base/helpers.ts
@@ -8,7 +8,7 @@ import path from "path"
  * This is a set of helpers meant for use with @cosmjs/cli
  * With these you can easily use the cw20 contract without worrying about forming messages and parsing queries.
  *
- * Usage: npx @cosmjs/cli@^0.23 --init https://raw.githubusercontent.com/CosmWasm/cosmwasm-plus/master/contracts/cw20-base/helpers.ts
+ * Usage: npx @cosmjs/cli@^0.25 --init https://raw.githubusercontent.com/CosmWasm/cosmwasm-plus/master/contracts/cw20-base/helpers.ts
  *
  * Create a client:
  *   const [addr, client] = await useOptions(oysternetOptions).setup('password');
--- a/contracts/cw20-base/helpers.ts
+++ b/contracts/cw20-base/helpers.ts
@@ -2,7 +2,7 @@
  * This is a set of helpers meant for use with @cosmjs/cli
  * With these you can easily use the cw20 contract without worrying about forming messages and parsing queries.
  * 
- * Usage: npx @cosmjs/cli@^0.23 --init https://raw.githubusercontent.com/CosmWasm/cosmwasm-plus/master/contracts/cw20-base/helpers.ts
+ * Usage: npx @cosmjs/cli@^0.24 --init https://raw.githubusercontent.com/CosmWasm/cosmwasm-plus/master/contracts/cw20-base/helpers.ts
  * 
  * Create a client:
  *   const client = await useOptions(hackatomOptions).setup(password);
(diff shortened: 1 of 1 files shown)


contracts/cw20-base/helpers.ts
```diff
@@ -67,47 +78,40 @@ const useOptions = (options: Options): Network => {
...
     await axios.post(faucetUrl, { ticker, address });
   }
-  
-  const setup = async (password: string, filename?: string): Promise<SigningCosmWasmClient> => {
+
+  const setup = async (password: string, filename?: string): Promise<CW20Client> => {
     const keyfile = filename || options.defaultKeyFile;
     const wallet = await loadOrCreateWallet(hackatomOptions, keyfile, password);
     const client = await connect(wallet, hackatomOptions);
+    const account = (await wallet.getAccounts())[0].address;
```
webmaster128, line 105: Better call this `address`

contracts/cw20-base/helpers.ts
```diff
@@ -67,47 +78,40 @@ const useOptions = (options: Options): Network => {
...
 
     // ensure we have some tokens
     if (options.faucetUrl) {
-      const account = await client.getAccount();
       if (!account) {
```
webmaster128, line 109: This is always false now, because `account` is always a non-empty address. I.e. you never get tokens

contracts/cw20-base/helpers.ts
```diff
@@ -67,47 +78,40 @@ const useOptions = (options: Options): Network => {
...
+
+  const setup = async (password: string, filename?: string): Promise<CW20Client> => {
     const keyfile = filename || options.defaultKeyFile;
     const wallet = await loadOrCreateWallet(hackatomOptions, keyfile, password);
     const client = await connect(wallet, hackatomOptions);
+    const account = (await wallet.getAccounts())[0].address;
 
     // ensure we have some tokens
     if (options.faucetUrl) {
-      const account = await client.getAccount();
```
webmaster128, line 103: This tells you whether or not the account exists on chain. You probably want to keep it

contracts/cw20-base/helpers.ts
```diff
@@ -180,75 +184,75 @@ interface CW20Contract {
...
 }
 
 
-const CW20 = (client: SigningCosmWasmClient): CW20Contract => {
+const CW20 = (cw20Client: CW20Client): CW20Contract => {
   const use = (contractAddress: string): CW20Instance => {
     const balance = async (account?: string): Promise<string> => {
-      const address = account || client.senderAddress;  
-      const result = await client.queryContractSmart(contractAddress, {balance: { address }});
+      const address = account || (await cw20Client.wallet.getAccounts())[0].address;
```
webmaster128, line 197: You stored it here:

```suggestion
      const address = account || cw20Client.sender;
```

orkunkl:
An external contributor fixed helper.ts and sent to me. This version works now.

contracts/cw20-base/helpers.ts
```diff
@@ -1,20 +1,24 @@
...
+import { Slip10RawIndex } from "@cosmjs/crypto"
+import path from "path"
 /*
  * This is a set of helpers meant for use with @cosmjs/cli
  * With these you can easily use the cw20 contract without worrying about forming messages and parsing queries.
- * 
- * Usage: npx @cosmjs/cli@^0.23 --init https://raw.githubusercontent.com/CosmWasm/cosmwasm-plus/master/contracts/cw20-base/helpers.ts
- * 
+ *
+ * Usage: npx @cosmjs/cli@^0.25 --init https://raw.githubusercontent.com/CosmWasm/cosmwasm-plus/master/contracts/cw20-base/helpers.ts
```
ethanfrey, line 11: :+1:

contracts/cw20-base/helpers.ts
ethanfrey: Checking if the account is empty (if they had tokens earlier but spent them all) is a fair enough check.
It's on the client side, so being stricter doesn't add security

ethanfrey (approved):
Looks good. Thanks

PR #249, opened 2021-03-22
Migrate to entry_point macro

### LINKED ISSUE #230: Use #[entry_point] macro in contracts
This is now the most flexible way to do it and has been demoed in #216

Let's adopt it for the contracts (note we need to add some cfg statement so we only compile if wasm target AND not library feature) This is demoed in the linked PR

maurolacy:
Closes #230.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -834,7 +834,7 @@ jobs:
           # Uses --debug for compilation speed
           # FIXME: Change when `check_contract` (part of `cosmwasm-0.14.0`) is published
           #command: cargo install --debug --features iterator --example check_contract -- cosmwasm-vm
-          command: cargo install --debug --features iterator --git https://github.com/CosmWasm/cosmwasm --branch=main --example check_contract -- cosmwasm-vm
+          command: cargo install --debug --features iterator --git https://github.com/CosmWasm/cosmwasm --tag=v0.14.0-beta1 --example check_contract -- cosmwasm-vm
       - save_cache:
           paths:
             - /usr/local/cargo/registry
--- a/contracts/cw1-subkeys/src/contract.rs
+++ b/contracts/cw1-subkeys/src/contract.rs
@@ -2,9 +2,11 @@ use schemars::JsonSchema;
 use std::fmt;
 use std::ops::{AddAssign, Sub};
 
+#[cfg(not(feature = "library"))]
+use cosmwasm_std::entry_point;
 use cosmwasm_std::{
-    attr, entry_point, to_binary, BankMsg, Binary, CanonicalAddr, Coin, CosmosMsg, Deps, DepsMut,
-    Empty, Env, HumanAddr, MessageInfo, Order, Response, StakingMsg, StdError, StdResult,
+    attr, to_binary, BankMsg, Binary, CanonicalAddr, Coin, CosmosMsg, Deps, DepsMut, Empty, Env,
+    HumanAddr, MessageInfo, Order, Response, StakingMsg, StdError, StdResult,
 };
 use cw0::{maybe_canonical, Expiration};
 use cw1::CanExecuteResponse;
@@ -30,7 +32,7 @@ use cw_storage_plus::Bound;
 const CONTRACT_NAME: &str = "crates.io:cw1-subkeys";
 const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
 
-#[entry_point]
+#[cfg_attr(not(feature = "library"), entry_point)]
 pub fn instantiate(
     mut deps: DepsMut,
     env: Env,
@@ -42,7 +44,7 @@ pub fn instantiate(
     Ok(result)
 }
 
-#[entry_point]
+#[cfg_attr(not(feature = "library"), entry_point)]
 pub fn execute(
     deps: DepsMut,
     env: Env,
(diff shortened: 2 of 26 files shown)


contracts/cw1-subkeys/src/contract.rs
```diff
@@ -30,6 +30,7 @@ use cw_storage_plus::Bound;
 const CONTRACT_NAME: &str = "crates.io:cw1-subkeys";
 const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
 
+#[entry_point]
```
ethanfrey, line 33: Can you use `#[cfg_attr(not(feature = "library"), entry_point)]` as you did for whitelist?

contracts/cw1-whitelist/src/contract.rs
```diff
@@ -16,6 +19,7 @@ use crate::state::{AdminList, ADMIN_LIST};
 const CONTRACT_NAME: &str = "crates.io:cw1-whitelist";
 const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
 
+#[cfg_attr(not(feature = "library"), entry_point)]
```
ethanfrey, line 22: This is proper, and should be the default for all, whether or not se currently import them as libraries, they should be importable as libraries.

contracts/cw1-whitelist/src/contract.rs
```diff
@@ -1,10 +1,13 @@
 use schemars::JsonSchema;
 use std::fmt;
 
+#[cfg(not(feature = "library"))]
```
ethanfrey, line 4: Very good point here, I think I messed up in other contracts, making this a default import.

contracts/cw20-bonding/src/contract.rs
```diff
@@ -23,6 +23,7 @@ use cw0::{must_pay, nonpayable};
 const CONTRACT_NAME: &str = "crates.io:cw20-bonding";
 const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
 
+#[entry_point]
```
ethanfrey, line 26: As above, please gate all of these `entry_point`s in all contracts with library feature

ethanfrey (changes requested):
Please don't use the plain `#[entry_point]` decorator, but rather the library feature flag version. I actually asked Simon to make that the default functionality (disable on library flag), but he thought that was putting too much app-specific logic in cosmwasm base. 

If you find the current status too annoying, please look at the original cosmwasm PR; https://github.com/CosmWasm/cosmwasm/pull/701#issuecomment-757221171 and re-open this issue. I just accepted using this larger macro heading everywhere.

contracts/cw1-subkeys/src/contract.rs
maurolacy: Sure. I did that only when currently necessary. I agree that it's more robust / forward-looking to do it by default.

contracts/cw1-whitelist/src/contract.rs
maurolacy: Yes, this triggers an `unused_imports` warning if not present, when importing as library.

maurolacy:
> If you find the current status too annoying, please look at the original cosmwasm PR; [CosmWasm/cosmwasm#701 (comment)](https://github.com/CosmWasm/cosmwasm/pull/701#issuecomment-757221171) and re-open this issue. I just accepted using this larger macro heading everywhere.

It's OK with me. Maybe the macro can be written in a way that has this check inside? (just curious)

maurolacy:
Will create a new branch to migrate cw-plus to Rust 0.51.0, as there are a number of linting issues to deal with.

ethanfrey:
> It's OK with me. Maybe the macro can be written in a way that has this check inside? (just curious)

Of course there is, the macro could also emit nothing if the "library" feature flag was set. Simon was worried that we would then be adding "magic" feature flag support into the macros, and said to do that in this repo. I see his point (make it explicit) and your point (make it easy)

ethanfrey:
> Will create a new branch to migrate cw-plus to Rust 0.51.0, as there are a number of linting issues to deal with.

Sounds good. So, you will do that first and then rebase this on it? I will approve it, so as soon as CI passes, I am happy.

ethanfrey (approved):
Looks good. As soon as the CI passes (I think the build script just needs to be updated to Rust 1.51)

maurolacy:
This was caused by thec `check_contract` dep being built from main istead of from a tag.

Changed it so we can merge this independently of the Rust 1.51.0 migration (than @ethanfrey).

.circleci/config.yml
```diff
@@ -834,7 +834,7 @@ jobs:
           # Uses --debug for compilation speed
           # FIXME: Change when `check_contract` (part of `cosmwasm-0.14.0`) is published
           #command: cargo install --debug --features iterator --example check_contract -- cosmwasm-vm
-          command: cargo install --debug --features iterator --git https://github.com/CosmWasm/cosmwasm --branch=main --example check_contract -- cosmwasm-vm
+          command: cargo install --debug --features iterator --git https://github.com/CosmWasm/cosmwasm --tag=v0.14.0-beta1 --example check_contract -- cosmwasm-vm
```
ethanfrey, line 837: :rocket:

PR #251, opened 2021-03-23
implement demo cw1155 contract

yihuang:
there are several more unit tests to be done, the overall implementation is good for review now.
There are two `FIXME` currently:
- [code](https://github.com/CosmWasm/cosmwasm-plus/pull/251/files#diff-7a4ab5d1010ab74822d59ab1b9c2de4d53f46b0d1b7c88c3ba5fdff7ba186ddcR90), we should add overflow error branch in `StdError`.
- [code](https://github.com/CosmWasm/cosmwasm-plus/pull/251/files#diff-7a4ab5d1010ab74822d59ab1b9c2de4d53f46b0d1b7c88c3ba5fdff7ba186ddcR1102), the list returned by query is out of order, I'm not sure if it's an issue of the test storage itself?

### DIFF
--- a/contracts/cw1155-base/Cargo.toml
+++ b/contracts/cw1155-base/Cargo.toml
@@ -1,0 +1,31 @@
+[package]
+name = "cw1155-base"
+version = "0.6.0-alpha1"
+authors = ["Huang Yi <huang@crypto.com>"]
+edition = "2018"
+description = "Basic implementation of a CosmWasm-1155 compliant token"
+license = "Apache-2.0"
+repository = "https://github.com/CosmWasm/cosmwasm-plus"
+homepage = "https://cosmwasm.com"
+documentation = "https://docs.cosmwasm.com"
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[features]
+backtraces = ["cosmwasm-std/backtraces"]
+# use library feature to disable all init/handle/query exports
+library = []
+
+[dependencies]
+cw0 = { path = "../../packages/cw0", version = "0.6.0-alpha1" }
+cw2 = { path = "../../packages/cw2", version = "0.6.0-alpha1" }
+cw1155 = { path = "../../packages/cw1155", version = "0.6.0-alpha1" }
+cw-storage-plus = { path = "../../packages/storage-plus", version = "0.6.0-alpha1" , features = ["iterator"]}
+cosmwasm-std = { version = "0.14.0-beta1", features = ["iterator"] }
+schemars = "0.7"
+serde = { version = "1.0.103", default-features = false, features = ["derive"] }
+thiserror = { version = "1.0.20" }
+
+[dev-dependencies]
+cosmwasm-schema = { version = "0.14.0-alpha2" }
--- a/contracts/cw1155-base/Cargo.toml
+++ b/contracts/cw1155-base/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cw1155-base"
-version = "0.6.0-alpha1"
+version = "0.6.0-alpha3"
 authors = ["Huang Yi <huang@crypto.com>"]
 edition = "2018"
 description = "Basic implementation of a CosmWasm-1155 compliant token"
@@ -18,10 +18,10 @@ backtraces = ["cosmwasm-std/backtraces"]
 library = []
 
 [dependencies]
-cw0 = { path = "../../packages/cw0", version = "0.6.0-alpha1" }
-cw2 = { path = "../../packages/cw2", version = "0.6.0-alpha1" }
-cw1155 = { path = "../../packages/cw1155", version = "0.6.0-alpha1" }
-cw-storage-plus = { path = "../../packages/storage-plus", version = "0.6.0-alpha1" , features = ["iterator"]}
+cw0 = { path = "../../packages/cw0", version = "0.6.0-alpha3" }
+cw2 = { path = "../../packages/cw2", version = "0.6.0-alpha3" }
+cw1155 = { path = "../../packages/cw1155", version = "0.6.0-alpha3" }
+cw-storage-plus = { path = "../../packages/storage-plus", version = "0.6.0-alpha3" , features = ["iterator"]}
 cosmwasm-std = { version = "0.14.0-beta1", features = ["iterator"] }
 schemars = "0.7"
 serde = { version = "1.0.103", default-features = false, features = ["derive"] }
--- a/contracts/cw1155-base/src/contract.rs
+++ b/contracts/cw1155-base/src/contract.rs
@@ -1,3 +1,4 @@
+use cosmwasm_std::entry_point;
 use cosmwasm_std::{
     to_binary, Api, Binary, Deps, DepsMut, Env, HumanAddr, MessageInfo, Order, Response, StdResult,
     Uint128, KV,
@@ -23,6 +24,7 @@ const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
 const DEFAULT_LIMIT: u32 = 10;
 const MAX_LIMIT: u32 = 30;
 
+#[cfg_attr(not(feature = "library"), entry_point)]
 pub fn instantiate(
     deps: DepsMut,
     _env: Env,
@@ -42,6 +44,7 @@ pub struct ExecuteEnv<'a> {
     info: MessageInfo,
 }
 
+#[cfg_attr(not(feature = "library"), entry_point)]
 pub fn execute(
     deps: DepsMut,
     env: Env,
@@ -85,7 +88,9 @@ pub fn execute(
 
 /// When from is None: mint new coins
 /// When to is None: burn coins
-/// When both are None: not token balance is changed, meaningless but valid
+/// When both are None: no token balance is changed, pointless but valid
+///
+/// Make sure permissions are checked before calling this.
 fn execute_transfer_inner<'a>(
     deps: &'a mut DepsMut,
     from: Option<&'a HumanAddr>,
@@ -140,14 +145,6 @@ fn check_can_approve(
     })
 }
 
-fn guard(cond: bool, err: ContractError) -> Result<(), ContractError> {
-    if cond {
-        Ok(())
-    } else {
-        Err(err)
-    }
-}
-
 pub fn execute_send_from(
     env: ExecuteEnv,
     from: HumanAddr,
@@ -162,18 +159,17 @@ pub fn execute_send_from(
         info,
     } = env;
 
-    guard(
-        check_can_approve(deps.as_ref(), &env, &from, &info.sender)?,
-        ContractError::Unauthorized {},
-    )?;
+    if !check_can_approve(deps.as_ref(), &env, &from, &info.sender)? {
+        return Err(ContractError::Unauthorized {});
+    }
 
     let mut rsp = Response::default();
 
     let event = execute_transfer_inner(&mut deps, Some(&from), Some(&to), &token_id, amount)?;
     event.add_attributes(&mut rsp);
 
-    rsp.messages = if let Some(msg) = msg {
-        vec![Cw1155ReceiveMsg {
+    if let Some(msg) = msg {
+        rsp.messages = vec![Cw1155ReceiveMsg {
             operator: info.sender,
             from: Some(from.clone()),
             amount,
@@ -181,9 +177,7 @@ pub fn execute_send_from(
             msg,
         }
         .into_cosmos_msg(to)?]
-    } else {
-        vec![]
-    };
+    }
 
     Ok(rsp)
 }
@@ -207,8 +201,8 @@ pub fn execute_mint(
     let event = execute_transfer_inner(&mut deps, None, Some(&to), &token_id, amount)?;
     event.add_attributes(&mut rsp);
 
-    rsp.messages = if let Some(msg) = msg {
-        vec![Cw1155ReceiveMsg {
+    if let Some(msg) = msg {
+        rsp.messages = vec![Cw1155ReceiveMsg {
             operator: info.sender,
             from: None,
             amount,
@@ -216,9 +210,7 @@ pub fn execute_mint(
             msg,
         }
         .into_cosmos_msg(to)?]
-    } else {
-        vec![]
-    };
+    }
 
     // insert if not exist
     let key = TOKENS.key(&token_id);
@@ -242,10 +234,9 @@ pub fn execute_burn(
     } = env;
 
     // whoever can transfer these tokens can burn
-    guard(
-        check_can_approve(deps.as_ref(), &env, &from, &info.sender)?,
-        ContractError::Unauthorized {},
-    )?;
+    if !check_can_approve(deps.as_ref(), &env, &from, &info.sender)? {
+        return Err(ContractError::Unauthorized {});
+    }
 
     let mut rsp = Response::default();
     let event = execute_transfer_inner(&mut deps, Some(&from), None, &token_id, amount)?;
@@ -266,10 +257,9 @@ pub fn execute_batch_send_from(
         info,
     } = env;
 
-    guard(
-        check_can_approve(deps.as_ref(), &env, &from, &info.sender)?,
-        ContractError::Unauthorized {},
-    )?;
+    if !check_can_approve(deps.as_ref(), &env, &from, &info.sender)? {
+        return Err(ContractError::Unauthorized {});
+    }
 
     let mut rsp = Response::default();
     for (token_id, amount) in batch.iter() {
@@ -277,16 +267,14 @@ pub fn execute_batch_send_from(
         event.add_attributes(&mut rsp);
     }
 
-    rsp.messages = if let Some(msg) = msg {
-        vec![Cw1155BatchReceiveMsg {
+    if let Some(msg) = msg {
+        rsp.messages = vec![Cw1155BatchReceiveMsg {
             operator: info.sender,
             from: Some(from),
             batch,
             msg,
         }
         .into_cosmos_msg(to)?]
-    } else {
-        vec![]
     };
 
     Ok(rsp)
@@ -309,26 +297,23 @@ pub fn execute_batch_mint(
     for (token_id, amount) in batch.iter() {
         let event = execute_transfer_inner(&mut deps, None, Some(&to), &token_id, *amount)?;
         event.add_attributes(&mut rsp);
-    }
 
-    for (token_id, _) in batch.iter() {
         // insert if not exist
         let key = TOKENS.key(&token_id);
         if deps.storage.get(&key).is_none() {
+            // insert an empty entry so token enumeration can find it
             key.save(deps.storage, &"".to_owned())?;
         }
     }
 
-    rsp.messages = if let Some(msg) = msg {
-        vec![Cw1155BatchReceiveMsg {
+    if let Some(msg) = msg {
+        rsp.messages = vec![Cw1155BatchReceiveMsg {
             operator: info.sender,
             from: None,
             batch,
             msg,
         }
         .into_cosmos_msg(to)?]
-    } else {
-        vec![]
     };
 
     Ok(rsp)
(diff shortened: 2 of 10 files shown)


contracts/cw1155-base/src/contract.rs
```diff
@@ -0,0 +1,1283 @@
...
+        } => execute_burn(env, from, token_id, value),
+        Cw1155HandleMsg::BatchBurn { from, batch } => execute_batch_burn(env, from, batch),
+        Cw1155HandleMsg::ApproveAll { operator, expires } => {
+            execute_approve_all(env, operator, expires)
+        }
+        Cw1155HandleMsg::RevokeAll { operator } => execute_revoke_all(env, operator),
+    }
+}
+
+fn checked_add(a: Uint128, b: Uint128) -> StdResult<Uint128> {
```
ethanfrey, line 86: Interesting. We define `sub` for `Uint128` to possibly return an error if it underflows (not enough balance). We don't do this for adding as we assumed this would never happen. 

We do however, [test that this extremely rare condition causes a panic](https://github.com/CosmWasm/cosmwasm/blob/main/packages/std/src/math.rs#L774-L780), which is enough to keep the contract secure. It won't provide a pretty error message to the user, but will block the attack vector. I would just use the normal `Uint128 + Uint128` approach

contracts/cw1155-base/src/contract.rs
```diff
@@ -0,0 +1,1283 @@
...
+                mock_info(users[0].clone(), &[]),
+                Cw1155HandleMsg::ApproveAll {
+                    operator: user.clone(),
+                    expires: None,
+                },
+            )
+            .unwrap();
+        }
+
+        // // FIXME test failed, query result out of order
```
ethanfrey, line 1102: They are not "out of order", it is that the keys are stored in `CanonicalAddr`, which you cannot assume is the same ordering as `HumanAddr`. For bech32 the order sometimes is the same, sometimes different. For tests, we semi-randomize the mapping to ensure you do not depend on it.

If you hardcode some addresses, just figure out via a few test runs what the ordering is as CanonicalAddr and then assert that one. It will be stable, but it makes you be sure if you are comparing canonical or human addresses.

Discussion here: https://github.com/CosmWasm/cosmwasm/issues/552

ethanfrey (commented):
I just got back from vacation and have lots to do before I can take a deeper look, but I did add comments on the 2 FIXMEs. These are subtle points about how `cosmwasm-std` is designed and actually intended.

contracts/cw1155-base/src/contract.rs
yihuang: I reverted back to simple `+` operator.
Opened issue in cosmwasm-std for the overflow handling: https://github.com/CosmWasm/cosmwasm/issues/850

contracts/cw1155-base/src/contract.rs
yihuang: thanks for clarification, FIXME is resolved now.

yihuang:
> I just got back from vacation and have lots to do before I can take a deeper look, but I did add comments on the 2 FIXMEs. These are subtle points about how `cosmwasm-std` is designed and actually intended.

Thanks, two FIXMEs are resolved now.

yihuang:
Add unit tests to expect panic when minting token amount overflow.

contracts/cw1155-base/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "cw1155-base"
+version = "0.6.0-alpha1"
```
ethanfrey, line 3: I think I just bumped this up to `-alpha3` trying to get multitest working for another app I'm building.

ethanfrey: See #257 and #258

contracts/cw1155-base/src/state.rs
```diff
@@ -0,0 +1,23 @@
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+use cosmwasm_std::{CanonicalAddr, Uint128};
+use cw1155::Expiration;
+use cw_storage_plus::{Item, Map};
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
```
ethanfrey, line 8: Where is this used? In query response?

contracts/cw1155-base/src/state.rs
```diff
@@ -0,0 +1,23 @@
...
+pub struct Approval {
+    /// Account that can transfer/send the token
+    pub spender: CanonicalAddr,
+    /// When the Approval expires (maybe Expiration::never)
+    pub expires: Expiration,
+}
+/// Store the minter address who have permission to mint new tokens.
```
ethanfrey, line 15: Nice job with the state, and good to comment each one what those &[u8] and &str mean :+1:

packages/cw1155/src/event.rs
```diff
@@ -1,42 +1,54 @@
-use cosmwasm_std::{attr, Attribute, HumanAddr, Uint128};
+use cosmwasm_std::{HumanAddr, Response, Uint128};
 use cw0::Event;
 
-use crate::msg::TokenId;
-
+/// Tracks token transfer/mint/burn actions
```
ethanfrey, line 4: Interesting that this is all references.
Definitely more efficient, but harder to get write. But nicely done.

contracts/cw1155-base/src/contract.rs
```diff
@@ -0,0 +1,1239 @@
...
+use crate::state::{APPROVES, BALANCES, MINTER, TOKENS};
+
+// version info for migration info
+const CONTRACT_NAME: &str = "crates.io:cw1155-base";
+const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+const DEFAULT_LIMIT: u32 = 10;
+const MAX_LIMIT: u32 = 30;
+
+pub fn instantiate(
```
ethanfrey, line 26: :100:

contracts/cw1155-base/src/contract.rs
```diff
@@ -0,0 +1,1239 @@
...
+    _info: MessageInfo,
+    msg: InitMsg,
+) -> StdResult<Response> {
+    set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+    let minter = deps.api.canonical_address(&msg.minter)?;
+    MINTER.save(deps.storage, &minter)?;
+    Ok(Response::default())
+}
+
+/// To mitigate clippy::too_many_arguments warning
```
ethanfrey, line 38: Interesting idea, I never tried that, but makes sense here

contracts/cw1155-base/src/contract.rs
```diff
@@ -0,0 +1,1239 @@
...
+fn check_can_approve(
+    deps: Deps,
+    env: &Env,
+    owner: &HumanAddr,
+    operator: &HumanAddr,
+) -> StdResult<bool> {
+    // owner can approve
+    let owner_raw = deps.api.canonical_address(owner)?;
+    let operator_raw = deps.api.canonical_address(operator)?;
+    if owner_raw == operator_raw {
```
ethanfrey, line 132: Important check.
Good logic here

contracts/cw1155-base/src/contract.rs
```diff
@@ -0,0 +1,1239 @@
...
+    to: Option<&'a HumanAddr>,
+    token_id: &'a str,
+    amount: Uint128,
+) -> Result<TransferEvent<'a>, ContractError> {
+    if let Some(from) = from {
+        let from_raw = deps.api.canonical_address(from)?;
+        BALANCES.update(
+            deps.storage,
+            (from_raw.as_slice(), token_id),
+            |balance: Option<Uint128>| balance.unwrap_or_default() - amount,
```
ethanfrey, line 101: Hmmm... do you want `checked_add` here? 
Or are you waiting for the next cosmwasm-std release?

contracts/cw1155-base/src/contract.rs
```diff
@@ -0,0 +1,1239 @@
...
+            execute_approve_all(env, operator, expires)
+        }
+        Cw1155ExecuteMsg::RevokeAll { operator } => execute_revoke_all(env, operator),
+    }
+}
+
+/// When from is None: mint new coins
+/// When to is None: burn coins
+/// When both are None: not token balance is changed, meaningless but valid
+fn execute_transfer_inner<'a>(
```
ethanfrey, line 89: Nice internal abstraction (after permission checks are done)

contracts/cw1155-base/src/contract.rs
```diff
@@ -0,0 +1,1239 @@
...
+    amount: Uint128,
+    msg: Option<Binary>,
+) -> Result<Response, ContractError> {
+    let ExecuteEnv {
+        mut deps,
+        env,
+        info,
+    } = env;
+
+    guard(
```
ethanfrey, line 165: You find this easier to write than

```rust
if !check_can_approve(deps.as_ref(), &env, &from, &info.sender)? {
  return Err(ContractError::Unauthorized {});
}
```

?? It's fine either way.

contracts/cw1155-base/src/contract.rs
```diff
@@ -0,0 +1,1239 @@
...
+        check_can_approve(deps.as_ref(), &env, &from, &info.sender)?,
+        ContractError::Unauthorized {},
+    )?;
+
+    let mut rsp = Response::default();
+
+    let event = execute_transfer_inner(&mut deps, Some(&from), Some(&to), &token_id, amount)?;
```
ethanfrey, line 172: This is very cool use of events

contracts/cw1155-base/src/contract.rs
```diff
@@ -0,0 +1,1239 @@
...
+    event.add_attributes(&mut rsp);
+
+    rsp.messages = if let Some(msg) = msg {
```
ethanfrey, line 175: How about just setting this inside the `if` and dropping the else?
`rsp.messages = vec![]` is already set with `default` above.

contracts/cw1155-base/src/contract.rs
```diff
@@ -0,0 +1,1239 @@
...
+    if sender != MINTER.load(deps.storage)? {
+        return Err(ContractError::Unauthorized {});
+    }
+
+    let mut rsp = Response::default();
+
+    let event = execute_transfer_inner(&mut deps, None, Some(&to), &token_id, amount)?;
+    event.add_attributes(&mut rsp);
+
+    rsp.messages = if let Some(msg) = msg {
```
ethanfrey, line 210: same as above

contracts/cw1155-base/src/contract.rs
```diff
@@ -0,0 +1,1239 @@
...
+        }
+        .into_cosmos_msg(to)?]
+    } else {
+        vec![]
+    };
+
+    // insert if not exist
+    let key = TOKENS.key(&token_id);
+    if deps.storage.get(&key).is_none() {
+        key.save(deps.storage, &"".to_owned())?;
```
ethanfrey, line 226: This means creating new token with no metadata?

Nice trick with `deps.storage.get(&key).is_none()` to do a "has" query and avoid parsing.
Should we support something like `TOKENS.has(&token_id)?` directly in cw-storage-plus?

contracts/cw1155-base/src/contract.rs
```diff
@@ -0,0 +1,1239 @@
...
+
+    guard(
+        check_can_approve(deps.as_ref(), &env, &from, &info.sender)?,
+        ContractError::Unauthorized {},
+    )?;
+
+    let mut rsp = Response::default();
+    for (token_id, amount) in batch.iter() {
+        let event = execute_transfer_inner(&mut deps, Some(&from), Some(&to), token_id, *amount)?;
+        event.add_attributes(&mut rsp);
```
ethanfrey, line 277: And here the events really shine. :1st_place_medal:

contracts/cw1155-base/src/contract.rs
```diff
@@ -0,0 +1,1239 @@
...
+    }
+
+    let mut rsp = Response::default();
+
+    for (token_id, amount) in batch.iter() {
+        let event = execute_transfer_inner(&mut deps, None, Some(&to), &token_id, *amount)?;
+        event.add_attributes(&mut rsp);
+    }
+
+    for (token_id, _) in batch.iter() {
```
ethanfrey, line 314: You don't want to roll this in the same above for loop?

contracts/cw1155-base/src/contract.rs
```diff
@@ -0,0 +1,1239 @@
...
+
+    let mut rsp = Response::default();
+    for (token_id, amount) in batch.into_iter() {
+        let event = execute_transfer_inner(&mut deps, Some(&from), None, &token_id, amount)?;
+        event.add_attributes(&mut rsp);
+    }
+    Ok(rsp)
+}
+
+pub fn execute_approve_all(
```
ethanfrey, line 361: Looks good (approve and revoke)

contracts/cw1155-base/src/contract.rs
```diff
@@ -0,0 +1,1239 @@
...
+        sender: &info.sender,
+        operator: &operator,
+        approved: false,
+    }
+    .add_attributes(&mut rsp);
+    Ok(rsp)
+}
+
+pub fn query(deps: Deps, env: Env, msg: Cw1155QueryMsg) -> StdResult<Binary> {
+    match msg {
```
ethanfrey, line 406: It does make sense to inline these short functions. I never thought of it.

contracts/cw1155-base/src/contract.rs
```diff
@@ -0,0 +1,1239 @@
...
+        Cw1155QueryMsg::IsApprovedForAll { owner, operator } => {
+            let approved = check_can_approve(deps, &env, &owner, &operator)?;
+            to_binary(&IsApprovedForAllResponse { approved })
+        }
+        Cw1155QueryMsg::ApprovedForAll {
+            owner,
+            include_expired,
+            start_after,
+            limit,
+        } => to_binary(&query_all_approvals(
```
ethanfrey, line 435: And clearly some deserve their own function

contracts/cw1155-base/src/contract.rs
```diff
@@ -0,0 +1,1239 @@
...
+    include_expired: bool,
+    start_after: Option<HumanAddr>,
+    limit: Option<u32>,
+) -> StdResult<ApprovedForAllResponse> {
+    let limit = limit.unwrap_or(DEFAULT_LIMIT).min(MAX_LIMIT) as usize;
+    let start_canon = maybe_canonical(deps.api, start_after)?;
+    let start = start_canon.map(Bound::exclusive);
+
+    let owner_raw = deps.api.canonical_address(&owner)?;
+    let operators = APPROVES
```
ethanfrey, line 478: Nice work with the ranges (here and below)

contracts/cw1155-base/src/contract.rs
```diff
@@ -0,0 +1,1239 @@
...
+                mock_info(user1.clone(), &[]),
+                Cw1155ExecuteMsg::BatchBurn {
+                    from: user1.clone(),
+                    batch: vec![(token2.clone(), 1u64.into()), (token3.clone(), 1u64.into())]
+                }
+            )
+            .unwrap(),
+            Response {
+                attributes: vec![
+                    attr("action", "transfer"),
```
ethanfrey, line 881: This is interesting. I am not sure how this ends up when it comes out through tendermint (I think it does some "collapsing" of events).

It totally makes sense. I wonder if you had a wish on a cleaner way to encode these or pass them through to the client? Although I guess just as it is is perfect for searching.

contracts/cw1155-base/src/state.rs
yihuang: it's initially copied from cw721 and never used, removed now.

contracts/cw1155-base/Cargo.toml
yihuang: fixed now.

contracts/cw1155-base/src/contract.rs
yihuang: I think we can change them at once in the cosmwasm-std upgrade.

contracts/cw1155-base/src/contract.rs
yihuang: agree, it's pointless, removed the `guard`, ;D

contracts/cw1155-base/src/contract.rs
yihuang: done

contracts/cw1155-base/src/contract.rs
yihuang: done

contracts/cw1155-base/src/contract.rs
yihuang: > This means creating new token with no metadata?

yes, insert an empty entry so token enumeration can find it.

> 
> Nice trick with `deps.storage.get(&key).is_none()` to do a "has" query and avoid parsing.
> Should we support something like `TOKENS.has(&token_id)?` directly in cw-storage-plus?

I think that'll be useful, another option is `key.exists(deps.storage)`.

contracts/cw1155-base/src/contract.rs
yihuang: > You don't want to roll this in the same above for loop?

thanks, rolled in now, looks better.

contracts/cw1155-base/src/contract.rs
yihuang: I'm not sure how it's going to be indexed by tendermint though, need to look into details in tendermint I guess.

contracts/cw1155-base/src/contract.rs
yihuang: Further encapsulate a `guard_can_approve` which return `ContractError` directly.

ethanfrey: > Further encapsulate a guard_can_approve which return ContractError directly.

Yes, something like `guard_can_approve(deps.as_ref(), &env, &from, &info.sender)?;` or `assert_can_approve(deps.as_ref(), &env, &from, &info.sender)?;` seems like a nice abstraction, just not the two separately which doesn't save typing

contracts/cw1155-base/src/contract.rs
```diff
@@ -207,18 +201,16 @@ pub fn execute_mint(
     let event = execute_transfer_inner(&mut deps, None, Some(&to), &token_id, amount)?;
     event.add_attributes(&mut rsp);
 
-    rsp.messages = if let Some(msg) = msg {
-        vec![Cw1155ReceiveMsg {
+    if let Some(msg) = msg {
+        rsp.messages = vec![Cw1155ReceiveMsg {
```
ethanfrey, line 216: This could also be `rsp.add_message(Cw1155ReceiveMsg { .. }.into_cosmos_msg(to)?);`

But I don't think that is any clearer in this case.

contracts/cw1155-base/src/contract.rs
```diff
@@ -309,26 +297,23 @@ pub fn execute_batch_mint(
     for (token_id, amount) in batch.iter() {
         let event = execute_transfer_inner(&mut deps, None, Some(&to), &token_id, *amount)?;
         event.add_attributes(&mut rsp);
-    }
 
-    for (token_id, _) in batch.iter() {
         // insert if not exist
         let key = TOKENS.key(&token_id);
         if deps.storage.get(&key).is_none() {
+            // insert an empty entry so token enumeration can find it
```
ethanfrey, line 311: thanks for the comment

contracts/cw1155-base/src/contract.rs
```diff
@@ -42,6 +44,7 @@ pub struct ExecuteEnv<'a> {
     info: MessageInfo,
 }
 
+#[cfg_attr(not(feature = "library"), entry_point)]
```
ethanfrey, line 47: Nice update

contracts/cw1155-base/src/contract.rs
```diff
@@ -159,9 +172,7 @@ pub fn execute_send_from(
         info,
     } = env;
 
-    if !check_can_approve(deps.as_ref(), &env, &from, &info.sender)? {
-        return Err(ContractError::Unauthorized {});
-    }
+    guard_can_approve(deps.as_ref(), &env, &from, &info.sender)?;
```
ethanfrey, line 175: Yeah, this is a nicer way to inline the common check.

contracts/cw1155-base/src/contract.rs
ethanfrey: Yeah, they should all match

ethanfrey (approved):
Good stuff.

ethanfrey:
Created https://github.com/CosmWasm/cosmwasm-plus/issues/261 inspired by your comments

PR #253, opened 2021-03-26
PartialEq for errors

### LINKED ISSUE #179: Support PartialEq on error
Note: this is a long background process. Let's update contracts one by one as we refactor them (a bit like updating them all to storage-plus)

With `StdError::PartialEq` implementation from https://github.com/CosmWasm/cosmwasm/pull/662, we can replace lots of statements like:

```rust
        match err {
            ContractError::Unauthorized {} => {}
            e => panic!("Unexpected error: {}", e),
        }
```

with simpler statements like:

```rust
assert_eq!(ContractError::Unauthorized {}, err);
```

It requires `#[derive(PartialEq)]` on `ContractError` and the above-mentioned support for `StdError`

Contracts to update:

- [x] cw1-subkeys
- [x] cw1-whitelist
- [x] cw3-fixed-multisig
- [x] cw3-flex-multisig
- [x] cw4-group
- [x] cw4-stake
- [x] cw20-atomic-swap
- [x] cw20-base
- [x] cw20-bonding
- [x] cw20-escrow
- [x] cw20-staking
- [x] cw721-base

maurolacy:
Closes #179.

### DIFF
--- a/contracts/cw1-subkeys/src/contract.rs
+++ b/contracts/cw1-subkeys/src/contract.rs
@@ -1184,10 +1184,10 @@ mod tests {
 
         // And then cannot (not enough funds anymore)
         let res = execute(deps.as_mut(), mock_env(), info, execute_msg.clone());
-        match res.unwrap_err() {
-            ContractError::Std(StdError::Underflow { .. }) => {}
-            e => panic!("unexpected error: {:?}", e),
-        }
+        assert!(matches!(
+            res.unwrap_err(),
+            ContractError::Std(StdError::Underflow { .. })
+        ));
 
         // Owner / admins can do anything (at the contract level)
         let info = mock_info(&owner.clone(), &[]);
--- a/contracts/cw1-subkeys/src/contract.rs
+++ b/contracts/cw1-subkeys/src/contract.rs
@@ -1170,8 +1170,8 @@ mod tests {
 
         // spender2 cannot spend funds (no initial allowance)
         let info = mock_info(&spender2, &[]);
-        let res = execute(deps.as_mut(), mock_env(), info, execute_msg.clone());
-        assert_eq!(ContractError::NoAllowance {}, res.unwrap_err());
+        let err = execute(deps.as_mut(), mock_env(), info, execute_msg.clone()).unwrap_err();
+        assert_eq!(err, ContractError::NoAllowance {});
 
         // But spender1 can (he has enough funds)
         let info = mock_info(&spender1, &[]);
@@ -1183,9 +1183,9 @@ mod tests {
         );
 
         // And then cannot (not enough funds anymore)
-        let res = execute(deps.as_mut(), mock_env(), info, execute_msg.clone());
+        let err = execute(deps.as_mut(), mock_env(), info, execute_msg.clone()).unwrap_err();
         assert!(matches!(
-            res.unwrap_err(),
+            err,
             ContractError::Std(StdError::Underflow { .. })
         ));
 
@@ -1214,8 +1214,8 @@ mod tests {
 
         // But not for mere mortals
         let info = mock_info(&spender1, &[]);
-        let res = execute(deps.as_mut(), mock_env(), info, execute_msg.clone());
-        assert_eq!(ContractError::MessageTypeRejected {}, res.unwrap_err());
+        let err = execute(deps.as_mut(), mock_env(), info, execute_msg.clone()).unwrap_err();
+        assert_eq!(err, ContractError::MessageTypeRejected {});
     }
 
     #[test]
--- a/contracts/cw1-subkeys/src/contract.rs
+++ b/contracts/cw1-subkeys/src/contract.rs
@@ -1171,10 +1171,7 @@ mod tests {
         // spender2 cannot spend funds (no initial allowance)
         let info = mock_info(&spender2, &[]);
         let res = execute(deps.as_mut(), mock_env(), info, execute_msg.clone());
-        match res.unwrap_err() {
-            ContractError::NoAllowance { .. } => {}
-            e => panic!("unexpected error: {}", e),
-        }
+        assert_eq!(ContractError::NoAllowance {}, res.unwrap_err());
 
         // But spender1 can (he has enough funds)
         let info = mock_info(&spender1, &[]);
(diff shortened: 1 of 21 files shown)


contracts/cw1-subkeys/src/contract.rs
```diff
@@ -1166,10 +1166,7 @@ mod tests {
         // spender2 cannot spend funds (no initial allowance)
         let info = mock_info(&spender2, &[]);
         let res = execute(deps.as_mut(), mock_env(), info, execute_msg.clone());
-        match res.unwrap_err() {
-            ContractError::NoAllowance { .. } => {}
-            e => panic!("unexpected error: {}", e),
-        }
+        assert_eq!(ContractError::NoAllowance {}, res.unwrap_err());
```
ethanfrey, line 1169: Nice, this is cleaner.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -1184,7 +1181,7 @@ mod tests {
         let res = execute(deps.as_mut(), mock_env(), info, execute_msg.clone());
         match res.unwrap_err() {
             ContractError::Std(StdError::Underflow { .. }) => {}
-            e => panic!("unexpected error: {}", e),
+            e => panic!("unexpected error: {:?}", e),
```
ethanfrey, line 1184: Ah, I remember, the StdError with backtraces cannot be compared with the `assert_eq!` macro. I couldn't figure a way to make this easier when migrating other contracts, but it would be nice to have this shorter as well.

I wonder if you could use [`matches!`](https://doc.rust-lang.org/std/macro.matches.html) to make this easier?

`assert!(matches!(res.unwrap_err(), ContractError::Std(StdError::Underflow { .. }))` or something like that?

contracts/cw20-base/src/contract.rs
```diff
@@ -497,10 +497,10 @@ mod tests {
         let info = mock_info(&HumanAddr("creator".to_string()), &[]);
         let env = mock_env();
         let res = instantiate(deps.as_mut(), env.clone(), info.clone(), instantiate_msg);
-        match res.unwrap_err() {
-            StdError::GenericErr { msg, .. } => assert_eq!(&msg, "Initial supply greater than cap"),
-            e => panic!("Unexpected error: {}", e),
-        }
+        assert_eq!(
+            StdError::generic_err("Initial supply greater than cap"),
```
ethanfrey, line 501: Does this work with the backtraces enabled (or do we ever even enable them here)? It seems you left the match for StdError above. I'm fine with whatever works, let's just keep it unified.

contracts/cw20-base/src/contract.rs
```diff
@@ -839,7 +818,7 @@ mod tests {
         let res = execute(deps.as_mut(), env, info, msg);
         match res.unwrap_err() {
             ContractError::Std(StdError::Underflow { .. }) => {}
-            e => panic!("Unexpected error: {}", e),
+            e => panic!("Unexpected error: {:?}", e),
```
ethanfrey, line 821: Again, one pattern for StdError checks would be best

ethanfrey (commented):
Looks good and improves a lot of tests.

However, I had some questions on what the canonical format for checking StdError types is and let us just define one pattern and stick to it (I see 2 approached above).

Other that that, just a rebase to fix merge conflicts and this is good to merge.

contracts/cw20-base/src/contract.rs
maurolacy: It does.

`matches!` cannot be used because of the "Initial supply greater than cap" msg, as it's not possible to create a `String` inside a pattern.
To be clear, something like
```
assert!(matches!(res.unwrap_err(), StdError::GenericErr { .. }));
```
works, whereas something like
```
assert!(matches!(res.unwrap_err(), StdError::GenericErr { msg: "Initial supply greater than cap", .. }));
```
does not compile, because the `msg` string is not a `String` but a `&str`, and converting to `String` is not allowed inside patterns.

So, we either use the `matches!` syntax here, and lose matching precision, or we stick with the current version, every time we want to match an explicit message / error content.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -1179,10 +1179,10 @@ mod tests {
 
         // And then cannot (not enough funds anymore)
         let res = execute(deps.as_mut(), mock_env(), info, execute_msg.clone());
-        match res.unwrap_err() {
-            ContractError::Std(StdError::Underflow { .. }) => {}
-            e => panic!("unexpected error: {:?}", e),
-        }
+        assert!(matches!(
```
ethanfrey, line 1182: This is definitely a step forward.

I wonder if this can somehow be compressed further. Could you experiment please? Maybe `let err = ....unwrap_err()` for the line above (chained with `execute`). Or maybe there is a shorthand for        `ContractError::Std(StdError::Underflow { .. })` (which I doubt).

It is fine as is, just wondering if there is a way to simplify further before we leave this as a pattern for all to copy.

contracts/cw20-atomic-swap/src/contract.rs
```diff
@@ -480,11 +480,10 @@ mod tests {
...
         };
         let res = execute(deps.as_mut(), mock_env(), info.clone(), release);
-        match res {
-            Ok(_) => panic!("expected error"),
-            Err(ContractError::InvalidPreimage {}) => {}
-            Err(e) => panic!("unexpected error: {:?}", e),
-        }
+        assert!(matches!(
+            res.unwrap_err(),
+            ContractError::InvalidPreimage {}
```
ethanfrey, line 485: Hmmm... even the shorter error messages break on multiple lines... maybe try the `.unwrap_err()` on the line above (482)

contracts/cw4-stake/src/contract.rs
```diff
@@ -557,17 +557,11 @@ mod tests {
...
-            }) => {
-                assert_eq!(minuend.as_str(), "5000");
-                assert_eq!(subtrahend.as_str(), "5100");
-            }
-            e => panic!("Unexpected error: {:?}", e),
-        }
+        let res = execute(deps.as_mut(), env, info, msg);
+        assert_eq!(
+            res.unwrap_err(),
+            ContractError::Std(StdError::underflow(5000, 5100))
```
ethanfrey, line 563: Nice simplification.

ethanfrey (approved):
Looks good. Just one idea for further refinement (optional)

contracts/cw4-stake/src/contract.rs
maurolacy: Yes. I think this is the way to go, when you want to match / check error values or messages, not only error types.

contracts/cw1-subkeys/src/contract.rs
maurolacy: We can put the `unwrap_err()` in the line above, sure. I'll do it.

maurolacy:
Merging now, so as to integrate with the other branches.

PR #254, opened 2021-03-30
Update to Rust v1.51.0

### LINKED ISSUE #252: Re-enable field_reassign_with_default
Similar to cosmwasm [issue 685](https://github.com/CosmWasm/cosmwasm/issues/685).

This depends on a new cosmwasm-schema version, i.e. one that uses schemars 0.8.

maurolacy:
Updates to shiny new Rust v1.51.0.

Closes #252 in passing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,7 +40,7 @@ workflows:
 jobs:
   contract_cw1_subkeys:
     docker:
-      - image: rust:1.50.0
+      - image: rust:1.51.0
     working_directory: ~/project/contracts/cw1-subkeys
     steps:
       - checkout:
@@ -50,7 +50,7 @@ jobs:
           command: rustc --version; cargo --version; rustup --version
       - restore_cache:
           keys:
-            - cargocache-cw1-subkeys-rust:1.50.0-{{ checksum "~/project/Cargo.lock" }}
+            - cargocache-cw1-subkeys-rust:1.51.0-{{ checksum "~/project/Cargo.lock" }}
       - run:
           name: Unit Tests
           environment:
@@ -72,11 +72,11 @@ jobs:
           paths:
             - /usr/local/cargo/registry
             - target
-          key: cargocache-cw1-subkeys-rust:1.50.0-{{ checksum "~/project/Cargo.lock" }}
+          key: cargocache-cw1-subkeys-rust:1.51.0-{{ checksum "~/project/Cargo.lock" }}
 
   contract_cw1_whitelist:
     docker:
-      - image: rust:1.50.0
+      - image: rust:1.51.0
     working_directory: ~/project/contracts/cw1-whitelist
     steps:
       - checkout:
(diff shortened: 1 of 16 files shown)


contracts/cw1-subkeys/src/contract.rs
```diff
@@ -87,9 +87,11 @@ where
     let owner_raw = &deps.api.canonical_address(&info.sender)?;
     // this is the admin behavior (same as cw1-whitelist)
     if cfg.is_admin(owner_raw) {
-        let mut res = Response::default();
-        res.messages = msgs;
-        res.attributes = vec![attr("action", "execute"), attr("owner", info.sender)];
+        let res = Response {
+            messages: msgs,
+            attributes: vec![attr("action", "execute"), attr("owner", info.sender)],
+            ..Response::default()
```
ethanfrey, line 93: Do we want this pattern (okay with me) or explicitly asigning all field?

contracts/cw1-whitelist/src/contract.rs
```diff
@@ -74,9 +74,11 @@ where
     if !can_execute(deps.as_ref(), &info.sender)? {
         Err(ContractError::Unauthorized {})
     } else {
-        let mut res = Response::default();
-        res.messages = msgs;
-        res.attributes = vec![attr("action", "execute")];
+        let res = Response {
+            messages: msgs,
+            attributes: vec![attr("action", "execute")],
+            ..Default::default()
```
ethanfrey, line 80: I know they are the same, but I prefer `Response::default` to `Default::default` when we have access to the explicit types (not just traits)

.circleci/config.yml
```diff
@@ -834,12 +833,12 @@ jobs:
           # Uses --debug for compilation speed
           # FIXME: Change when `check_contract` (part of `cosmwasm-0.14.0`) is published
           #command: cargo install --debug --features iterator --example check_contract -- cosmwasm-vm
-          command: cargo install --debug --features iterator --git https://github.com/CosmWasm/cosmwasm --tag=v0.14.0-beta1 --example check_contract -- cosmwasm-vm
+          command: cargo install --debug --features iterator --git https://github.com/CosmWasm/cosmwasm --branch=main --example check_contract -- cosmwasm-vm
```
ethanfrey, line 836: I wouldn't make this change here. Rather keep it on a tag (can bump to -beta2 when ready)

.circleci/config.yml
```diff
@@ -792,23 +792,22 @@ jobs:
           command: cargo fmt -- --check
       - run:
           name: Clippy linting on workspace
-          # Silence false positive `field_reassing_with_default` (https://github.com/rust-lang/rust-clippy/issues/6545)
-          command: cargo clippy -- -D warnings -A clippy::field_reassign_with_default
```
ethanfrey, line 796: :+1:

ethanfrey (approved):
Looks good

contracts/cw1-subkeys/src/contract.rs
maurolacy: This is the suggested way. I think it's better / more general than the explicit assignment.

.circleci/config.yml
maurolacy: This should in principle be changed to a released version (i.e. the commented line above). I'm OK with disabling the main version, though.

PR #257, opened 2021-04-01
Enhance multi test

ethanfrey:
Make this compatible with more advanced workflows

- [x] Support returning Response<C> with custom message (this was a TODO)
- [x] Add support for contract wrappers with `sudo` entrypoint

TODO (other PR?)
- [ ] Expose `sudo` on top level
- [ ] Add support for `reply` entry point
- [ ] Process `submessages` in app and pass response into `reply`

### DIFF
--- a/packages/multi-test/src/test_helpers.rs
+++ b/packages/multi-test/src/test_helpers.rs
@@ -33,7 +33,8 @@ fn query_error(_deps: Deps, _env: Env, _msg: EmptyMsg) -> Result<Binary, StdErro
 }
 
 pub fn contract_error() -> Box<dyn Contract> {
-    let contract = ContractWrapper::new(handle_error, init_error, query_error);
+    let contract: ContractWrapper<_, _, _, _, _, _, _, String, String> =
+        ContractWrapper::new(handle_error, init_error, query_error);
     Box::new(contract)
 }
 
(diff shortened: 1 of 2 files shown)


ethanfrey:
I'm going to merge and tag this to unblock some work, and add the other items in a future PR.

ethanfrey:
Okay, I made a `0.6.0-alpha2` release with this and got one step further. Now the issue is:

```
 --> contracts/tgrade-valset/src/contract.rs:423:40
    |
423 |         let valset_id = app.store_code(contract_valset());
    |                                        ^^^^^^^^^^^^^^^^^ expected struct `cosmwasm_std::Empty`, found enum `tgrade_bindings::TgradeMsg`
    |
    = note: expected struct `Box<(dyn Contract + 'static)>`
               found struct `Box<(dyn Contract<tgrade_bindings::TgradeMsg> + 'static)>`
```

So, I guess and app must accept only contracts that are `Contract<C>` with the same `C`. We assume there is only one custom message type on the chain, so in general this is good, but we must support `Contract<Empty>` as well.

So, basically, we will need to store `&dyn Contract<C>` in the app, but the contract wrapper should be able to accept functions that return `Response<Empty>` and convert it to `Response<C>` to handle this.

PR #260, opened 2021-04-02
Update cosmwasm-std

yihuang:
Maybe merge after the cosmwasm-std version released.


ethanfrey:
Thanks for this initiative. There are a few more changes, and let's wait til a 0.14.0-beta2 is out to finalize this.

ethanfrey:
https://github.com/CosmWasm/cosmwasm/releases/tag/v0.14.0-beta2 is now out (inspired by your work to cut this now).

No pressure, but feel free to upgrade to that and I'd be happy to merge.
Only remaining (major) plans for 0.14.0 is some refactoring of the IBC bindings, but everything else should be pretty stable now.

yihuang:
> https://github.com/CosmWasm/cosmwasm/releases/tag/v0.14.0-beta2 is now out (inspired by your work to cut this now).
> 
> No pressure, but feel free to upgrade to that and I'd be happy to merge.
> Only remaining (major) plans for 0.14.0 is some refactoring of the IBC bindings, but everything else should be pretty stable now.

There are many changes needed in `multi-test` for recent address change, I'm not sure when to use `Addr::unchecked` or `addr_validate`.

maurolacy:
> There are many changes needed in `multi-test` for recent address change, I'm not sure when to use `Addr::unchecked` or `addr_validate`.

Hi, this comment https://github.com/CosmWasm/cosmwasm/blob/main/packages/std/src/addresses.rs#L30-L44 may be of help.

yihuang:
> > There are many changes needed in `multi-test` for recent address change, I'm not sure when to use `Addr::unchecked` or `addr_validate`.
> 
> This comment https://github.com/CosmWasm/cosmwasm/blob/main/packages/std/src/addresses.rs#L30:L44 may be of help.

Since `multi-test` is a test library, i suppose it's free to use `Addr:unchecked` as needed?

maurolacy:
> > > There are many changes needed in `multi-test` for recent address change, I'm not sure when to use `Addr::unchecked` or `addr_validate`.
> > 
> > 
> > This comment https://github.com/CosmWasm/cosmwasm/blob/main/packages/std/src/addresses.rs#L30:L44 may be of help.
> 
> Since `multi-test` is a test library, i suppose it's free to use `Addr:unchecked` as needed?

I guess it depends on the particular case. Basically, use `Addr::unchecked` to generate an `Addr` from a string in test code, or in contract code, if it has already been validated. For the other cases, you should use `addr_validate()` to generate and `Addr` from a string. Or `addr_humanize()`, to generate an `Addr` from a (binary repr) `CanonicalAddr`.

If you post a specific example / instance, we can discuss it.

maurolacy:
On second thought, I think you're right. If we want the multi-test framework to work generically over addresses, `uncheched` would be the right choice.

ethanfrey:
I just merged in your cw1155 contract. Many thanks for that.

If you want to rebase and push with latest master and v0.14.0-beta2, I will look at the compile failures in multi-test and maybe can fix them (or at least find a good solution). `multi-test` is kind of my baby and in beta.

ethanfrey:
My short feedback is in non-test code `s/HumanAddr/Addr/` and if we really need to create them, we look at it (and avoid `unchecked` if possible).

In test code, `HumanAddr::from("owner")` -> `Addr::unchecked("owner")`

yihuang:
@ethanfrey @maurolacy I pushed lots of address related changes, it's not complete yet, I guess you can check it out to see if I'm on the right track 😄

ethanfrey:
Thank you, I will look

packages/multi-test/src/app.rs
```diff
@@ -354,17 +350,18 @@ where
...
         msg: WasmMsg,
-    ) -> Result<(HumanAddr, ActionResponse<C>), String> {
+    ) -> Result<(Addr, ActionResponse<C>), String> {
         match msg {
             WasmMsg::Execute {
                 contract_addr,
                 msg,
                 send,
             } => {
+                let contract_addr = Addr::unchecked(contract_addr);
```
ethanfrey, line 362: I would validate here instead of unchecked.

`let contract_addr = self.api.addr_validate(&contract_addr)?;` should work here, well, except for the String error.

Maybe add `App.add_validate(addr: &str) -> Result<Addr, String>` which is a small wrapper around the above, then we can use that where needed in production code.

packages/multi-test/src/app.rs
```diff
@@ -381,9 +378,9 @@ where
                 send,
                 label: _,
             } => {
-                let contract_addr = self.wasm.register_contract(code_id as usize)?;
+                let contract_addr = Addr::unchecked(self.wasm.register_contract(code_id as usize)?);
```
ethanfrey, line 381: register_contract should return Addr.

ethanfrey (commented):
Took a brief look at multi-test.

I added some comments, but I am not sure if I am just making a huge pile of work here. Let's go with your approach using `Addr::unchecked` as needed here and I can always do another PR later to add validation. Multi-test is only test code, so this is more pedantic.

I will look at the contracts to see use in production code.

packages/cw0/src/pagination.rs
```diff
@@ -1,21 +1,18 @@
...
-    human.map(|x| api.canonical_address(&x)).transpose()
+pub fn maybe_canonical(api: &dyn Api, human: Option<Addr>) -> StdResult<Option<CanonicalAddr>> {
+    human.map(|x| api.addr_canonicalize(x.as_ref())).transpose()
 }
 
 // this will set the first key after the provided key, by appending a 0 byte
 pub fn calc_range_start_human(
     api: &dyn Api,
-    start_after: Option<HumanAddr>,
+    start_after: Option<Addr>,
```
ethanfrey, line 11: I don't think this is used anywhere anymore. It can be removed.

packages/cw1/src/helpers.rs
```diff
@@ -1,32 +1,32 @@
...
 use crate::msg::Cw1ExecuteMsg;
 
-/// Cw1Contract is a wrapper around HumanAddr that provides a lot of helpers
+/// Cw1Contract is a wrapper around Addr that provides a lot of helpers
 /// for working with this.
 ///
 /// If you wish to persist this, convert to Cw1CanonicalContract via .canonical()
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
-pub struct Cw1Contract(pub HumanAddr);
+pub struct Cw1Contract(pub Addr);
```
ethanfrey, line 13: :+1:

packages/cw1/src/helpers.rs
```diff
@@ -42,7 +42,7 @@ pub struct Cw1CanonicalContract(pub CanonicalAddr);
 impl Cw1CanonicalContract {
```
ethanfrey, line 42: You can remove this entirely, as I we are moving away from CanonicalContracts.

Fine if you leave this for now to make the PR smaller, and I can take a second PR for that one.

packages/cw2/src/lib.rs
```diff
@@ -44,7 +42,7 @@ pub fn set_contract_version<T: Into<String>, U: Into<String>>(
 /// if the other contract exists and claims to be a cw20-base contract for example.
 /// (Note: you usually want to require *interfaces* not *implementations* of the
 /// contracts you compose with, so be careful of overuse)
-pub fn query_contract_info<Q: Querier, T: Into<HumanAddr>>(
+pub fn query_contract_info<Q: Querier, T: Into<String>>(
```
ethanfrey, line 45: `Into<String>` :+1:

packages/cw3/src/query.rs
```diff
@@ -31,20 +31,20 @@ pub enum Cw3QueryMsg {
     /// Query the vote made by the given voter on `proposal_id`. This should
     /// return an error if there is no such proposal. It will return a None value
     /// if the proposal exists but the voter did not vote. Returns VoteResponse
-    Vote { proposal_id: u64, voter: HumanAddr },
+    Vote { proposal_id: u64, voter: Addr },
```
ethanfrey, line 34: All `*Msg` types should use `String` (this is unverified input from users)
All state should use `Addr`.

By doing so, the compiler should enforce we validate input before saving it

packages/cw3/src/query.rs
```diff
@@ -168,7 +168,7 @@ pub struct VoteListResponse {
 /// the address of the voter who submitted it
 #[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
 pub struct VoteInfo {
-    pub voter: HumanAddr,
+    pub voter: Addr,
```
ethanfrey, line 171: Again `String` here

packages/cw4/src/hook.rs
```diff
@@ -10,17 +10,13 @@ use cosmwasm_std::{to_binary, Binary, CosmosMsg, HumanAddr, StdResult, WasmMsg};
...
     pub new: Option<u64>,
 }
 
 impl MemberDiff {
-    pub fn new<T: Into<HumanAddr>>(
-        addr: T,
-        old_weight: Option<u64>,
-        new_weight: Option<u64>,
-    ) -> Self {
+    pub fn new<T: Into<Addr>>(addr: T, old_weight: Option<u64>, new_weight: Option<u64>) -> Self {
```
ethanfrey, line 19: Is there anything that really implements `Into<Addr>`? I think we can just use `Addr` here (or does this allow `&addr` as well as `addr` as args?)

packages/cw4/src/hook.rs
```diff
@@ -53,10 +49,10 @@ impl MemberChangedHookMsg {
     }
 
     /// creates a cosmos_msg sending this struct to the named contract
-    pub fn into_cosmos_msg(self, contract_addr: HumanAddr) -> StdResult<CosmosMsg> {
+    pub fn into_cosmos_msg(self, contract_addr: Addr) -> StdResult<CosmosMsg> {
```
ethanfrey, line 52: Maybe `Into<String>` here? So we can use `Addr`, `String`, or `&str`?

packages/cw4/src/hook.rs
```diff
@@ -10,17 +10,13 @@ use cosmwasm_std::{to_binary, Binary, CosmosMsg, HumanAddr, StdResult, WasmMsg};
 /// old = Some, new = None -> Delete
 #[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
 pub struct MemberDiff {
-    pub key: HumanAddr,
+    pub key: Addr,
```
ethanfrey, line 13: Wait, this is a message, not state.
Again, use String.
`Addr` only in state and env/info from the runtime (already trusted)

ethanfrey (commented):
Added some feedback.

This is hard to get right, and I only understood the proper usage after discussing with Simon on the PR. We need to update the migration / readme - any suggestions are welcome.

Basically, anything that can be set by external users should be `String`. This includes anything entering and leaving the contract (query response is parsed by someone else and will not trust the contract by default).

Anything we save to state should be `Addr`, to ensure it is valid.

We can remove almost all usage of CanonicalAddr. You can keep them for keys (like `maybe_canonical`), but I may replace that with `Addr` in the keys in the future.

ethanfrey:
This may be easier to collaborate if you could make a branch on this repo, so I would push some code and collaborate on this PR (this is a huge cleanup task we don't need to dump entirely on you, but thank you very much for taking it on). 
Do you want write access to this repo?

ethanfrey:
Okay, I will try that and do a quick Addr -> String in the messages.

ethanfrey:
Okay, I just pushed a commit for cw1-4 to show what I mean. I am off for lunch for an hour or so.

If this makes sense, feel free to continue. If not, I can pick this up later today

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -394,9 +394,11 @@ fn map_proposal(
     })
 }
 
-fn query_vote(deps: Deps, proposal_id: u64, voter: HumanAddr) -> StdResult<VoteResponse> {
-    let voter_raw = deps.api.canonical_address(&voter)?;
-    let ballot = BALLOTS.may_load(deps.storage, (proposal_id.into(), &voter_raw))?;
+fn query_vote(deps: Deps, proposal_id: u64, voter: String) -> StdResult<VoteResponse> {
+    let ballot = BALLOTS.may_load(
+        deps.storage,
+        (proposal_id.into(), AddrRef::unchecked(&voter)),
```
webmaster128, line 400: This is an interesting diff. Before the user input was checked and only if it was valid, the query to the data store was made. I think we should continue to do that. The other option is to use String/str in the map.

But fake-casting to `Addr` seems to defeat the whole purpose if the strict type.

contracts/cw4-group/src/contract.rs
```diff
@@ -309,7 +324,7 @@ mod tests {
         let err = update_members(
             deps.as_mut(),
             height + 5,
-            USER1.into(),
+            Addr::unchecked(USER1),
```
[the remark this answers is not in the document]
webmaster128: Which would by the way also allow check `Copy`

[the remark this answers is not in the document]
ethanfrey: Did you see what I was doing with `AddrRef`?
It is `&str` to `Addr`'s `String`. This could be done with `const fn`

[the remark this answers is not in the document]
ethanfrey: We just need better back and forth between those types.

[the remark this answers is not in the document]
webmaster128: `const fn` are executed at compile time and cannot do anything heap related. This is why you cannot have a Sting or Vec in a `const VARIABLE` . If we want to go that way, the address must be stored in a strict like

```rust
#[derive(Copy, ...)]
struct Addr {
    data: [u8; 128];
    length: usize;
}
```

However, this only gets handy if we can make a `const fn unchecked(input: &str) -> Addr` and I don't know if it is possible to copy the `&str` data into the array at compile time via other const functions.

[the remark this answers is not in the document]
maurolacy: > We just need better back and forth between those types.

My thoughts exactly. `Into` calling `unchecked()` is a no no, as it would defeat `addr_validate`. Maybe a simple helper like `addr_unchecked()`, that lives in `cosmwasm_std` is the best we can ask for.

Another alternative is of course having a wrapper / marker type like `StringAddr` that implements `From<Addr>` using `unchecked`, but that is what we were trying to avoid when removing `HumanAddr`.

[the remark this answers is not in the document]
webmaster128: I see all this concerns and we can tacle them one-by-one. Looking at the particular line I commented: Fewer conversions would be necessary if USER1 was an `Addr` instead of a `&str`. Converting `&str -> Addr` is hard because unsafe. But `Addr -> &str` is safe and easy. 

https://github.com/CosmWasm/cosmwasm/pull/869 shows an implementation that makes this possible.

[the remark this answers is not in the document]
maurolacy: Nice. Will review tomorrow.

maurolacy:
@yihuang , can you take it from here and implement the needed changes (add schema) for `cw1155-base` to pass CI?

ethanfrey:
> Besides, I think this will require at least a quick review before merging, to try and spot any errors / omissions.

I will do a review on everything but cw1155 for now, and that when @yihuang flags it as ready

yihuang:
> @yihuang , can you take it from here and implement the needed changes (add schema) for `cw1155-base` to pass CI?

it seems already done by @ethanfrey ?

maurolacy:
> > @yihuang , can you take it from here and implement the needed changes (add schema) for `cw1155-base` to pass CI?
> 
> it seems has done by @ethanfrey ? +1

Yes. We wanted to see the CI solid green. :-)

ethanfrey:
I'm pulling out all `AddrRef` to `&Addr` before merging. Otherwise, looking good

webmaster128:
🎉

ethanfrey:
Bumping to beta3 in a new PR now. #269

PR #262, opened 2021-04-08
Use workspace-optimizer:0.11.0

webmaster128:
workspace-optimizer:0.11.0 is the best workspace-optimizer we ever built


ethanfrey (approved):
Optimizer better and better

Issue #198, opened 2020-12-21
Better return values from range/prefix


ethanfrey:
Builds on #197 

When iterating over a `range` from a Prefix we get from `Map`, we get back `(Vec<u8> , T)`. However, we know what the Suffix is.Eg for `K = (&[u8], U32Key)` we could get back `(U32Key, T)`

ethanfrey:
When we do implement this, we should:

1. Add a new trait (`ParsableKey` ?) with this method, it would not be required for all keys
2. Add a new method to Map, etc like `range_parsed` that returns `(K, T)` not `(Vec<u8>, T)` but only when K implements `ParseableKey`.
3. Implement something like https://github.com/CosmWasm/cosmwasm-plus/pull/265 so we can return an Owned structure from parsing, but pass in a ref to query.

ethanfrey:
Question: what happens when the primary key is references, eg. `&Addr`, or `&str`? It is relatively clear how to do so when the key is owned like eg. `String` or `U32Key`. Let's figure out how to define this better

PR #264, opened 2021-04-12
Implement PrimaryKey for &Addr by removing unused parse_key

webmaster128:
The trait `PrimaryKey` puts a lot of requirements on the type implementing it. It requires zero-copy serialization and zero-copy deserialization.

However, deserializing the bytes into the key type is never used. If we remove it, we can easily implemement PrimaryKey for &Addr.

------

Do not merge as is. Target branch is a copy of the fork from https://github.com/CosmWasm/cosmwasm-plus/pull/260.

### DIFF
--- a/contracts/cw3-fixed-multisig/src/contract.rs
+++ b/contracts/cw3-fixed-multisig/src/contract.rs
@@ -13,7 +13,7 @@ use cw3::{
     ProposalListResponse, ProposalResponse, Status, ThresholdResponse, Vote, VoteInfo,
     VoteListResponse, VoteResponse, VoterDetail, VoterListResponse, VoterResponse,
 };
-use cw_storage_plus::{AddrRef, Bound};
+use cw_storage_plus::Bound;
 
 use crate::error::ContractError;
 use crate::msg::{ExecuteMsg, InstantiateMsg, QueryMsg};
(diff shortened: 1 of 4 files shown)


maurolacy (approved):
I like the simplicity of this. I think `parse_key` is not used because we're implementing parsing of keys when we need it using custom helper functions.

ethanfrey:
>  I think parse_key is not used because we're implementing parsing of keys when we need it using custom helper functions.

There is an open issue to make use of `parse_key`: https://github.com/CosmWasm/cosmwasm-plus/issues/198 and thus avoid the helper functions.

It does simplify things to remove the call. Let's discuss whether we want to make #198 impossible first. If that is fine, let's do this approach. Otherwise, let's do the &K approach, which is quite nice

maurolacy:
> > I think parse_key is not used because we're implementing parsing of keys when we need it using custom helper functions.
> 
> There is an open issue to make use of `parse_key`: #198 and thus avoid the helper functions.

My understanding is that #198 can be done independently of `parse_key` (at least, implicitly, i.e. automatically). Anyway, consolidating `parse_key` functionality would be good.

> 
> It does simplify things to remove the call. Let's discuss whether we want to make #198 impossible first. If that is fine, let's do this approach. Otherwise, let's do the &K approach, which is quite nice

I agree #265 looks like the way to go. If we're OK with tolerating some syntax issues / constructs (mostly in test code), I would say, let's go for it.

ethanfrey:
I'm happy to merge this.

Let's add the `parse_key` when we want to take on #198 and we can do #265 at the same time, with a real benefit to such work

ethanfrey (approved):
Good solution.

ethanfrey:
Part was merged in #267 

The rest cherry-picked into #260 

All code used, just not this PR

PR #265, opened 2021-04-12
Take storake keys as reference (K -> &K)

webmaster128:
This change encourages the use of ownes types as `PrimaryKey`, which makes it very easy to deserialize bytes into owned keys. This can be implemented for `Addr`.

Since you cannot deserialize borrowed bytes into an owned object, a copy is created in `parse_key`. However, `parse_key` is never used.

In order to avoid copying around owned keys when thes are needed in multiple locations, the key arguments of all the storage functions were turned into references.

This is basically how the type system of standard library maps like https://doc.rust-lang.org/std/collections/struct.HashMap.html works: You have an owned key type `K` that is used plain in a few places. But [get](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.get) takes a reference. In `HashMap` the `K` can be of unknown size (e.g. `[u8]`). This is something we cannot do as long as we want to keep `parse_key`, which deserializes into an owned `K`.

--------

Do not merge as is. Target branch is a copy of the fork from #260.

### DIFF
--- a/packages/storage-plus/src/snapshot.rs
+++ b/packages/storage-plus/src/snapshot.rs
@@ -59,7 +59,7 @@ impl<'a, K, T> SnapshotMap<'a, K, T> {
 
     pub fn add_checkpoint(&self, store: &mut dyn Storage, height: u64) -> StdResult<()> {
         self.checkpoints
-            .update::<_, StdError>(store, height.into(), |count| {
+            .update::<_, StdError>(store, &height.into(), |count| {
                 Ok(count.unwrap_or_default() + 1)
             })?;
         Ok(())
@@ -68,13 +68,13 @@ impl<'a, K, T> SnapshotMap<'a, K, T> {
     pub fn remove_checkpoint(&self, store: &mut dyn Storage, height: u64) -> StdResult<()> {
         let count = self
             .checkpoints
-            .may_load(store, height.into())?
+            .may_load(store, &height.into())?
             .unwrap_or_default();
         if count <= 1 {
-            self.checkpoints.remove(store, height.into());
+            self.checkpoints.remove(store, &height.into());
             Ok(())
         } else {
-            self.checkpoints.save(store, height.into(), &(count - 1))
+            self.checkpoints.save(store, &height.into(), &(count - 1))
         }
     }
 }
@@ -84,7 +84,7 @@ where
     T: Serialize + DeserializeOwned + Clone,
     K: PrimaryKey<'a> + Prefixer<'a>,
 {
-    pub fn key(&self, k: K) -> Path<T> {
+    pub fn key(&self, k: &K) -> Path<T> {
         self.primary.key(k)
     }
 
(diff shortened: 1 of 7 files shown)


packages/storage-plus/src/snapshot.rs
```diff
@@ -297,24 +300,24 @@ mod tests {
     // Values at beginning of 3 -> A = 5, B = 7
     // Values at beginning of 5 -> A = 8, C = 13
     fn init_data(map: &TestMap, storage: &mut dyn Storage) {
-        map.save(storage, b"A", &5, 1).unwrap();
-        map.save(storage, b"B", &7, 2).unwrap();
+        map.save(storage, &b"A".as_ref(), &5, 1).unwrap();
```
maurolacy, line 303: This is introducing double references (`&&`). You could be defining the key as a value (`[u8]`) to avoid that; and all these would look nicer.

maurolacy (approved):
This is nice in that it really "forces" the use of references on the user.

But, it's not different from the case in which you define the key as a reference.
In the original case, the `&` is flexible / optional, and can go inside the `K`, whereas in this case the `&` is fixed (outside the `K`). And, you can still put it inside of it (leading to double references).

I prefer the former version, for simplicity. Even when you could be using values by defining a value key, that can be easily fixed (by changing the key definition in one place). Plus, you have the freedom to do that if you need to, for whatever reason.

That said, both versions are OK with me. The fact of `parse_key` deserializing to an owned value is irrelevant, in my opinion.

packages/storage-plus/src/snapshot.rs
maurolacy: If you define the `TestMap` using `[u8]` both references can be removed, by the way. One will be removed explicitly by you, the other will be "added" (deducted) by the compiler, and therefore does not need to be specified.
```suggestion
        map.save(storage, b"A", &5, 1).unwrap();
```

webmaster128:
> The fact of `parse_key` deserializing to an owned value is irrelevant, in my opinion.

This is crucial. Before you had `K` in and `K` out. If `K` was a reference, you created the requirement described by Ethan here: https://github.com/CosmWasm/cosmwasm/issues/868#issuecomment-816531741 which is almost impossible to fulfill by anything other than renamed `&[u8]`s. You cannot put `K` in and `*K` out.

But what you can do is put `&K` in and get `K` out, which is shown here.

This makes a few things a bit less fancy. But I think those get nicer over time when you migrate `K` from refs to values consistently. One problem you run into is that the literal `b"foo"` gives you an `&[u8; 3]` instead of an `&[u8]`, which makes some things slightly less pretty than before. But this is really just a problem of test code.

packages/storage-plus/src/snapshot.rs
webmaster128: `[u8]` (`u8` slice of unknown size) can be used if you have `K: ?Sized` (i.e. remove the `Sized` constraint that exists by default), see https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.get.

If you want to deserialize into this unsized `K`, the `parse_key` needs to put it on the heap and return something like `Box<K>` instead of `K`.

maurolacy:
Agreed. But as you mentioned, `parse_key` is currently not being used. I think we're creating our own parsers when we need them.

Anyway, both options are OK with me.

packages/storage-plus/src/snapshot.rs
maurolacy: That said, syntax is a bit uglier, but I think we can live with that, if we agree this solution is better in the end, in that it forces users to use references.

ethanfrey:
Let's revisit this when we tackle #198 

For now, going for the much more straightforward #264

PR #267, opened 2021-04-12
Remove unused PrimaryKey::parse_key

webmaster128:
This is the first commit of https://github.com/CosmWasm/cosmwasm-plus/pull/264.

It unlocks but does not include implementing `PrimaryKey<'a> for &'a Addr`


ethanfrey (approved):
Good stuff

PR #270, opened 2021-04-13
Run clippy on test code

webmaster128:


.circleci/config.yml
```diff
@@ -868,7 +868,7 @@ jobs:
           command: cargo fmt -- --check
       - run:
           name: Clippy linting on workspace
-          command: cargo clippy -- -D warnings
+          command: cargo clippy --tests -- -D warnings
```
ethanfrey, line 871: Nice idea. Did you check if this also runs clippy on non-test code (`cargo clippy --help` doesn't clarify that)

The [recommendation from clippy](https://github.com/rust-lang/rust-clippy#travis-ci) seems to be:

```
  # if you want the build job to fail when encountering warnings, use
  - cargo clippy -- -D warnings
  # in order to also check tests and non-default crate features, use
  - cargo clippy --all-targets --all-features -- -D warnings
```

ethanfrey (approved):
Good stuff and thanks to fix the tests below.

Just curious as to the best flags to run clippy with

.circleci/config.yml
webmaster128: Nice find. We cannot use `--all-features` because of

```
[features]
backtraces = ["cosmwasm-std/backtraces"]
```

which only compiles using Rust nightly. But `--all-targets` is a nice alternative to `--tetsts`.

PR #271, opened 2021-04-15
Indexed snapshot map

### LINKED ISSUE #255: Add secondary index support to SnapshotMap.
Extend the [SnapshotMap with a secondary index](https://github.com/CosmWasm/cosmwasm-plus/blob/master/contracts/cw4-group/src/state.rs#L10-L15). The historical queries (snapshots) are only for the primary key. The secondary index is not snapshotted. Maybe this needs a new type (IndexedSnapshotMap) that wraps a snapshot map?

This would be useful for eg. adding a secondary index based on weight to any of the group contracts that use a snapshot map. But in general to allow composibiity between these two advanced functions (snapshots and indexes) allows much more power to storage-plus

maurolacy:
Closes #255.

Straightforward "clone" of `IndexedMap`.

### DIFF
--- a/packages/storage-plus/src/indexed_snapshot.rs
+++ b/packages/storage-plus/src/indexed_snapshot.rs
@@ -146,7 +146,7 @@ where
     K: PrimaryKey<'a> + Prefixer<'a>,
     T: Serialize + DeserializeOwned + Clone,
     I: IndexList<T>,
-    K::Prefix: EmptyPrefix,
+    K::SubPrefix: EmptyPrefix,
 {
     // I would prefer not to copy code from Prefix, but no other way
     // with lifetimes (create Prefix inside function and return ref = no no)
@@ -160,7 +160,8 @@ where
     where
         T: 'c,
     {
-        self.prefix(K::Prefix::new()).range(store, min, max, order)
+        self.sub_prefix(K::SubPrefix::new())
+            .range(store, min, max, order)
     }
 }
 
--- a/packages/storage-plus/src/indexed_snapshot.rs
+++ b/packages/storage-plus/src/indexed_snapshot.rs
@@ -103,8 +103,8 @@ where
         &self,
         store: &mut dyn Storage,
         key: K,
-        action: A,
         height: u64,
+        action: A,
     ) -> Result<T, E>
     where
         A: FnOnce(Option<T>) -> Result<T, E>,
@@ -600,17 +600,12 @@ mod test {
         height += 1;
 
         // change john to mary
-        map.update(
-            &mut store,
-            pks[2],
-            |d| -> StdResult<_> {
-                let mut x = d.unwrap();
-                assert_eq!(&x.name, "John");
-                x.name = "Mary".to_string();
-                Ok(x)
-            },
-            height,
-        )
+        map.update(&mut store, pks[2], height, |d| -> StdResult<_> {
+            let mut x = d.unwrap();
+            assert_eq!(&x.name, "John");
+            x.name = "Mary".to_string();
+            Ok(x)
+        })
         .unwrap();
 
         // find 1 maria, 1 maria luisa, no john, and 1 mary
--- a/packages/storage-plus/src/indexed_snapshot.rs
+++ b/packages/storage-plus/src/indexed_snapshot.rs
@@ -1,0 +1,670 @@
+// this module requires iterator to be useful at all
+#![cfg(feature = "iterator")]
+
+use cosmwasm_std::{StdError, StdResult, Storage};
+use serde::de::DeserializeOwned;
+use serde::Serialize;
+
+use crate::keys::{EmptyPrefix, Prefixer, PrimaryKey};
+use crate::prefix::{Bound, Prefix};
+use crate::snapshot::SnapshotMap;
+use crate::{IndexList, Strategy};
+
+/// IndexedSnapshotMap works like a SnapshotMap but has a secondary index
+/// TODO: remove traits here and make this const fn new
+pub struct IndexedSnapshotMap<'a, K, T, I>
+where
+    K: PrimaryKey<'a>,
+    T: Serialize + DeserializeOwned + Clone,
+    I: IndexList<T>,
+{
+    pk_namespace: &'a [u8],
+    primary: SnapshotMap<'a, K, T>,
+    /// This is meant to be read directly to get the proper types, like:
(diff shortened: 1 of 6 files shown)


packages/storage-plus/src/indexed_snapshot.rs
```diff
@@ -0,0 +1,677 @@
...
+use serde::de::DeserializeOwned;
+use serde::Serialize;
+
+use crate::keys::{EmptyPrefix, Prefixer, PrimaryKey};
+use crate::prefix::{Bound, Prefix};
+use crate::snapshot::SnapshotMap;
+use crate::{IndexList, Strategy};
+
+/// IndexedSnapshotMap works like a SnapshotMap but has a secondary index
+/// TODO: remove traits here and make this const fn new
```
ethanfrey, line 14: Copied comment I presume - I think the const fn new is impossible.

packages/storage-plus/src/indexed_snapshot.rs
```diff
@@ -0,0 +1,677 @@
...
+pub struct IndexedSnapshotMap<'a, K, T, I>
+where
+    K: PrimaryKey<'a>,
+    T: Serialize + DeserializeOwned + Clone,
+    I: IndexList<T>,
+{
+    pk_namespace: &'a [u8],
+    primary: SnapshotMap<'a, K, T>,
```
ethanfrey, line 22: This is a copy of indexed_map.rs with some mods, right?

But how do you make the historical queries? Those special functions of SnapshotMap need to be exposed somehow.

ethanfrey (commented):
You need to expose the SnapshotMap methods.
A straight copy is okay, but maybe a bit hard to maintain.

------

One thought (which may be misguided, but please consider) is to make some `IndexedCollection` which takes a trait for `primary`. It needs:

```rust
trait Indexable {
  save(&self, ...);
  remove(&self, ...);
  may_load(&self, ...);
  // and maybe
  load(&self, ...);
}
```

Then you can wrap either a Map or IndexedMap as "Indexable".

In order to expose the custom methods on SnapshotMap, one could make `pub primary: T <T: Indexable>`, and call `indexed.primary.load_at_height(...)` for example.

(Note, maybe it is best that `Indexable` uses different names, so the common names don't require importing a trait, just a trait wrapper that defers to those, but you get the idea).

Then again, the majority of the work is done inside the Indexes (not IndexedMap), so maybe this is the simplest and most stable solution.

packages/storage-plus/src/indexed_snapshot.rs
maurolacy: I'm now working on using this `IndexedSnapshotMap` to implement `list_members` in `cw4-group`.

Looks good so far, but I'm making some shortcuts:

I'm making the `primary` member of `IndexedSnapshotMap` `pub`, so that I can access the wrapped `SnapshotMap` through it directly. This is not very good, but it's simple, and it's what is being done for accessing the `idx`s.

An alternative would be to wrap the needed methods; or use the `Deref` trait trick to expose them.

Another alternative would be, just create an external, similar `SnapshotMap`s. Sone one is inside the `Indexed*` wrapper, the other is not (and can be `const`), like the current `MEMBERS` is. These are just namespaces in the end.

And, even another alternative, would be to pass a `SnapshotMap` instance as parameter for the `IndexedSnapshotMap` constructor.

More about this later.

Other shortcut is: I'm just re-implementing `list_members`, and the associated `ListMembers`  message, using this new indexed map. Maybe you would prefer having two different messages? i.e. `ListMembers` which is the traditional, unordered functionality, and a `ListMembersByWeight` which implements members ordered by descending weight.

Another option would be to have a message called `TopMembers`, which can implement limits but not pagination, by example.

Good news is that it's working. I'm having some issues with prefixes and ranges, but the overall functionality is already working.

ethanfrey: > Maybe you would prefer having two different messages? i.e. ListMembers which is the traditional, unordered functionality, and a ListMembersByWeight which implements members ordered by (descending) weight.

I would prefer a new method. I think the existing implementation works well enough. You can gladly add an extra method to `cw4-group` (but not `cw4` spec) to demo this.

ethanfrey: > I'm making the primary member of IndexedSnapshotMap pub, so that I can access the wrapped SnapshotMap through it directly. This is not very good, but it's simple, and it's what is being done for accessing the idxs.

This is a very good choice for trying it out.

> An alternative to that would be to wrap the needed methods; or use the Deref trait trick to expose them.

`Deref` seems too much of a hack here (this is far from a "smart pointer") and there are not so many methods, so I would just manually add them.

(And on later reflection, I think making this generic over Map and SnapshotMap is more complexity than it is worth and we will end up with something we will have trouble maintaining)

maurolacy: See #272 for a (working) impl. We can then discuss some details.

I'll separate it into two different calls, as mentioned above, when finding some time tomorrow.

maurolacy: > > I'm making the primary member of IndexedSnapshotMap pub, so that I can access the wrapped SnapshotMap through it directly. This is not very good, but it's simple, and it's what is being done for accessing the idxs.
> 
> This is a very good choice for trying it out.

We can simply do that then.

> 
> > An alternative to that would be to wrap the needed methods; or use the Deref trait trick to expose them.
> 
> `Deref` seems too much of a hack here (this is far from a "smart pointer") and there are not so many methods, so I would just manually add them.
> 
> (And on later reflection, I think making this generic over Map and SnapshotMap is more complexity than it is worth and we will end up with something we will have trouble maintaining)

Agreed.

ethanfrey:
Looks good, so this is the basic implementation, and I look at #272 for usage.

Do you want to merge those together and then go to master? Or merge this first, then the other when it is ready?

In either case, happy to review it as is, and give some pointers. But let's revert the `pub primary` once we have finalized the design and before merging to master.

ethanfrey:
Once #272 passes CI with full tests, I would just address this comment as part of this PR: https://github.com/CosmWasm/cosmwasm-plus/pull/272#discussion_r614703523

Otherwise, looking good

maurolacy:
Thanks for the comments. I would merge this first, as it's done and contained.

And yes, we can always revert the `pub primary` in the other branch, when the proper methods are exposed.

Issue #27, opened 2020-07-28
Add cw20 support to atomic swap contract


ethanfrey:
Builds on #26

Just like `cw20-escrow` supports both native tokens and cw20 tokens (via `Receive`), the atomic swap contract should be extended with this. We do not want a `TopUp` like `cw20-escrow`, but the creation via [direct function call](https://github.com/CosmWasm/cosmwasm-plus/blob/master/contracts/cw20-escrow/src/contract.rs#L56-L88) or via the [cw20::Receive entry point](https://github.com/CosmWasm/cosmwasm-plus/blob/master/contracts/cw20-escrow/src/contract.rs#L90-L123) serve as a useful model on how to implement this.

To test it, try swaping a native token vs a cw20 token using the same CosmWasm contract for both sides.

Once completed we can swap any native or cw20 token against eth or any erc20 token. Well.. we would need a UI for that, but the contracts are ready.

maurolacy:
Initial implementation is under the [cw20-atomic-swaps](https://github.com/CosmWasm/cosmwasm-plus/tree/cw20-atomic-swaps) branch. Unit tests to follow; PR to follow after it.

I implemented functionality for the two tokens (native vs. cw20) into the same functions, for simplicity and consistency. Can separate them into different functions for clarity, if you think it would be better.

maurolacy:
There are some `FIXME?` in the code: basically, what if somebody sents a non-empty array of native coins, or a cw20 coin, but sets their amounts to zero? Do we need to check / reject those messages? Or are they rejected by the SDK?

Another question: Do I need to improve the logs? Log the sender and amounts on creation, by example.

maurolacy:
PR: #72.

ethanfrey:
Counting zero amounts as missing is a good check. I would filter out zeros in the native coins, and then ensure it has length >0 and only save the filtered list.

For cw20 coins, yeah, rejecting 0 amounts sounds good. We should also reject such sends in cw20-base. Can you make an issue for that?

maurolacy:
Sure. #76 created.

ethanfrey:
Closed by #72

PR #272, opened 2021-04-15
Cw4 group list members by weight

maurolacy:
Uses `IndexedSnapshotMap` to do the `list_members` query sorted by weight.

Continuation of #255. Uses #271.

### DIFF
--- a/contracts/cw4-group/src/contract.rs
+++ b/contracts/cw4-group/src/contract.rs
@@ -176,7 +176,7 @@ fn query_total_weight(deps: Deps) -> StdResult<TotalWeightResponse> {
 fn query_member(deps: Deps, addr: String, height: Option<u64>) -> StdResult<MemberResponse> {
     let addr = deps.api.addr_validate(&addr)?;
     let weight = match height {
-        Some(h) => members().primary.may_load_at_height(deps.storage, &addr, h),
+        Some(h) => members().may_load_at_height(deps.storage, &addr, h),
         None => members().may_load(deps.storage, &addr),
     }?;
     Ok(MemberResponse { weight })
--- a/contracts/cw4-group/src/contract.rs
+++ b/contracts/cw4-group/src/contract.rs
@@ -222,7 +222,7 @@ fn list_members_by_weight(
     let members: StdResult<Vec<_>> = members()
         .idx
         .weight
-        .range(deps.storage, start, None, Order::Descending)
+        .range(deps.storage, None, start, Order::Descending)
         .take(limit)
         .map(|item| {
             let (key, weight) = item?;
--- a/contracts/cw4-group/src/contract.rs
+++ b/contracts/cw4-group/src/contract.rs
@@ -308,9 +308,10 @@ mod tests {
         let members = list_members(deps.as_ref(), None, None).unwrap();
         assert_eq!(members.members.len(), 2);
         // TODO: assert the set is proper
-
-        // TODO: Test pagination / limits
     }
+
+    // TODO: Test member_by_weight
+    // Test pagination / limits
 
     fn assert_users<S: Storage, A: Api, Q: Querier>(
         deps: &OwnedDeps<S, A, Q>,
--- a/contracts/cw4-group/src/contract.rs
+++ b/contracts/cw4-group/src/contract.rs
@@ -232,7 +232,6 @@ fn list_members_by_weight(
     let members: StdResult<Vec<_>> = members()
         .idx
         .weight
-        .sub_prefix(()) // FIXME: Use range() directly, and impl EmptyPrefix for IntKey<T>?
         .range(deps.storage, start, None, Order::Descending)
         .take(limit)
         .map(|item| {
--- a/contracts/cw4-group/src/contract.rs
+++ b/contracts/cw4-group/src/contract.rs
@@ -193,7 +193,7 @@ fn list_members(
 ) -> StdResult<MemberListResponse> {
     let limit = limit.unwrap_or(DEFAULT_LIMIT).min(MAX_LIMIT) as usize;
     let addr = maybe_addr(deps.api, start_after)?;
-    let start = addr.map(|addr| Bound::exclusive(addr.to_string()));
+    let start = addr.map(|addr| Bound::exclusive(addr.as_ref()));
 
     let members: StdResult<Vec<_>> = members()
         .range(deps.storage, start, None, Order::Ascending)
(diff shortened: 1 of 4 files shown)


contracts/cw4-group/src/contract.rs
```diff
@@ -185,15 +184,28 @@ const DEFAULT_LIMIT: u32 = 10;
...
+            .concat(),
+        )
+    });
 
-    let members: StdResult<Vec<_>> = MEMBERS
-        .range(deps.storage, start, None, Order::Ascending)
+    let members: StdResult<Vec<_>> = members()
+        .idx
+        .weight
+        .sub_prefix(()) // FIXME: Use range() directly, and impl EmptyPrefix for IntKey<T>?
```
maurolacy, line 207: Removing this line, we get:
```
error[E0599]: the method `range` exists for struct `MultiIndex<'_, (cw_storage_plus::keys::IntKey<u64>, PkOwned), u64>`, but its trait bounds were not satisfied
   --> contracts/cw4-group/src/contract.rs:207:10
    |
207 |         .range(deps.storage, start, None, Order::Descending)
    |          ^^^^^ method cannot be called on `MultiIndex<'_, (cw_storage_plus::keys::IntKey<u64>, PkOwned), u64>` due to unsatisfied trait bounds
    | 
   ::: /usr2/home/mauro/work/cosmwasm-plus/packages/storage-plus/src/keys.rs:193:1
    |
193 | pub struct IntKey<T: Endian> {
    | ---------------------------- doesn't satisfy `_: cw_storage_plus::keys::EmptyPrefix`
    |
    = note: the following trait bounds were not satisfied:
            `cw_storage_plus::keys::IntKey<u64>: cw_storage_plus::keys::EmptyPrefix`

```

maurolacy: This is because a `Prefix` for a tuple is of course not an `EmptyPrefix`. The solution straightforward: use `SubPrefix: EmptyPrefix` as marker type instead.

Will do the same for the other top-level `range()`s.

contracts/cw4-group/src/contract.rs
```diff
@@ -173,8 +177,8 @@ fn query_total_weight(deps: Deps) -> StdResult<TotalWeightResponse> {
 fn query_member(deps: Deps, addr: String, height: Option<u64>) -> StdResult<MemberResponse> {
     let addr = deps.api.addr_validate(&addr)?;
     let weight = match height {
-        Some(h) => MEMBERS.may_load_at_height(deps.storage, &addr, h),
-        None => MEMBERS.may_load(deps.storage, &addr),
+        Some(h) => members().primary.may_load_at_height(deps.storage, &addr, h),
```
ethanfrey, line 180: Thing to expose in `IndexedSnapshotMap`: 
* `may_load_at_height`

ethanfrey: That should be enough for now.

In theory:
* `add_checkpoint`
* `remove_checkpoint`

would be good as well, but those are not used in any code yet (no `Strategy::Selected`)

contracts/cw4-group/src/contract.rs
```diff
@@ -207,6 +211,42 @@ fn list_members(
     Ok(MemberListResponse { members: members? })
 }
 
+fn list_members_by_weight(
+    deps: Deps,
+    start_after: Option<(u64, String)>,
+    limit: Option<u32>,
+) -> StdResult<MemberListResponse> {
+    let limit = limit.unwrap_or(DEFAULT_LIMIT).min(MAX_LIMIT) as usize;
+    // FIXME: Avoid building the (length-prefixed) bound manually
```
ethanfrey, line 220: Can't you use `.prefix((u64, String)).range(deps.storage, None, None, Order:Descending)`?

Ah, I see this is different. Definitely worth adding a helper function to storage-plus for this to encode from a key.
Maybe it is there?

```rust
  let start = start_after.map(|(w, a)| {
        let key = (UintKey64::from(w), &s).joined_key();
        Bound::exclusive(key)
     });
```

contracts/cw4-group/src/contract.rs
```diff
@@ -270,6 +310,9 @@ mod tests {
         // TODO: assert the set is proper
     }
 
+    // TODO: Test member_by_weight
```
ethanfrey, line 313: Yup, this just seems to miss a test, otherwise this approach looks good.

ethanfrey (commented):
Nice start. Added some comments

contracts/cw4-group/src/contract.rs
maurolacy: Nice. Thanks for that. A slightly different version compiles:
```
    let start =
        start_after.map(|(w, a)| Bound::exclusive((U64Key::from(w), a.as_str()).joined_key()))```

ethanfrey: Maybe we could make this easier by exposing a method on `Map` (and all derivatives). Wait.. there is `Map.key(K)`. And we have a [Deref to &[u8] from Path](https://github.com/CosmWasm/cosmwasm-plus/blob/master/packages/storage-plus/src/path.rs#L20-L29). Maybe adding `Into<Vec<u8>>` would make this more efficient, but

`Bound::exclusive(&map.key((w.into(), &a)))` should probably work as well.

Feel free to add any helpers or docs as you see (and expose `key` on the various `Map` derivatives). We should make this easy.

maurolacy: I tried using the `Map::key(K)` variant, but that requires a map of the proper key type.

I think it's OK now. We can always revisit / improve on this on another iteration.

contracts/cw4-group/src/contract.rs
maurolacy: Implemented in #275.

contracts/cw4-group/src/contract.rs
maurolacy: For indexed maps, I think we could add a `primary_key(K)` method that wraps `self.primary.key()`. That to differentiate it from the index keys.

Then, each index can have a `key()` method, that helps in building the index key. Not sure how to deal with generics here. We would probably need to add a `IK` trait to indexes. Need to think about it, play with it to know how to do it. Will open an issue for trying and doing this.

maurolacy: Createdhttps://github.com/CosmWasm/cosmwasm-plus/issues/277 to track / implement this.

maurolacy:
Closed in favour of adding this functionality to tgrade. See [tgrade issue #27](https://github.com/confio/tgrade-contracts/issues/27) and [pull #34](https://github.com/confio/tgrade-contracts/pull/34).

PR #274, opened 2021-04-17
Cw4 stake list members by weight

maurolacy:
Similar to #272 but for `cw4-stake`.

I think we should consider moving / adding this functionality to the `cw4` spec, and implement `list_members_by_weight` under `Cw4Contract`, to avoid repetition, and ease maintenance / use.

### DIFF
--- a/contracts/cw4-stake/schema/query_msg.json
+++ b/contracts/cw4-stake/schema/query_msg.json
@@ -99,6 +99,47 @@
       "additionalProperties": false
     },
     {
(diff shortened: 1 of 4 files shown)


maurolacy:
Closed in favour of adding this functionality to tgrade. See [tgrade issue #27](https://github.com/confio/tgrade-contracts/issues/27) and [pull #34](https://github.com/confio/tgrade-contracts/pull/34).

PR #275, opened 2021-04-17
Indexed snapshot. Expose primary methods

### LINKED ISSUE #255: Add secondary index support to SnapshotMap.

maurolacy:
Completes / improves #255.

Let's merge this first, and then rebase #272 and #274 on top, and adjust.

### DIFF
--- a/packages/storage-plus/src/indexed_snapshot.rs
+++ b/packages/storage-plus/src/indexed_snapshot.rs
@@ -18,7 +18,7 @@ where
     I: IndexList<T>,
 {
     pk_namespace: &'a [u8],
-    pub primary: SnapshotMap<'a, K, T>,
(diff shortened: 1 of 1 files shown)


ethanfrey (approved):
nice

PR #279, opened 2021-04-20
Better error message with missing sudo (no parse error)

### LINKED ISSUE #278: `Sudo` over no `new_with_sudo` contract wrapper error message
In the multi-test context, to send the `Sudo` message to a contract requires creating the contract wrapper with `new_with_sudo`.
This is obvious, but when you forget to do it, the resulting error message is not clear / obvious at all:
```rust
`Result::unwrap()` on an `Err` value: "Error parsing into type alloc::string::String: Invalid type"
```

Consider improving on the error message, so a contract / test writer has a better grasp on what's failing when forgetting / not knowing about `new_with_sudo`.

ethanfrey:
Closes #278 

I introduced `Any` rather than `String`. `Any` is like `Empty` (parses any struct), but it has a needed `Display` implementation.
When we are happy with this, and merged, we can port over the `Display` implementation to `cosmwasm_std::Empty` for the future.

### DIFF
--- a/packages/multi-test/src/wasm.rs
+++ b/packages/multi-test/src/wasm.rs
@@ -1,5 +1,5 @@
 use schemars::JsonSchema;
-use serde::de::DeserializeOwned;
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use std::collections::HashMap;
 use std::fmt;
 use std::ops::Deref;
@@ -10,6 +10,18 @@ use cosmwasm_std::{
 };
 
 use crate::transactions::{RepLog, StorageTransaction};
+use std::fmt::Formatter;
+
+/// This is used as a placeholder type that can be anything
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct Any {}
+
+// We couldn't use Empty as it didn't implement Display. Shall we upstream this?
+impl fmt::Display for Any {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        f.write_str("{Any}")
+    }
+}
 
 /// Interface to call into a Contract
 pub trait Contract<T>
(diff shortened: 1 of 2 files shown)


packages/multi-test/src/wasm.rs
```diff
@@ -10,6 +10,18 @@ use cosmwasm_std::{
 };
 
 use crate::transactions::{RepLog, StorageTransaction};
+use std::fmt::Formatter;
+
+/// This is used as a placeholder type that can be anything
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct Any {}
+
+// We couldn't use Empty as it didn't implement Display. Shall we upstream this?
```
maurolacy, line 19: I think so, yes. This can be a (minor) recurrent problem.

maurolacy (approved):

packages/multi-test/src/wasm.rs
```diff
@@ -48,7 +60,7 @@ type QueryClosure<T, E> = Box<dyn Fn(Deps, Env, T) -> Result<Binary, E>>;
 
 /// Wraps the exported functions from a contract and provides the normalized format
 /// Place T4 and E4 at the end, as we just want default placeholders for most contracts that don't have sudo
-pub struct ContractWrapper<T1, T2, T3, E1, E2, E3, C = Empty, T4 = String, E4 = String>
+pub struct ContractWrapper<T1, T2, T3, E1, E2, E3, C = Empty, T4 = Any, E4 = Any>
```
webmaster128, line 63: Why don't you use `T4 = Empty` and `E4 = String`?

`String` is a great default error type. The only restructions those types have is `ToString`. So requiring `Display` is more than necessary.

ethanfrey: Hmmm.. that might work and be a simpler solution.
Will try

ethanfrey: Just approved your PR to that end

PR #280, opened 2021-04-20
Update to 0.14.0 beta4

maurolacy:


contracts/cw1-subkeys/src/contract.rs
```diff
@@ -147,11 +147,6 @@ pub fn check_staking_permissions(
                 return Err(ContractError::ReDelegatePerm {});
             }
         }
-        StakingMsg::Withdraw { .. } => {
```
ethanfrey, line 150: Hmmm... this withdraw permission just disappears?
I would add the support for the distribution messages that replace it

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -1497,13 +1475,6 @@ mod tests {
         )
         .unwrap();
         assert_eq!(res.can_execute, true);
-        let res = query_can_execute(
```
ethanfrey, line 1500: The new `Distribution::Withdraw` method should be supported in this high level interface. (People think of it as staking). Not sure how you want to divide the internal processing, but this API should remain similar

contracts/cw20-staking/src/contract.rs
```diff
@@ -356,9 +356,8 @@ pub fn reinvest(deps: DepsMut, env: Env, _info: MessageInfo) -> Result<Response,
     let res = Response {
         submessages: vec![],
         messages: vec![
-            StakingMsg::Withdraw {
+            DistributionMsg::WithdrawDelegatorReward {
```
ethanfrey, line 359: :+1:

ethanfrey (commented):
Generally good.
Let's ensure the cw1-subkeys still allows withdrawls

contracts/cw1-subkeys/src/contract.rs
maurolacy: FIxed.

ethanfrey (approved):
Nice fixes. Thank you

PR #281, opened 2021-04-21
Avoid the need for Any type by using Empty as message type and String as error type

webmaster128:


ethanfrey (approved):
Looks reasonable and better solution.

PR #282, opened 2021-04-21
Use Cw20 token in cw4-stake

### LINKED ISSUE #143: Add cw20 support to token-weighted group

ethanfrey:
Closes #143 

- [x] Update implementation to support cw20 tokens everywhere.
- [x] Existing tests pass
- [x] Add tests for the cw20 use case

### DIFF
--- a/contracts/cw4-stake/src/contract.rs
+++ b/contracts/cw4-stake/src/contract.rs
@@ -2,12 +2,12 @@
 use cosmwasm_std::entry_point;
 use cosmwasm_std::{
     attr, coin, coins, from_slice, to_binary, Addr, BankMsg, Binary, Coin, CosmosMsg, Deps,
-    DepsMut, Env, MessageInfo, Order, Response, StdError, StdResult, Storage, Uint128,
+    DepsMut, Env, MessageInfo, Order, Response, StdError, StdResult, Storage, Uint128, WasmMsg,
 };
 
 use cw0::{maybe_addr, NativeBalance};
 use cw2::set_contract_version;
-use cw20::{Balance, Cw20CoinVerified, Cw20ReceiveMsg, Denom};
+use cw20::{Balance, Cw20CoinVerified, Cw20ExecuteMsg, Cw20ReceiveMsg, Denom};
 use cw4::{
     Member, MemberChangedHookMsg, MemberDiff, MemberListResponse, MemberResponse,
     TotalWeightResponse,
@@ -262,20 +262,32 @@ pub fn execute_claim(
     }
 
     let config = CONFIG.load(deps.storage)?;
-    let amount;
-    match &config.denom {
-        Denom::Native(denom) => amount = coins(release.u128(), denom),
-        Denom::Cw20(_canonical_addr) => {
-            unimplemented!("The CW20 coins release functionality is in progress")
+    let (amount_str, message) = match &config.denom {
+        Denom::Native(denom) => {
+            let amount = coins(release.u128(), denom);
+            let amount_str = coins_to_string(&amount);
+            let message = BankMsg::Send {
+                to_address: info.sender.clone().into(),
+                amount,
+            }
+            .into();
+            (amount_str, message)
         }
-    }
-
-    let amount_str = coins_to_string(&amount);
-    let messages = vec![BankMsg::Send {
-        to_address: info.sender.clone().into(),
-        amount,
-    }
-    .into()];
+        Denom::Cw20(canonical_addr) => {
+            let amount_str = format!("{}{}", release.u128(), canonical_addr);
+            let transfer = Cw20ExecuteMsg::Transfer {
+                recipient: info.sender.clone().into(),
+                amount: release,
+            };
+            let message = WasmMsg::Execute {
+                contract_addr: canonical_addr.into(),
+                msg: to_binary(&transfer)?,
+                send: vec![],
+            }
+            .into();
+            (amount_str, message)
+        }
+    };
 
     let attributes = vec![
         attr("action", "claim"),
@@ -284,7 +296,7 @@ pub fn execute_claim(
     ];
     Ok(Response {
         submessages: vec![],
-        messages,
+        messages: vec![message],
         attributes,
         data: None,
     })
--- a/contracts/cw4-stake/src/contract.rs
+++ b/contracts/cw4-stake/src/contract.rs
@@ -1,8 +1,8 @@
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::entry_point;
 use cosmwasm_std::{
-    attr, coin, coins, from_slice, to_binary, Addr, BankMsg, Binary, Coin, CosmosMsg, Deps,
-    DepsMut, Env, MessageInfo, Order, Response, StdError, StdResult, Storage, Uint128, WasmMsg,
+    attr, coins, from_slice, to_binary, Addr, BankMsg, Binary, Coin, CosmosMsg, Deps, DepsMut, Env,
+    MessageInfo, Order, Response, StdResult, Storage, Uint128, WasmMsg,
 };
 
 use cw0::{maybe_addr, NativeBalance};
@@ -339,17 +339,8 @@ fn query_total_weight(deps: Deps) -> StdResult<TotalWeightResponse> {
 pub fn query_staked(deps: Deps, addr: String) -> StdResult<StakedResponse> {
     let addr = deps.api.addr_validate(&addr)?;
     let stake = STAKE.may_load(deps.storage, &addr)?.unwrap_or_default();
-    let denom = match CONFIG.load(deps.storage)?.denom {
-        Denom::Native(want) => want,
-        _ => {
-            return Err(StdError::generic_err(
-                "The stake for CW20 is not yet implemented",
-            ));
-        }
-    };
-    Ok(StakedResponse {
-        stake: coin(stake.u128(), denom),
-    })
+    let denom = CONFIG.load(deps.storage)?.denom;
+    Ok(StakedResponse { stake, denom })
 }
 
 fn query_member(deps: Deps, addr: String, height: Option<u64>) -> StdResult<MemberResponse> {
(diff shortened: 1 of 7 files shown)


maurolacy:
Now reviewing.

contracts/cw4-stake/src/contract.rs
```diff
@@ -239,20 +262,32 @@ pub fn execute_claim(
...
-    match &config.denom {
-        Denom::Native(denom) => amount = coins(release.u128(), denom),
-        Denom::Cw20(_canonical_addr) => {
-            unimplemented!("The CW20 coins release functionality is in progress")
+    let (amount_str, message) = match &config.denom {
+        Denom::Native(denom) => {
+            let amount = coins(release.u128(), denom);
+            let amount_str = coins_to_string(&amount);
+            let message = BankMsg::Send {
+                to_address: info.sender.clone().into(),
```
maurolacy, line 270:
```suggestion
                to_address: info.sender.to_string(),
```
saves some chars.

contracts/cw4-stake/src/contract.rs
```diff
@@ -239,20 +262,32 @@ pub fn execute_claim(
...
-    }
-    .into()];
+        Denom::Cw20(canonical_addr) => {
+            let amount_str = format!("{}{}", release.u128(), canonical_addr);
+            let transfer = Cw20ExecuteMsg::Transfer {
+                recipient: info.sender.clone().into(),
+                amount: release,
+            };
+            let message = WasmMsg::Execute {
+                contract_addr: canonical_addr.into(),
```
maurolacy, line 283:
```suggestion
       Denom::Cw20(addr) => {
            let amount = coins(release.u128(), addr);
            let amount_str = coins_to_string(&amount);
            let transfer = Cw20ExecuteMsg::Transfer {
                recipient: info.sender.to_string(),
                amount: release,
            };
            let message = WasmMsg::Execute {
                contract_addr: addr.into(),
```

contracts/cw4-stake/src/contract.rs
```diff
@@ -304,17 +339,8 @@ fn query_total_weight(deps: Deps) -> StdResult<TotalWeightResponse> {
...
-            return Err(StdError::generic_err(
-                "The stake for CW20 is not yet implemented",
-            ));
-        }
-    };
-    Ok(StakedResponse {
-        stake: coin(stake.u128(), denom),
-    })
+    let denom = CONFIG.load(deps.storage)?.denom;
+    Ok(StakedResponse { stake, denom })
```
maurolacy, line 343: Nice.

contracts/cw4-stake/src/contract.rs
maurolacy: Also, in `coins_to_string()` below, the format might better be:
```rust
.map(|c| format!("{} {}", c.amount, c.denom))
```
with a space between amount and denom; just for clarity.

maurolacy (approved):
Looks good.

Will review tests later during the afternoon, and add isolated comments if I find anything.

contracts/cw4-stake/src/contract.rs
```diff
@@ -574,6 +630,78 @@ mod tests {
...
+        let mut env = mock_env();
+        env.block.height += 10;
+        let expires = unbonding.after(&env.block);
+        assert_eq!(
+            get_claims(deps.as_ref(), &Addr::unchecked(USER1)),
+            vec![Claim::new(7_900, expires)]
+        );
+
+        // wait til they expire and get payout
+        env.block.height += 60;
```
maurolacy, line 675:
```suggestion
        env.block.height += 50;
```
will do.

contracts/cw4-stake/src/contract.rs
```diff
@@ -574,6 +630,78 @@ mod tests {
...
+        // unbond part
+        unbond(deps.as_mut(), 7_900, 4_600, 0, 10);
+
+        // Assert updated weights
+        assert_stake(deps.as_ref(), 12_100, 8_900, 500);
+        assert_users(deps.as_ref(), Some(12), Some(8), None, None);
+
+        // with proper claims
+        let mut env = mock_env();
+        env.block.height += 10;
```
maurolacy, line 667: It would be nice to have a `const HEIGHT_DELTA` or so, and use it here and above; just for clarity.

contracts/cw4-stake/src/contract.rs
```diff
@@ -415,6 +454,23 @@ mod tests {
         }
     }
 
+    fn bond_cw20(mut deps: DepsMut, user1: u128, user2: u128, user3: u128, height_delta: u64) {
```
maurolacy, line 457: Nice cw20 helpers.

contracts/cw4-stake/src/contract.rs
```diff
@@ -129,6 +130,28 @@ pub fn execute_bond(
...
+        None => Err(ContractError::NoData {}),
+    }?;
+    let balance = Balance::Cw20(Cw20CoinVerified {
+        address: info.sender,
+        amount: wrapper.amount,
+    });
+    let api = deps.api;
+    match msg {
+        ReceiveMsg::Bond {} => {
+            execute_bond(deps, env, balance, api.addr_validate(&wrapper.sender)?)
```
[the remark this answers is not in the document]
ethanfrey: Let me check this again.

`info.sender` is the address of the cw20 contract (that re-sent this message).
`wrapper.sender` is the address of the user that requested the cw20 contract to send this. This cannot be fully trusted (the cw20 contract can fake it), so only use it for things in the address's favor (like paying/bonding tokens, not withdrawls)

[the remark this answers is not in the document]
ethanfrey: It is used as I described. I added a comment on it.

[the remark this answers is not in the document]
maurolacy: Thanks for the clarification.

contracts/cw4-stake/src/contract.rs
ethanfrey: Changed the format and used the same function to format in both variants of the match

PR #284, opened 2021-04-21
Add job to test optimizer in CI

maurolacy:


maurolacy:
Closed this because it was being used just for experimentation. Will do some more experimentation, to see if we can understand this better.

PR #286, opened 2021-04-26
Cw20ReceiveMsg msg field

### LINKED ISSUE #283: Make message required in Cw20ReceiveMsg
It is defined as `Option<Binary>`, but we *always* use `Some(...)`.
There is no reason not to send one and this only makes confusion as came up with the cw1155 design.

Let's just make `msg: Binary` and update the apps to use it. (It is JSON compatible for all those who send with values)

maurolacy:
Closes #283.


ethanfrey (approved):
Good fix. Thanks.

PR #288, opened 2021-04-29
Primary key / index key helpers

### LINKED ISSUE #277: Improve on indexed maps primary key / index keys helpers
It would be nice to have a `key(K)` method handy for every indexed map type (`IndexedMap` and `IndexedSnapshotMap`, currently), so that building a `range()` (primary) key is easier, and well documented.

The same for every index, i.e. a `key(IK)` (or `index_key(IK)`, for clarity) for every index an indexed map has, so that building a `range()` key over an index is easy.

`key(K)` is already implemented for `Map`, and probably other map types. Creating this issue to extend / consolidate functionality accross all map types and their indexes.

maurolacy:
Closes #277.

### DIFF
--- a/packages/storage-plus/src/indexed_map.rs
+++ b/packages/storage-plus/src/indexed_map.rs
@@ -342,7 +342,7 @@ mod test {
             .count();
         assert_eq!(0, count);
 
-        // index_key() with empty pk works
+        // index_key() over MultiIndex works (empty pk)
         // In a MultiIndex, an index key is composed by the index and the primary key.
         // Primary key may be empty (so that to iterate over all elements that match just the index)
         let key = (PkOwned(b"Maria".to_vec()), PkOwned(b"".to_vec()));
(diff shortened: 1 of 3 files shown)


packages/storage-plus/src/indexed_map.rs
```diff
@@ -337,6 +342,19 @@ mod test {
             .count();
         assert_eq!(0, count);
 
+        // index_key() works
+        let key = map
+            .idx
+            .name
+            .index_key((PkOwned(b"Maria".to_vec()), PkOwned(b"1".to_vec())));
```
ethanfrey, line 349: This is very unclear what the second item is.
Is this the primary key that she is stored under? (pk)?

If so, let's filter with only some of them (eg. all Marias with pk >= 2)

packages/storage-plus/src/indexes.rs
```diff
@@ -138,6 +138,10 @@ where
         )
     }
 
+    pub fn index_key(&self, k: K) -> Vec<u8> {
```
ethanfrey, line 141: :+1: Good idea to add this as well

ethanfrey (approved):
Looks good. Please update the index test to be a bit more clear.

packages/storage-plus/src/indexed_map.rs
maurolacy: It is the primary key, yes. Maybe a good idea would be to show an example with this is empty (`b""`). Will add a couple comments / more examples.

PR #290, opened 2021-04-30
Pkowned to vec u8

### LINKED ISSUE #199: Storage-plus: See if we can simplify PkOwned to Vec<u8>

maurolacy:
Closes #199.

After some restructuring of the `PrimaryKey` and `Prefixer` trait impls, it was straightforward to replace `PkOwned` by `Vec<u8>`, and remove `PkOwned` entirely.

### DIFF
--- a/packages/storage-plus/src/keys.rs
+++ b/packages/storage-plus/src/keys.rs
@@ -40,7 +40,7 @@ impl<'a> PrimaryKey<'a> for &'a str {
     }
 }
 
-// use generics for combining there - so we can use &[u8], PkOwned, or IntKey
+// use generics for combining there - so we can use &[u8], Vec<u8>, or IntKey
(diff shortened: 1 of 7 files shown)


packages/storage-plus/src/keys.rs
```diff
@@ -154,20 +150,20 @@ impl<'a> Prefixer<'a> for &'a Addr {
     }
 }
 
-// this auto-implements PrimaryKey for all the IntKey types (and more!)
-impl<'a, T: AsRef<PkOwned> + From<PkOwned> + Clone> PrimaryKey<'a> for T {
+// this auto-implements PrimaryKey for all the IntKey types
+impl<'a, T: Endian + Clone> PrimaryKey<'a> for IntKey<T> {
```
ethanfrey, line 154: Better trait limit

ethanfrey (approved):
Good stuff. Should have done this months ago.

PR #291, opened 2021-04-30
Clarify index_key() range() vs prefix() behaviour

maurolacy:
Adds an extra data item to the index tests, to explicitly show how `range()` with a bound differs from `prefix()`.


ethanfrey (approved):
lgtm

PR #293, opened 2021-05-19
Expose contract components

orkunkl:
When importing/extending contracts, access to base contract's errors, msgs required. This PR will allow that.


ethanfrey (approved):
Thanks for exporting these.

PR #317, opened 2021-06-30
Update remaining contracts to cosmwasm 0.15, fix problems

uint:


orkunkl:
Linter is happier than ever now 👯

PR #318, opened 2021-06-30
Migrate contracts to 0.15.0

### LINKED ISSUE #298: Upgrade to CosmWasm 0.15
This should be out shortly, and we should update these contracts to show how to do so.

Smart Contracts: 

- [ ] cw1-subkeys
- [ ] cw1-whitelist
- [ ] cw1155-base
- [ ] cw20-atomic-swap
- [x] #302
- [x] #307
- [x] #309
- [ ] #311
- [x] #312
- [ ] cw3-fixed-multisig
- [ ] cw3-flex-multisig
- [ ] cw4-group
- [ ] cw4-stake
- [ ] cw721-base

orkunkl:
All contracts migrated to `0.15`.
New cw-plus version is `0.7.0`. 
After merge cut we can cut new release.

Closes #298


contracts/cw20-bonding/src/curves.rs
```diff
@@ -218,14 +218,14 @@ impl DecimalPlaces {
         }
     }
 
-    pub fn to_reserve(&self, reserve: Decimal) -> Uint128 {
+    pub fn to_reserve(self, reserve: Decimal) -> Uint128 {
```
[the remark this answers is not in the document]
uint: I'll let @orkunkl defend this one :D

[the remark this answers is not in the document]
uint: Just checked and clippy doesn't complain about `&self`.

I guess I didn't question the change because the type is `Copy`, feels fairly primitive and is small enough that (on 64-bit machines) there's no gain passing it by reference. Feels right somehow. But probably doesn't really matter.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c0dd7e30065ca2135e8bce172cdda8c0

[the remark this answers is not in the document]
orkunkl: Yes it was requested by clippy and applied it.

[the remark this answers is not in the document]
orkunkl: local clippy version: `0.1.53 (53cb7b09 2021-06-17)`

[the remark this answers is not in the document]
uint: Oh, good catch. `0.1.51` doesn't complain, but `0.1.53` does.

contracts/cw3-fixed-multisig/src/state.rs
```diff
@@ -20,7 +20,7 @@ pub struct Proposal {
     pub title: String,
     pub description: String,
     pub expires: Expiration,
-    pub msgs: Vec<CosmosMsg<Empty>>,
+    pub msgs: Vec<SubMsg<Empty>>,
```
webmaster128, line 23: Let's leave this as `CosmosMsg` as well. No need to store all the execution parameters.

contracts/cw3-flex-multisig/src/msg.rs
```diff
@@ -109,7 +109,7 @@ pub enum ExecuteMsg {
     Propose {
         title: String,
         description: String,
-        msgs: Vec<CosmosMsg<Empty>>,
+        msgs: Vec<SubMsg<Empty>>,
```
webmaster128, line 112: Here we can keep `CosmosMsg` too. It can be wrapped when the message is actually executed.

contracts/cw3-flex-multisig/src/state.rs
```diff
@@ -31,7 +31,7 @@ pub struct Proposal {
     pub description: String,
     pub start_height: u64,
     pub expires: Expiration,
-    pub msgs: Vec<CosmosMsg<Empty>>,
+    pub msgs: Vec<SubMsg<Empty>>,
```
webmaster128, line 34: here too

contracts/cw4-stake/src/contract.rs
```diff
@@ -36,10 +36,10 @@ pub fn instantiate(
     ADMIN.set(deps.branch(), maybe_addr(api, msg.admin)?)?;
 
     // min_bond is at least 1, so 0 stake -> non-membership
-    let min_bond = match msg.min_bond {
-        Uint128(0) => Uint128(1),
+    let min_bond = Uint128::new(match msg.min_bond.u128() {
+        0 => 1,
         v => v,
-    };
+    });
```
webmaster128, line 42: Can't we use `let min_bond = std::cmp::max(msg.min_bond, Uint128::new(1));` instead?

packages/cw3/src/query.rs
```diff
@@ -129,7 +129,7 @@ where
     pub id: u64,
     pub title: String,
     pub description: String,
-    pub msgs: Vec<CosmosMsg<T>>,
+    pub msgs: Vec<SubMsg<T>>,
```
webmaster128, line 132: This should be `CosmosMsg` as well since the proposal can simply be `CosmosMsg`

packages/multi-test/src/app.rs
```diff
@@ -26,7 +26,7 @@ where
     C: Clone + fmt::Debug + PartialEq + JsonSchema,
 {
     // TODO: allow T != Empty
-    pub messages: Vec<CosmosMsg<C>>,
+    pub messages: Vec<SubMsg<C>>,
```
webmaster128, line 29: This can probably be changed to `CosmosMsg` too. 

Also, the comment above is outdated. We now have support for the custom type `C` and the comment was forgotten in this commit: https://github.com/CosmWasm/cosmwasm-plus/commit/ca59a96d696d535f8692b2d599aa7041677eb0ce

webmaster128 (approved):

ethanfrey:
Thank you for handling this

PR #319, opened 2021-07-09
CW20-base helpers.ts: update CW20 struct to use SigningCosmWasmClient…

randomshinichi:
… from stargate


ethanfrey:
Thank you for the contribution.

It looks like @orkunkl took the useful parts of this and cherry-picked it into #248, which also included the oysternet update.

In any case, the helper file should now work.

PR #322, opened 2021-07-12
Add has to Path and Map

### LINKED ISSUE #261: Add existence helpers to cw-storage-plus
Inspired by https://github.com/CosmWasm/cosmwasm-plus/pull/251#discussion_r608160548

We should add a quick existence operator to `Map` / `IndexedMap` that doesn't parse the data. This could be used like load:

`let exists: bool = TOKENS.has(deps.storage, &token_id);`

We should also expose this on the `Path` helper (which would be how the above is implemented`

`TOKENS.key(k).exists(deps.storage);`

Note that neither must return errors (like `remove`) as neither parses data

ethanfrey:
Closes #261

### DIFF
--- a/contracts/cw1155-base/src/contract.rs
+++ b/contracts/cw1155-base/src/contract.rs
@@ -233,9 +233,9 @@ pub fn execute_mint(
     }
 
     // insert if not exist
-    let key = TOKENS.key(&token_id);
-    if deps.storage.get(&key).is_none() {
-        key.save(deps.storage, &"".to_owned())?;
+    if !TOKENS.has(deps.storage, &token_id) {
+        // we must save some valid data here
+        TOKENS.save(deps.storage, &token_id, &String::new())?;
     }
 
     Ok(rsp)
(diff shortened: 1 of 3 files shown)


contracts/cw1155-base/src/contract.rs
```diff
@@ -233,9 +233,9 @@ pub fn execute_mint(
     }
 
     // insert if not exist
-    let key = TOKENS.key(&token_id);
-    if deps.storage.get(&key).is_none() {
-        key.save(deps.storage, &"".to_owned())?;
+    if !TOKENS.has(deps.storage, &token_id) {
```
maurolacy, line 236: 👍🏼

maurolacy (approved):
lgtm.

contracts/cw1155-base/src/contract.rs
ethanfrey: When writing tests I found out the (untested) case would actually panic.
If you attempt to write `b""` to the db.

(Which we added to avoid getting confusing results from the cosmos sdk storage layer). So I fixed that with real data as well.

maurolacy: So, the original code had that flaw too?

maurolacy: Oh, I guess you mean, as a key, not as a value.

PR #323, opened 2021-07-12
Support submessages in multitest

### LINKED ISSUE #259: Support submessages in multitest
This is a collection of items to do (multiple PRs, can be broken into issues).
Follow up from #257 and #258

- [ ] Support `reply` method on contracts
- [ ] Process submessages if returned from contract execution

ethanfrey:
Closes #259 

- [x] Add optional `reply` to contract wrapper contructor
- [x] Add sample contract that handles submessages
- [x] Add (failing) submessage tests
- [x] Support submessages

### DIFF
--- a/packages/multi-test/src/app.rs
+++ b/packages/multi-test/src/app.rs
@@ -409,7 +409,7 @@ mod test {
     };
     use crate::SimpleBank;
     use cosmwasm_std::testing::MockStorage;
-    use cosmwasm_std::{attr, coin, coins};
+    use cosmwasm_std::{attr, coin, coins, Reply};
 
     fn mock_router() -> App {
         let env = mock_env();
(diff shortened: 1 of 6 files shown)


ethanfrey:
I think I finally finished this.

Sometimes it felt like brute forcing to get the levels of caching working and I am sure it could use cleanup.

Happy for a review, even if only stylistic changes.

ethanfrey:
Okay, with #328 I fixed how we handle attributes to make it compatible with the events as returned with the SDK. And simplified/corrected how we handle SubMsg/reply.

I feel better about the functionality, but still think I need to organize the code better... especially all those traits I added to unify *Router and *Cache

packages/multi-test/src/app.rs
```diff
@@ -269,65 +312,107 @@ where
...
+
+        // call reply if meaningful
+        if let Ok(r) = res {
+            if matches!(msg.reply_on, ReplyOn::Always | ReplyOn::Success) {
+                let mut orig = r.clone();
+                let reply = Reply {
+                    id: msg.id,
+                    result: ContractResult::Ok(SubMsgExecutionResponse {
+                        events: r.events,
+                        data: r.data,
```
YunSuk-Yeo, line 351: For the better reply test, maybe we can wrap the data with Protobuf wrapper like 

https://github.com/terraswap/terraswap/blob/55c9ac2e54e95462c5c2ed3507cbdc27993b1b64/contracts/terraswap_factory/src/response.proto#L4-L9

https://github.com/CosmWasm/wasmd/blob/8bec411e4bd947cc7d872bd14bd1da865bf703bb/proto/cosmwasm/wasm/v1beta1/tx.proto#L69-L74

Then, we can properly test Protobuf decoding part at reply like https://github.com/terraswap/terraswap/blob/55c9ac2e54e95462c5c2ed3507cbdc27993b1b64/contracts/terraswap_factory/src/contract.rs#L152-L155

YunSuk-Yeo (approved):
LGTM 👏 👏 👏 

I think Protobuf style data decoding testing at `reply` can be done later.

Thanks for making good test-kit for all CosmWasm contracts.

packages/multi-test/src/app.rs
```diff
@@ -269,65 +312,107 @@ where
...
+        let res = self.wasm.reply(contract.clone(), self.querier, reply)?;
+        self.process_response(contract, res)
+    }
+
+    fn process_response(
+        &mut self,
+        contract: Addr,
+        response: Response<C>,
+    ) -> Result<AppResponse, String> {
+        let mut events = response.events;
```
YunSuk-Yeo, line 392: For the custom events, wasmd put `wasm-` prefix, so it seems better to put `wasm-` prefix to all custom events types.

https://github.com/CosmWasm/wasmd/blob/39a967ea38d70d84f6da4e7e3a72ba89b562f525/x/wasm/keeper/events.go#L43

packages/multi-test/src/wasm.rs
```diff
@@ -410,31 +534,64 @@ pub struct WasmOps {
...
-            state: WasmCacheState {
-                contracts: HashMap::new(),
-                contract_diffs: HashMap::new(),
-            },
+            parent_contracts: router,
+            state: WasmCacheState::new(),
+        }
+    }
+
+    pub fn cache(&self) -> WasmCache<C> {
```
maurolacy, line 590: I started reviewing this file, and this is a little off the top of my head at this point, but:

I _guess_ we need a multi-level (contract state) cache, to separate "server" contracts state from "client" contracts states. This, to mimic the transactional nature of the different contracts in the SDK.

Impressive piece of work, by the way.

packages/multi-test/src/wasm.rs
```diff
@@ -519,20 +674,45 @@ where
...
+        let env = self.router.get_env(address.clone());
+        let api = self.router.api.as_ref();
+
+        self.state.with_storage(
+            querier,
+            self.parent_contracts,
+            address,
+            env,
+            api,
+            |code_id, deps, env| {
```
maurolacy, line 694: Nice "generic executor" trick.

maurolacy (commented):
Partial review of `wasm.rs`. lgtm.

Will continue reviewing later, as I have some errands to do now.

packages/multi-test/src/app.rs
ethanfrey: Nice catch. I did something in `init_response`, but that was the old (pre-protobuf) version. Thanks for pointing this out.

packages/multi-test/src/app.rs
ethanfrey: Very good point here

packages/multi-test/src/wasm.rs
ethanfrey: Before this PR we had a one-level cache in `App.execute_multi`. That is, all the messages are run together in a cache, and any error will cause the state changes to rollback. They will only be committed to the "real" storage if all return a success.

With submessages, each submessage will need it's own cache. They can write state and then fail, and then reply could catch the error and turn it to a success. That would mean the "parent contract" writes should be committed but the submessage writes not. These semanatics are described here: https://github.com/CosmWasm/cosmwasm/blob/main/SEMANTICS.md#submessages

This leads to the need of an arbitrarily multi-level caching technique. I feel my approach in this PR is a bit brute force and could probably use a bigger refactor to only use one Storage (rather than multiple HashMaps), and unify a lot on the caching in the end, but that would be much more work than needed to get a 0.7.0 release out. This package really should get some more refactors when we have capacity.

packages/multi-test/src/wasm.rs
ethanfrey: thanks

packages/multi-test/src/app.rs
ethanfrey: Added #330 for this

ethanfrey: And closing with #331

PR #332, opened 2021-07-14
Update to cw schema 0.15.0

maurolacy:
This is irrelevant, as there are no changes between `cosmwasm-schema` 0.14 and 0.15. But anyway, for completeness.


ethanfrey:
Good catch

PR #334, opened 2021-07-16
Contract builders

ethanfrey:
Simplify creating contracts that use sudo or reply, allowing us to easily add more methods in the future.


packages/multi-test/src/contracts.rs
```diff
@@ -0,0 +1,267 @@
...
+        T4A: DeserializeOwned + 'static,
+        E4A: ToString + 'static,
+    {
+        ContractWrapper {
+            execute_fn: self.execute_fn,
+            instantiate_fn: self.instantiate_fn,
+            query_fn: self.query_fn,
+            sudo_fn: Some(Box::new(sudo_fn)),
+            reply_fn: self.reply_fn,
+        }
```
maurolacy, line 138: A trick I've seen somewhere:

```suggestion
        let Self {
            execute_fn,
            instantiate_fn,
            query_fn,
            reply_fn,
            ..
        } = self;
        ContractWrapper {
            execute_fn,
            instantiate_fn,
            query_fn,
            sudo_fn: Some(Box::new(sudo_fn)),
            reply_fn,
        }
```
Not sure what's better, as it's longer / more verbose. But it's neat nevertheless.

packages/multi-test/src/contracts.rs
```diff
@@ -0,0 +1,267 @@
...
+    E2: ToString + 'static,
+    E3: ToString + 'static,
+    E4: ToString + 'static,
+    E5: ToString,
+    C: Clone + fmt::Debug + PartialEq + JsonSchema + 'static,
+{
+    pub fn with_sudo<T4A, E4A>(
+        self,
+        sudo_fn: SudoFn<T4A, C, E4A>,
+    ) -> ContractWrapper<T1, T2, T3, E1, E2, E3, C, T4A, E4A, E5>
```
maurolacy, line 127: I guess there's no way around this, because of the defaults. Perhaps without specifying the defaults? They are used to specify a unset `Option` in the end.

maurolacy (approved):
lgtm.

packages/multi-test/src/contracts.rs
ethanfrey: Way around what?

Changing T4, E4 -> T4A, E4A?

I was quite happy with this discovery, as all attempts to use `&mut self` and change the struct failed (as mis-matched types), but the builder allows us to customise the types as we wish.

packages/multi-test/src/contracts.rs
ethanfrey: I've seen that. It is quite neat, and I have used it a few places. In this case it is a fair bit longer and we only deref `self`, not some larger construct

ethanfrey:
Thanks for the review

PR #335, opened 2021-07-19
Simplify multitest storage

ethanfrey:
merge after #334 

Major overhaul of how storage and caching works in multi-test.
This should allow for much more extensibility in multi-test in the future

PR #337, opened 2021-07-20
Multitest bank events

### LINKED ISSUE #329: multi-test: Bank returns realistic events
It should return a `transfer` event with a few attributes: https://github.com/cosmos/cosmos-sdk/blob/84c33215658131d87daf3c629e909e12ed9370fa/x/bank/keeper/send.go#L142-L147

Actually, maybe the `message` one as well, although we strip that off when it is called from a wasm contract.

Let's go and make this as close to x/wasm behavior as possible, especially as some `reply` blocks may look at the events.

ethanfrey:
Closes #329 

Merge after #335 

Add the standard cosmos-sdk events to the bank execute

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -213,6 +213,7 @@ dependencies = [
  "cosmwasm-storage",
  "cw-storage-plus",
  "cw0",
+ "itertools",
  "prost",
  "schemars",
  "serde",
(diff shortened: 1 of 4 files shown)


maurolacy (commented):
lgtm.

PR #338, opened 2021-07-20
Cosmwasm 0.16

### LINKED ISSUE #333: Upgrade to cosmwasm 0.16
Note: storage-plus should enable iterator feature by default for consistency with cosmwasm-std and cosmwasm-storage

Most changes are around the IBC APIs

uint:
Just your run-of-the-mill cosmwasm update.

We should probably hold off with merging this until we have a full release rather than an RC.

Closes #333


webmaster128:
Nice.

> We should probably hold off with merging this until we have a full release rather than an RC.

Agreed.

ethanfrey:
It's good to have with an rc to test.

I'd love a review on #335 and #337 if you have time

ethanfrey:
Can you rebase this and update to cosmwasm 0.16.0-rc4.

That is integrated in wasmd and close enough to final. Happy to have this merged in, so there are no more merge conflicts moving forward

ethanfrey:
Besides the missing rebase (see conflicts), I assume this is ready to merge?

Can you rebase and pull it out of draft mode and I will give a final review

.circleci/config.yml
```diff
@@ -907,7 +907,7 @@ jobs:
       - run:
           name: Install check_contract
           # Uses --debug for compilation speed
-          command: cargo install --debug --version 0.15.0 --features iterator --example check_contract -- cosmwasm-vm
+          command: cargo install --debug --version 0.16.0-rc2 --features iterator --example check_contract -- cosmwasm-vm
```
ethanfrey, line 910: rc4 here as well

contracts/cw20-ics20/src/ibc.rs
```diff
@@ -170,8 +178,9 @@ pub fn ibc_packet_receive(
             attributes: vec![
                 attr("action", "receive"),
                 attr("success", "false"),
-                attr("error", err),
+                attr("error", err.to_string()),
```
ethanfrey, line 181: I know you cannot implement attr for both `impl Into<String>` and `impl ToString` as some structs implement both and you get a compiler error there.

Is there a way to say `K: ToString & !Into<String>` ? I think I looked into this some months ago and found it worked in unstable rust (and used to optimise std lib Vec for example)

ethanfrey (commented):
Looks good. Two minor comments.

One push for getting this done pre-0.16.0 final is making sure we shape the API with this feedback. Also, some other crates depend on cosmwasm-plus, so if we could make a 0.8.0-rc1, we can get more feedback of use cases when updating those crates.

Usage is the best way to polish off the API.

uint:
Rebased and migrated to `cosmwasm 0.16.0-rc5`

contracts/cw20-ics20/src/ibc.rs
uint: Ayep. It's probably just something we have to live with.

uint:
I just want to point out we're removing 250 lines of code in this PR! Ha.

.circleci/config.yml
uint: updated now!

webmaster128:
🔥

ethanfrey:
and the target branch keeps moving....

let's pull this out of draft and rebase again.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -119,11 +117,9 @@ where
...
     }
     // Relay messages
-    let res = Response {
-        messages: msgs,
-        attributes: vec![attr("action", "execute"), attr("owner", info.sender)],
-        ..Response::default()
-    };
+    let res = Response::new()
+        .add_messages(msgs)
+        .add_attributes(vec![attr("action", "execute"), attr("owner", info.sender)]);
```
ethanfrey, line 122: wouldn't that be simpler with 2 add_attribute calls, since they are different V.

Not to be a stickler, but these contracts are like an example for other devs to follow

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -202,16 +198,13 @@ where
...
-            attr("amount", amount.amount),
-        ],
-        ..Response::default()
-    };
+    let res = Response::new().add_attributes(vec![
+        attr("action", "increase_allowance"),
+        attr("owner", info.sender),
+        attr("spender", spender),
+        attr("denomination", amount.denom),
+        attr("amount", amount.amount),
```
ethanfrey, line 206: same here, I'd only use add_attributes when you don't need to use attr

ethanfrey (commented):
I'd prefer more idiomatic add_attribute usage, even if it does take a bit more time. These are reference contracts

contracts/cw1-subkeys/src/contract.rs
uint: Done almost everywhere now. We barely ever use `attr`.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -271,13 +271,12 @@ where
...
     }
 
-    let res = Response::new().add_attributes(vec![
-        attr("action", "decrease_allowance"),
-        attr("owner", info.sender),
-        attr("spender", spender),
-        attr("denomination", amount.denom),
-        attr("amount", amount.amount),
-    ]);
+    let res = Response::new()
```
ethanfrey, line 274: :rocket:

contracts/cw20-atomic-swap/src/contract.rs
```diff
@@ -683,7 +682,7 @@ mod tests {
         )
         .unwrap();
         assert_eq!(0, res.messages.len());
-        assert_eq!(attr("action", "create"), res.attributes[0]);
+        assert_eq!(("action", "create"), res.attributes[0]);
```
ethanfrey, line 685: Good fix as well

contracts/cw3-flex-multisig/src/contract.rs
```diff
@@ -1111,10 +1108,10 @@ mod tests {
             .unwrap();
         assert_eq!(
             res.custom_attrs(0),
-            &[
-                attr("action", "execute"),
-                attr("sender", SOMEBODY),
-                attr("proposal_id", proposal_id.to_string()),
+            [
```
ethanfrey, line 1111: Wow, this works over arrays as well? Good stuff.

ethanfrey (approved):
Awesome!

contracts/cw3-flex-multisig/src/contract.rs
uint: Oh, yeah. Sometimes if you implement a trait like `PartialEq` for type `T`, it turns out some generic implementation for `[T; 3]`or `Vec<T>` or something kicks in from Rust's `std` and cool things happen.

Issue #327, opened 2021-07-13
storage-plus: Need better docs and examples for IndexedMap


orkunkl:
I am testing out `IndexedMap` as workshop and article content.
I found `Index` and prefixes under documented and not enough tests to understand as a noobie.

For example here is a sample code I written to show an **CompositeKey** example.

Tokens are secondary indexed by (admin, ticker) composite key.

```rust

pub const TOKEN_COUNT: Item<u64> = Item::new("num_tokens");

pub fn num_tokens(storage: &dyn Storage) -> StdResult<u64> {
    Ok(TOKEN_COUNT.may_load(storage)?.unwrap_or_default())
}

pub fn increment_tokens(storage: &mut dyn Storage) -> StdResult<u64> {
    let val = num_tokens(storage)? + 1;
    TOKEN_COUNT.save(storage, &val)?;
    Ok(val)
}

#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
pub struct Token {
    pub admin: Addr,
    pub ticker: String
}

pub struct TokenIndexes<'a> {
    // indexed by admin_ticker composite key
    // last U64Key is the primary key which is auto incremented token counter
    pub admin_ticker: MultiIndex<'a, (Vec<u8>, Vec<u8>, Vec<u8>), Token>,
}

// this may become macro, not important just boilerplate, builds the list of indexes for later use
impl<'a> IndexList<Token> for TokenIndexes<'a> {
    fn get_indexes(&'_ self) -> Box<dyn Iterator<Item = &'_ dyn Index<Token>> + '_> {
        let v: Vec<&dyn Index<Token>> = vec![&self.admin_ticker];
        Box::new(v.into_iter())
    }
}

const TOKEN_NAMESPACE: &str = "tokens";

pub fn tokens<'a>() -> IndexedMap<'a, &'a [u8], Token, TokenIndexes<'a>> {
    let indexes = TokenIndexes {
        admin_ticker: MultiIndex::new(
            |d, k| (index_string(d.admin.as_ref()), index_string(&d.ticker), k),
            TOKEN_NAMESPACE,
            "tokens__admin_ticker",
        )
    };
    IndexedMap::new(TOKEN_NAMESPACE, indexes)
}
```

Here are some tests to show what I want to achieve.
```rust
    #[test]
    fn test_tokens() {
        let mut store = MockStorage::new();

        let admin1 = Addr::unchecked("addr1");
[...]

PR #340, opened 2021-07-21
Indexed map docs

maurolacy:
Continuation / closure of #327.

### DIFF
--- a/contracts/cw721-base/src/contract.rs
+++ b/contracts/cw721-base/src/contract.rs
@@ -513,16 +513,18 @@ fn query_tokens(
     let res: Result<Vec<_>, _> = tokens()
         .idx
         .owner
-        .prefix(Vec::from(owner_addr.as_ref()))
-        .range(deps.storage, start, None, Order::Ascending)
+        .pks(
+            deps.storage,
+            Vec::from(owner_addr.as_ref()),
+            start,
+            None,
+            Order::Ascending,
+        )
         .take(limit)
         .collect();
-    let tokens = res?;
+    let pks = res?;
 
-    let res: Result<Vec<_>, _> = tokens
-        .iter()
-        .map(|v| String::from_utf8(v.0.clone()))
-        .collect();
+    let res: Result<Vec<_>, _> = pks.iter().map(|v| String::from_utf8(v.to_vec())).collect();
     let tokens = res.map_err(StdError::invalid_utf8)?;
     Ok(TokensResponse { tokens })
 }
--- a/contracts/cw721-base/src/contract.rs
+++ b/contracts/cw721-base/src/contract.rs
@@ -507,7 +507,7 @@ fn query_tokens(
     limit: Option<u32>,
 ) -> StdResult<TokensResponse> {
     let limit = limit.unwrap_or(DEFAULT_LIMIT).min(MAX_LIMIT) as usize;
-    let start = start_after.map(|token_id| Bound::exclusive(token_id));
+    let start = start_after.map(Bound::exclusive);
 
     let owner_addr = deps.api.addr_validate(&owner)?;
     let res: Result<Vec<_>, _> = tokens()
--- a/contracts/cw721-base/src/contract.rs
+++ b/contracts/cw721-base/src/contract.rs
@@ -507,25 +507,22 @@ fn query_tokens(
     limit: Option<u32>,
 ) -> StdResult<TokensResponse> {
     let limit = limit.unwrap_or(DEFAULT_LIMIT).min(MAX_LIMIT) as usize;
-    let start_addr = maybe_addr(deps.api, start_after)?;
-    let start = start_addr.map(|addr| Bound::exclusive(addr.as_ref()));
+    let start = start_after.map(|token_id| Bound::exclusive(token_id));
 
     let owner_addr = deps.api.addr_validate(&owner)?;
     let res: Result<Vec<_>, _> = tokens()
         .idx
         .owner
-        .pks(
-            deps.storage,
-            Vec::from(owner_addr.as_ref()),
-            start,
-            None,
-            Order::Ascending,
-        )
+        .prefix(Vec::from(owner_addr.as_ref()))
+        .range(deps.storage, start, None, Order::Ascending)
         .take(limit)
         .collect();
-    let pks = res?;
+    let tokens = res?;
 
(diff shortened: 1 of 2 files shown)


maurolacy:
I'm now thinking that this can be improved by using `&Addr` as index key. Which will force me to fix all the docs. :-)

packages/storage-plus/README.md
```diff
@@ -384,7 +384,172 @@ fn demo() -> StdResult<()> {
...
+
+### Definition
+
+```rust
+pub struct TokenIndexes<'a> {
+  // pk goes to second tuple element
+  pub owner: MultiIndex<'a, (Vec<u8>, Vec<u8>), TokenInfo>,
+}
+
+impl<'a> IndexList<TokenInfo> for TokenIndexes<'a> {
```
ethanfrey, line 400: This is rather annoyingly complex boilerplate.
I would consider a macro to generate this (one day)

packages/storage-plus/README.md
```diff
@@ -384,7 +384,172 @@ fn demo() -> StdResult<()> {
...
+```rust
+impl<'a> IndexList<TokenInfo> for TokenIndexes<'a> {
+    fn get_indexes(&'_ self) -> Box<dyn Iterator<Item = &'_ dyn Index<TokenInfo>> + '_> {
+        let v: Vec<&dyn Index<TokenInfo>> = vec![&self.owner];
+        Box::new(v.into_iter())
+    }
+}
+```
+
+This implements the `IndexList` trait for `TokenIndexes`. This kind of serves two purposes (which are really
```
ethanfrey, line 489: You can mention this is more or less boiler-plate and needed for the internals. Do not try to customize this, just return a list of all indexes

packages/storage-plus/README.md
```diff
@@ -384,7 +384,172 @@ fn demo() -> StdResult<()> {
...
+    IndexedMap::new("tokens", indexes)
+}
+```
+
+Here `tokens()` is just a helper function, that simplifies the `IndexedMap` construction for us. First the
+index (es) is (are) created, and then, the `IndexedMap` is created (using `IndexedMap::new`), and returned.
+
+During index creation, we must supply an index function per index
+```rust
+        owner: MultiIndex::new(
```
ethanfrey, line 511: Nice to highlight the essential line

ethanfrey: I agree that if we could use &Addr here or something, this would be nice.
But I think there are lifetime issues there (like the &str issue).

I would merge the docs as are and update if you get &str working

packages/storage-plus/README.md
```diff
@@ -384,7 +384,172 @@ fn demo() -> StdResult<()> {
...
+for the pk, with the defined indexes.
+
+So, `IndexedMap` (and the other `Indexed*` types) is just a wrapper / extension around `Map`, that provides
+a number of index functions and namespaces to create indexes over the original `Map` data. It also implements
+calling these index functions during value storage / update / removal, so that you can forget about it,
+and just use the indexed data.
+
+### Usage
+
+An example of use, where `owner` is a `String` value passed as a parameter, and `start_after` and `limit` optionally
```
ethanfrey, line 539: Can you add a simpler example first...

`tokens().idx.owner.get(...)` to show how we can query by owner?
Or is that not possible, as we get a list... and we need to use this?

If so, it would be good to explain the `prefix` before the `range`

contracts/cw721-base/src/contract.rs
```diff
@@ -507,25 +507,22 @@ fn query_tokens(
...
     let limit = limit.unwrap_or(DEFAULT_LIMIT).min(MAX_LIMIT) as usize;
-    let start_addr = maybe_addr(deps.api, start_after)?;
-    let start = start_addr.map(|addr| Bound::exclusive(addr.as_ref()));
+    let start = start_after.map(Bound::exclusive);
 
     let owner_addr = deps.api.addr_validate(&owner)?;
     let res: Result<Vec<_>, _> = tokens()
         .idx
         .owner
-        .pks(
```
ethanfrey, line 517: This old code, just read the pks.

The new code will load all the token data for each of them (2x reads)

contracts/cw721-base/src/contract.rs
```diff
@@ -507,25 +507,22 @@ fn query_tokens(
...
+        .range(deps.storage, start, None, Order::Ascending)
         .take(limit)
         .collect();
-    let pks = res?;
+    let tokens = res?;
 
-    let res: Result<Vec<_>, _> = pks.iter().map(|v| String::from_utf8(v.to_vec())).collect();
+    let res: Result<Vec<_>, _> = tokens
+        .iter()
+        .map(|v| String::from_utf8(v.0.clone()))
```
ethanfrey, line 524: And we throw away the data here, just using the pk.

I understand why you show `range` for the example, but since we only want IDs, it is rather inefficient to do it this way
(maybe show both alternatives in the doc and not modify the code?)

ethanfrey (commented):
Nice overview.

I added a few comments for improvements.

I would love it if @orkunkl could try reading it and give his feedback (as he has not been working on coding these types) to ensure we are not missing something obvious

contracts/cw721-base/src/contract.rs
maurolacy: No problem. I Just wanted to document  some actual code, yes.

maurolacy: I think a good idea would be to move `pks()` to `Prefix`. So that `prefix()` can be used with `pks()` the same way we use it with `range()`. And then we write a `pks()` helper in the base map for the empty prefix case.

packages/storage-plus/README.md
maurolacy: There's no `get`, as you would have to return a list or iterator.

I can explain `prefix()` / `sub_prefix()` and then show both `range()` and the current `pks()`.

packages/storage-plus/README.md
```diff
@@ -538,9 +541,12 @@ and just use the indexed data.
 
 An example of use, where `owner` is a `String` value passed as a parameter, and `start_after` and `limit` optionally
 define the pagination range:
+
+Notice this uses `prefix()`, explained above in the `Map` section.
+
 ```rust
-    let start = start_after.map(|token_id| Bound::exclusive(token_id));
+    let start = start_after.map(Bound::exclusive);
```
ethanfrey, line 549: nice one

ethanfrey (approved):
Good stuff. Thanks for the updates

PR #342, opened 2021-07-22
Cw721 token indexes refactor

maurolacy:
An addendum to #327 / #340, improving the cw721 tokens by owner index, and updating the docs.

### DIFF
--- a/contracts/cw721-base/src/state.rs
+++ b/contracts/cw721-base/src/state.rs
@@ -65,7 +65,11 @@ impl<'a> IndexList<TokenInfo> for TokenIndexes<'a> {
 
 pub fn tokens<'a>() -> IndexedMap<'a, &'a str, TokenInfo, TokenIndexes<'a>> {
     let indexes = TokenIndexes {
-        owner: MultiIndex::new(|d, k| (d.owner.clone(), k), "tokens", "tokens__owner"),
+        owner: MultiIndex::new(
+            |d: &TokenInfo, k: Vec<u8>| (d.owner.clone(), k),
+            "tokens",
+            "tokens__owner",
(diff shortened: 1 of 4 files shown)


contracts/cw721-base/src/state.rs
```diff
@@ -65,11 +65,7 @@ impl<'a> IndexList<TokenInfo> for TokenIndexes<'a> {
 
 pub fn tokens<'a>() -> IndexedMap<'a, &'a str, TokenInfo, TokenIndexes<'a>> {
     let indexes = TokenIndexes {
-        owner: MultiIndex::new(
-            |d, k| (Vec::from(d.owner.as_ref()), k),
-            "tokens",
-            "tokens__owner",
-        ),
+        owner: MultiIndex::new(|d, k| (d.owner.clone(), k), "tokens", "tokens__owner"),
```
ethanfrey, line 68: Do you need clone here? Isn't it `|d: TokenInfo, k: Addr|`? Or is it `|d: &TokenInfo, k: Addr|`?

Inlining the proper types would definitely make this more readable if we use it as an learning example

ethanfrey (approved):
In general looks good. And by using `Addr` rather than `&Addr`, you avoid the lifetime issue.

Reviewing it, I realized some of the implied types are confusing and would be great to make explicit. Otherwise, good to merge

contracts/cw721-base/src/state.rs
maurolacy: It's the second one. That's why `clone` is needed.

Will do the type inlining.

ethanfrey (approved):
Good stuff, makes it a lot clearer.

PR #343, opened 2021-07-22
Refactor of cw1-whitelist unit tests

hashedone:
As there is a request to create additional UTs for some contracts: https://github.com/CosmWasm/cosmwasm-plus/issues/105, I went through the whitelist contract UT and found that kind of problem with adding many tests there is repetitive boilerplate in every test, which is not actually easy to read. Also because of some edge cases, some test may be actually incorrect (eg. if in future some items would be returned in different order because of any reason, UTs would start failing - while ordering in many cases shouldn't matter).

Therefore I came up with proposal to kind of unify the UTs structure to have visible setup => execute => verify structure, with - what is important - clear setup, so it is easier to figure out what is the test about.

Additional changes I am considering to make execution and possibly verify part simpler is to add `execute`/`query` methods directly on `Suite` (query via router), so the `deps` and `mock_env` are hidden. Additionally it would make testing queries through whole ecosystem, not just locally which might be more complete.

At the end I think, that the `Suite` structure could be partially generalized and extracted to `multi-test`, to encourage having clean an uniform test through the smart contracts.

### DIFF
--- a/contracts/cw1-subkeys/src/contract.rs
+++ b/contracts/cw1-subkeys/src/contract.rs
@@ -84,11 +84,11 @@ where
     T: Clone + fmt::Debug + PartialEq + JsonSchema,
 {
     // Wrap `msgs` in SubMsg.
-    let msgs: Vec<SubMsg<_>> = msgs.into_iter().map(SubMsg::new).collect();
+    let msgs: Vec<_> = msgs.into_iter().map(SubMsg::new).collect();
     let cfg = ADMIN_LIST.load(deps.storage)?;
-    // This is the admin behavior (same as cw1-whitelist)
-    if cfg.is_admin(info.sender.as_ref()) { // Just relay the messages
-    } else {
+
+    // Not an admin - need to check for permissions
+    if !cfg.is_admin(info.sender.as_ref()) {
         for msg in &msgs {
             match &msg.msg {
                 CosmosMsg::Staking(staking_msg) => {
@@ -130,7 +130,7 @@ where
 pub fn check_staking_permissions(
     staking_msg: &StakingMsg,
     permissions: Permissions,
-) -> Result<bool, ContractError> {
+) -> Result<(), ContractError> {
     match staking_msg {
         StakingMsg::Delegate { .. } => {
             if !permissions.delegate {
@@ -149,13 +149,13 @@ pub fn check_staking_permissions(
         }
         s => panic!("Unsupported staking message: {:?}", s),
     }
-    Ok(true)
+    Ok(())
 }
 
 pub fn check_distribution_permissions(
     distribution_msg: &DistributionMsg,
     permissions: Permissions,
-) -> Result<bool, ContractError> {
+) -> Result<(), ContractError> {
     match distribution_msg {
         DistributionMsg::SetWithdrawAddress { .. } => {
             if !permissions.withdraw {
@@ -169,7 +169,7 @@ pub fn check_distribution_permissions(
         }
         s => panic!("Unsupported distribution message: {:?}", s),
     }
-    Ok(true)
+    Ok(())
 }
 
 pub fn execute_increase_allowance<T>(
@@ -425,8 +425,10 @@ pub fn query_all_permissions(
 
 #[cfg(test)]
 mod tests {
-    use cosmwasm_std::testing::{mock_dependencies, mock_env, mock_info};
-    use cosmwasm_std::{coin, coins, Addr, StakingMsg, Timestamp};
+    use cosmwasm_std::testing::{
+        mock_dependencies, mock_env, mock_info, MockApi, MockQuerier, MockStorage,
+    };
+    use cosmwasm_std::{coin, coins, Addr, OwnedDeps, StakingMsg, Timestamp};
 
     use cw0::NativeBalance;
     use cw1_whitelist::msg::AdminListResponse;
@@ -435,6 +437,138 @@ mod tests {
     use crate::state::Permissions;
 
     use super::*;
+
+    const OWNER: &str = "owner";
+
+    const ADMIN1: &str = "admin1";
+    const ADMIN2: &str = "admin2";
+
+    const SPENDER1: &str = "spender1";
+    const SPENDER2: &str = "spender2";
+    const SPENDER3: &str = "spender3";
+    const SPENDER4: &str = "spender4";
+
+    const TOKEN: &str = "token";
+    const TOKEN1: &str = "token1";
+    const TOKEN2: &str = "token2";
+    const TOKEN3: &str = "token3";
+
+    const ALL_PERMS: Permissions = Permissions {
+        delegate: true,
+        redelegate: true,
+        undelegate: true,
+        withdraw: true,
+    };
+    const NO_PERMS: Permissions = Permissions {
+        delegate: false,
+        redelegate: false,
+        undelegate: false,
+        withdraw: false,
+    };
+
+    /// Helper structure for Suite configuration
+    #[derive(Default)]
+    struct SuiteConfig {
+        spenders: Vec<Spender>,
+        admins: Vec<&'static str>,
+    }
+
+    impl SuiteConfig {
+        fn init(self) -> Suite {
+            Suite::init_with_config(self)
+        }
+    }
+
+    #[derive(Default)]
+    struct Spender {
+        spender: &'static str,
+        allowances: Vec<Coin>,
+        allowances_expire: Option<Expiration>,
+        permissions: Option<Permissions>,
+    }
+
+    /// Test suite helper unifying test initialization, keeping access to created data
+    struct Suite {
+        deps: OwnedDeps<MockStorage, MockApi, MockQuerier>,
+        owner_info: MessageInfo,
+    }
+
+    impl Suite {
+        /// Initializes test case using default config
+        fn init() -> Self {
+            Self::init_with_config(SuiteConfig::default())
+        }
+
+        /// Initialized test case using provided config
+        fn init_with_config(config: SuiteConfig) -> Self {
+            let mut deps = mock_dependencies(&[]);
+            let admins = std::iter::once(OWNER)
+                .chain(config.admins)
+                .map(ToOwned::to_owned)
+                .collect();
+
+            let instantiate_msg = InstantiateMsg {
+                admins,
+                mutable: true,
+            };
+            let owner_info = mock_info(OWNER, &[]);
+
+            instantiate(
+                deps.as_mut().branch(),
+                mock_env(),
+                owner_info.clone(),
+                instantiate_msg,
+            )
+            .unwrap();
+
+            for Spender {
+                spender,
+                allowances,
+                allowances_expire: expires,
+                permissions,
+            } in config.spenders
+            {
+                for amount in allowances {
+                    let msg = ExecuteMsg::IncreaseAllowance {
+                        spender: spender.to_owned(),
+                        amount,
+                        expires,
+                    };
+                    execute(deps.as_mut().branch(), mock_env(), owner_info.clone(), msg).unwrap();
+                }
+
+                if let Some(permissions) = permissions {
+                    let msg = ExecuteMsg::SetPermissions {
+                        spender: spender.to_owned(),
+                        permissions,
+                    };
+                    execute(deps.as_mut().branch(), mock_env(), owner_info.clone(), msg).unwrap();
+                }
+            }
+
+            Self { deps, owner_info }
+        }
+    }
+
+    /// Helper function for comparing vectors or another slice-like object as they would represent
+    /// set with duplications. Compares sets by first sorting elements usning provided ordering.
+    /// This functions reshufless elements inplace, as it should never matter as compared
+    /// containers should represent same value regardless of ordering, and making this inplace just
+    /// safes obsolete copying.
+    ///
+    /// This is implemented as a macro instead of function to throw panic in the place of macro
+    /// usage instead of from function called inside test.
+    macro_rules! assert_sorted_eq {
+        ($left:expr, $right:expr, $cmp:expr $(,)?) => {
+            let mut left = $left;
+            left.sort_by(&$cmp);
+
+            let mut right = $right;
+            right.sort_by($cmp);
+
+            assert_eq!(left, right);
+        };
+    }
 
     // this will set up instantiation for other tests
     fn setup_test_case(
@@ -467,34 +601,7 @@ mod tests {
 
     #[test]
     fn get_contract_version_works() {
-        let mut deps = mock_dependencies(&[]);
-
-        let owner = "admin0001";
-        let admins = vec![owner, "admin0002"];
-
-        let spender1 = "spender0001";
-        let spender2 = "spender0002";
-        let initial_spenders = vec![spender1, spender2];
-
-        // Same allowances for all spenders, for simplicity
-        let denom1 = "token1";
-        let amount1 = 1111;
-
-        let allow1 = coin(amount1, denom1);
-        let initial_allowances = vec![allow1];
-
-        let expires_never = Expiration::Never {};
-        let initial_expirations = vec![expires_never, expires_never];
-
-        let info = mock_info(owner, &[]);
-        setup_test_case(
-            deps.as_mut(),
-            &info,
-            &admins,
-            &initial_spenders,
-            &initial_allowances,
-            &initial_expirations,
-        );
+        let Suite { deps, .. } = Suite::init();
 
         assert_eq!(
             ContractVersion {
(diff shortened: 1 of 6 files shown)


contracts/cw1-subkeys/src/contract.rs
```diff
@@ -507,477 +614,388 @@ mod tests {
...
-        let info = mock_info(owner, &[]);
-        setup_test_case(
-            deps.as_mut(),
-            &info,
-            &admins,
-            &initial_spenders,
-            &initial_allowances,
-            &initial_expirations,
-        );
+        let Suite { deps, .. } = SuiteConfig {
```
ethanfrey, line 617: I see the idea of the Suite to handle many things more clearly.
Somehow the usage seems more verbose than the older setup code.

I wonder if it is possible to make usage more compact. With eg. 2 helpers:

`Spender::with_allowance` and `Spender::with_expiring_allowance`. That would reduce those 4-5 line struct inits into 1 line, and then the usage of the new Suite would really shine

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -507,477 +614,388 @@ mod tests {
...
-                balance: NativeBalance(initial_allowances),
-                expires: expires_later,
-            }
+                spender: SPENDER3.to_owned(),
+                balance: NativeBalance(vec![coin(3456, TOKEN)]),
+                expires: Expiration::AtHeight(12345),
+            },
+        ];
+
+        assert_sorted_eq!(
```
ethanfrey, line 718: I like the check here to sort before compare.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -436,6 +438,138 @@ mod tests {
...
+
+    /// Helper function for comparing vectors or another slice-like object as they would represent
+    /// set with duplications. Compares sets by first sorting elements usning provided ordering.
+    /// This functions reshufless elements inplace, as it should never matter as compared
+    /// containers should represent same value regardless of ordering, and making this inplace just
+    /// safes obsolete copying.
+    ///
+    /// This is implemented as a macro instead of function to throw panic in the place of macro
+    /// usage instead of from function called inside test.
+    macro_rules! assert_sorted_eq {
```
ethanfrey, line 561: Interesting macro trick.

I write a number of helper functions and always re-run with `RUST_BACKTRACE=1` when they fail to find the real location. This does seem more ergonomic.

We need to reconsider the macro-resistance in this codebase.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -507,477 +614,388 @@ mod tests {
...
-        let info = mock_info(owner, &[]);
-        setup_test_case(
-            deps.as_mut(),
-            &info,
-            &admins,
-            &initial_spenders,
-            &initial_allowances,
-            &initial_expirations,
-        );
+        // let's try pagination.
```
ethanfrey, line 683: Nice comment

contracts/cw1-whitelist/src/msg.rs
```diff
@@ -46,3 +46,35 @@ pub struct AdminListResponse {
     pub admins: Vec<String>,
     pub mutable: bool,
 }
+
+impl AdminListResponse {
```
ethanfrey, line 50: This is only intended for use in test code, right?

I would not enforce it with `#[cfg(test)]`, but at least add a line in the comment to mention it.

ethanfrey (commented):
Nice initiative and you seem to be finding your way around the repo quite well.
Added some comments for a few requested changes, otherwise looks good.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -130,7 +130,7 @@ where
 pub fn check_staking_permissions(
     staking_msg: &StakingMsg,
     permissions: Permissions,
-) -> Result<bool, ContractError> {
+) -> Result<(), ContractError> {
```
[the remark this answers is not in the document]
hashedone: Yes, it is exactly why I changed it from bool to ZST - it was hurting my eyes (and also optimizer - in case when one type of Result is ZST, the Result itself becomes transparent).

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -149,13 +149,13 @@ pub fn check_staking_permissions(
...
         s => panic!("Unsupported staking message: {:?}", s),
     }
-    Ok(true)
+    Ok(())
 }
 
 pub fn check_distribution_permissions(
     distribution_msg: &DistributionMsg,
     permissions: Permissions,
-) -> Result<bool, ContractError> {
```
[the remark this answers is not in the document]
hashedone: I mean I find naming "check_" functions returning `Result` on failure handy (if I would like to return `bool` then I would call it more like question, let say - `has_distribution_permissions`). `assert_*` suggests, panic on failure from my POV (which basically happens here in one branch which I don't like - I would prefer debug assertion to check invalid branch as prerequirement, but on release I would just return error here).

contracts/cw1-subkeys/src/contract.rs
hashedone: I am actually thinking about this - this is very generic proposal, but suggestions are very much welcome. I even considered to split initialization and early execution, adding utility for executes, to achieve something like:
```
let Suite { deps, owner_info: owner, .. } = SuiteConfig {
  // Only minimal config, or maybe just `Suite::init(...)` with args required to setup properly
}
.init()
.execute_increase_allowance(&owner, SENDER1, vec![coin(1111, TOKEN)], None)
.execute_increase_allowance(&owner, SENDER2, vec![coin(2222, TOKEN)], None);
```

This would benefit also redability of test logic imho. There can be argument, that this thing hides what is actually called (the `execute` itself), but wrapping it in init hides it even more, so I am ok with that.

contracts/cw1-subkeys/src/contract.rs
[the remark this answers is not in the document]
ethanfrey: I returned `Result<(), ContractError>` as it captured more failure cases than 1 (a `bool` would just capture one success and one failure case).

We have 1 success case and 4 failure cases. 


Ahhh.... I misread the diff... I thought you made the bool. You removed it. :eyes:
Very good change

contracts/cw1-subkeys/src/contract.rs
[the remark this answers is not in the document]
ethanfrey: Nevermind, remove this suggestion. I misread the diff and the comments up here are wrong.
Your way is better (and what I was arguing for)

contracts/cw1-whitelist/src/msg.rs
hashedone: I agree, I've considered it. The only reason why I didn't put `#[cfg(test)]` is, that it prevented it from running example as doctest, but it is not a good reason tbh, and I think I am leading to conditional compile it (either as `#[cfg(test}]` or `#[cfg(feature=test-utils)]`.

ethanfrey:
For some reason I cannot reply to this, but I will inline it:

> I am actually thinking about this - this is very generic proposal, but suggestions are very much welcome. I even considered to split initialization and early execution, adding utility for executes, to achieve something like:

```
let Suite { deps, owner_info: owner, .. } = SuiteConfig {
  // Only minimal config, or maybe just `Suite::init(...)` with args required to setup properly
}
.init()
.execute_increase_allowance(&owner, SENDER1, vec![coin(1111, TOKEN)], None)
.execute_increase_allowance(&owner, SENDER2, vec![coin(2222, TOKEN)], None);
```

> This would benefit also redability of test logic imho. There can be argument, that this thing hides what is actually called (the `execute` itself), but wrapping it in init hides it even more, so I am ok with that.

I am happy adding all kinds of magic helpers to SuiteConfig / init. Anything to build up the Suite. This is just to set up initial state, we trust that code works.

The code that is later being tested (execute), I would like to explicitly call.
The test body and the expected results should be explicit. The setup code is more or less boilerplate we can optimize away somehow

ethanfrey (commented):
Misread. I agree with the points on top.
Only comments relate to test setup helpers

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -149,13 +149,13 @@ pub fn check_staking_permissions(
         }
         s => panic!("Unsupported staking message: {:?}", s),
```
ethanfrey, line 150: This should return Err().
Good eye

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -84,11 +84,11 @@ where
...
 {
     // Wrap `msgs` in SubMsg.
-    let msgs: Vec<SubMsg<_>> = msgs.into_iter().map(SubMsg::new).collect();
+    let msgs: Vec<_> = msgs.into_iter().map(SubMsg::new).collect();
     let cfg = ADMIN_LIST.load(deps.storage)?;
-    // This is the admin behavior (same as cw1-whitelist)
-    if cfg.is_admin(info.sender.as_ref()) { // Just relay the messages
-    } else {
+
+    // Not an admin - need to check for permissions
```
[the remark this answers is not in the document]
maurolacy: I was tempted to do exactly that, but hesitated and left it as it was, to try to clarify that in the case of an admin, the messages are still being relayed.

I'm OK with this as it's simpler.

contracts/cw1-subkeys/src/contract.rs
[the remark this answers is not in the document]
maurolacy: :+1

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -436,6 +438,138 @@ mod tests {
...
+        ($left:expr, $right:expr, $cmp:expr $(,)?) => {
+            let mut left = $left;
+            left.sort_by(&$cmp);
+
+            let mut right = $right;
+            right.sort_by($cmp);
+
+            assert_eq!(left, right);
+        };
+    }
```
maurolacy, line 571: Nice.

contracts/cw1-subkeys/src/state.rs
```diff
@@ -34,5 +34,38 @@ pub struct Allowance {
     pub expires: Expiration,
 }
 
+impl Allowance {
+    /// Utility function forconverting message to its canonical form, so two messages with
```
maurolacy, line 38:
```suggestion
    /// Utility function for converting message to its canonical form, so two messages with
```

contracts/cw1-subkeys/src/state.rs
```diff
@@ -34,5 +34,38 @@ pub struct Allowance {
...
+    /// different representation but same semantical meaning can be easly compared.
+    ///
+    /// It could be encapsulated in custom `PartialEq` implementation, but `PartialEq` is expected
+    /// to be quickly, so it seems to be reasonable to keep it as representation-equality, and
```
maurolacy, line 42:
```suggestion
    /// to be fast, so it seems to be reasonable to keep it as representation-equality, and
```
?

contracts/cw1-whitelist/src/msg.rs
```diff
@@ -46,3 +46,35 @@ pub struct AdminListResponse {
...
+    /// Utility function forconverting message to its canonical form, so two messages with
```
maurolacy, line 51:
```suggestion
    /// Utility function for converting message to its canonical form, so two messages with
```

contracts/cw1-whitelist/src/msg.rs
```diff
@@ -46,3 +46,35 @@ pub struct AdminListResponse {
...
+    /// different representation but same semantical meaning can be easly compared.
+    ///
+    /// It could be encapsulated in custom `PartialEq` implementation, but `PartialEq` is expected
+    /// to be quickly, so it seems to be reasonable to keep it as representation-equality, and
```
maurolacy, line 55: Same comment as above applies.

contracts/cw1-whitelist/src/msg.rs
maurolacy: Yes, let's indicate this and the `Allowance` above are for testing purposes.

If these turn out to be useful in multiple places / contracts, we may consider a `test_helpers.rs` somewhere.

contracts/cw1-whitelist/src/msg.rs
```diff
@@ -46,3 +46,35 @@ pub struct AdminListResponse {
...
+    /// let resp2 = AdminListResponse {
+    ///   admins: vec!["admin2".to_owned(), "admin1".to_owned(), "admin2".to_owned()],
+    ///   mutable: true,
+    /// };
+    ///
+    /// assert_eq!(resp1.canonical(), resp2.canonical());
+    /// ```
+    pub fn canonical(mut self) -> Self {
+        self.admins.sort();
+        self.admins.dedup();
```
maurolacy, line 77: Shouldn't be necessary, but does not hurt either.

maurolacy (commented):
I like these test helper structs / macros.

What we miss in being able to reference existing variables in test assertions, we gain in clarity and conciseness. Good work.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -507,477 +614,388 @@ mod tests {
...
+                spender: SPENDER3.to_owned(),
+                balance: NativeBalance(vec![coin(3456, TOKEN)]),
+                expires: Expiration::AtHeight(12345),
+            },
+        ];
+
+        assert_sorted_eq!(
+            expected,
+            [batch1, batch2].concat(),
+            |l: &AllowanceInfo, r: &AllowanceInfo| l.spender.cmp(&r.spender)
```
maurolacy, line 721: Nice trick relying on `PartialEq` for sorting, instead of requiring `Ord` here.

I understand it would make sense to just add `Ord` (and `Eq`) to `AllowanceInfo`, and simplify this.

As that can require adding them to other stuff, including some structs in cosmwasm-std, I would say, let's create an issue to do just this, at least for cosmwasm-std.

contracts/cw1-subkeys/src/state.rs
```diff
@@ -34,5 +34,38 @@ pub struct Allowance {
     pub expires: Expiration,
 }
 
+impl Allowance {
+    /// Utility function forconverting message to its canonical form, so two messages with
+    /// different representation but same semantical meaning can be easly compared.
```
maurolacy, line 39:
```suggestion
    /// different representation but same semantic meaning can be easily compared.
```

contracts/cw1-subkeys/src/contract.rs
hashedone: Defienetely implementing `Ord` for `AllowanceInfo` does not make sense - `Ord` is for types with reasonable sorting, and I find implementing it for this kind of type a missuse. Note that even standard library doesn't implement sort for types like floats (even if you could technically provide implementation which would somehow properly sort it). Actually it is questionable if there is proper way to pick proper key for comparison - is this anyhow intuitive that allowance with sender earlier in alphabet is in any way smaller than the other one?

My personal opinion is, that even if this adds some code, it avoids any confusion how things are sorted (and why). I think about two possible simplifications: 
1. try to write macro in the way, that it would be more elision friendly so type hints are not needed (actually type hints are half the length of the closure!) - unfortunatelly I have no idea how to approach this
2. extract the comparison function to `AllowanceInfo`, but not as an implementation of `Ord`, but rather testing purpose method like `AllowanceInfo::cmp_by_spender` - actually I like the idea

maurolacy: > 2\. extract the comparison function to `AllowanceInfo`, but not as an implementation of `Ord`, but rather testing purpose method like `AllowanceInfo::cmp_by_spender` - actually I like the idea

Sounds good. That, and implementing these helpers only for tests. That could include deriving something like `Ord`, which does not make sense / adds unnecessary complexity, only for tests, using conditional compilation.

contracts/cw1-subkeys/src/msg.rs
```diff
@@ -81,19 +81,108 @@ pub struct AllAllowancesResponse {
...
+    }
+}
+
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
 pub struct PermissionsInfo {
     pub spender: String,
     pub permissions: Permissions,
 }
 
+#[cfg(any(test, feature = "test-utils"))]
```
ethanfrey, line 152: Nice feature flag for imports

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -505,714 +651,1060 @@ mod tests {
...
-            &initial_spenders,
-            &initial_allowances,
-            &initial_expirations,
-        );
+    mod allowance {
+        use super::*;
+
+        #[test]
+        fn query() {
+            let Suite { deps, .. } = SuiteConfig::new()
```
ethanfrey, line 659: I'm loving the API design here.
Super concise and flexible. The builder-pattern SuiteConfig is working out very well. Much better than a number of setup functions with different args... so much more power here.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -425,17 +425,190 @@ pub fn query_all_permissions(
...
+        fn with_allowance(mut self, spender: &'static str, allowance: Coin) -> Self {
+            self.spenders
+                .entry(spender)
+                .or_default()
+                .allowances
+                .push(allowance);
+            self
+        }
+
+        fn expire_allowances(mut self, spender: &'static str, expires: Expiration) -> Self {
```
ethanfrey, line 494: Okay, so these are different calls that can be combined.

`with_allowance(SPENDER2, ..).expire_allowances(SPENDER2, ..)` works.
And `expire_allowances(SPENDER2, ..).with_allowance(SPENDER2, ..)` should also work, right?

That is a very nice API avoiding N^2 init function issues. I would love a test case (on SuiteConfig) showing this works regardless of ordering. But design is 🥇

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -505,714 +651,1060 @@ mod tests {
...
+                    spender: SPENDER2.to_owned(),
+                    permissions: NO_PERMS,
+                },
+                PermissionsInfo {
+                    spender: SPENDER3.to_owned(),
+                    permissions: NO_PERMS,
+                },
+            ];
+
+            assert_sorted_eq!(
```
ethanfrey, line 804: This concat (after we paginate through, we get all the values exactly once) is a nice test-case.
Cleaner to read and more stable if we change any implementation details, but captures business requirements

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -505,714 +651,1060 @@ mod tests {
...
+                },
+            )
+            .unwrap();
+
+            assert_eq!(rsp.messages, vec![]);
+            assert_eq!(rsp.events, vec![]);
+            assert_eq!(rsp.data, None);
+
+            assert_eq!(
+                query_all_allowances(deps.as_ref(), None, None)
```
ethanfrey, line 1064: since there is only one SPENDER here, why not use `query_allowance`?
It seems like it would be a bit simpler.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -505,714 +651,1060 @@ mod tests {
...
-        assert!(res.is_err());
-        // Verify everything stays the same for that spender
-        let allowance = query_allowance(deps.as_ref(), spender1.to_string()).unwrap();
-        assert_eq!(
-            allowance,
-            Allowance {
-                balance: NativeBalance(vec![allow1.clone(), allow2.clone()]),
-                expires: expires_height,
+        #[test]
+        #[ignore]
```
ethanfrey, line 1560: why ignore? does this fail?

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -505,714 +651,1060 @@ mod tests {
...
 
-        // Verify
-        let allowance = query_allowance(deps.as_ref(), spender2.to_string()).unwrap();
-        assert_eq!(
-            allowance,
-            Allowance {
-                balance: NativeBalance(vec![allow1.clone()]),
-                expires: expires_never,
+        #[test]
+        #[ignore]
```
ethanfrey, line 1595: ditto

ethanfrey (commented):
Very good stuff! Do you want to take it out of draft?

I made a few minor points, but the code looks great. The only concern I raise is two test cases are marked `ignore`. If you aren't sure of the expected conditions to finish them, please add a comment on them and either Mauro or I can help.

Other than that, I'd approve it.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -425,938 +425,1532 @@ pub fn query_all_permissions(
...
-                balance: NativeBalance(initial_allowances),
-                expires: expires_later,
-            }
-        );
+            Self { deps, owner }
+        }
+    }
+
+    /// Helper function for comparing vectors or another slice-like object as they would represent
+    /// set with duplications. Compares sets by first sorting elements usning provided ordering.
```
DzikiChrzan, line 593:
```suggestion
    /// set with duplications. Compares sets by first sorting elements using provided ordering.
```

contracts/cw1-subkeys/src/contract.rs
hashedone: Yes it fails. I am pretty sure this test is correct (but double check if I understand logic), and it should be implemented, and then test would be "unignored".

contracts/cw1-subkeys/src/contract.rs
hashedone: Same as above

contracts/cw1-subkeys/src/contract.rs
hashedone: Because as I am touching allowance, I want to be sure that only my specific allowance changed, and no side effect occurred. It is surprisingly almost the same (just one encapsulation more), and more generic test.

contracts/cw1-subkeys/src/contract.rs
hashedone: I personally don't like to test this simple functions as it turns out to be testing language instead of logic, but possibly it could be tested. The order should not matter as it just sets state, but making API so it matters is obviously possible, I just don't see a point to complicate testing utilities.

hashedone:
I just removed draft flag. One thing I wanted to have before merging is getting rid of `prepare_test_case` function to avoid two different approaches of setup in single test suite. One test I didn't clean up is the test of independence, but honestly my idea of how to improve this kind of testing is out of scope of this. Basically I am sure that introducing `proptest` would allow many test cases to be generated and not overflow the reader, but it is some work to do to introduce it, and it has its downsides.

contracts/cw1-subkeys/src/contract.rs
ethanfrey: If there is a bug in that testing utilities, then some test case below will surely fail.
I would only write them if a test fails unexpectedly and it seems to be a util bug.

ethanfrey (approved):
Looks good to me.
Let's merge it and I can inspect the item about independence when I have a bit of space and comment on that. (And then have a smaller PR to review there).

Maybe you can pull that out into another mod (`mod old_tests`) before merging, so it is clear which are using which util functions? Or just leave it as is, it should be clear enough what is missing still

PR #345, opened 2021-07-22
Deprecate pks

### LINKED ISSUE #344: Deprecate `pks()`
When attempting refactoring of `pks()` into `Prefix` (for consistency with `range()`), I've noticed there's already a `keys()` method(!)

So, `pks()` is redundant. Let's deprecate / remove it.

maurolacy:
Closes #344.

### DIFF
--- a/contracts/cw721-base/src/contract.rs
+++ b/contracts/cw721-base/src/contract.rs
@@ -510,13 +510,13 @@ fn query_tokens(
     let start = start_after.map(Bound::exclusive);
 
     let owner_addr = deps.api.addr_validate(&owner)?;
-    let res: Result<Vec<_>, _> = tokens()
+    let pks: Vec<_> = tokens()
         .idx
         .owner
-        .pks(deps.storage, owner_addr, start, None, Order::Ascending)
+        .prefix(owner_addr)
+        .keys(deps.storage, start, None, Order::Ascending)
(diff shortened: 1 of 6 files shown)


contracts/cw721-base/src/contract.rs
```diff
@@ -510,13 +510,13 @@ fn query_tokens(
     let start = start_after.map(Bound::exclusive);
 
     let owner_addr = deps.api.addr_validate(&owner)?;
-    let res: Result<Vec<_>, _> = tokens()
+    let pks: Vec<_> = tokens()
         .idx
         .owner
-        .pks(deps.storage, owner_addr, start, None, Order::Ascending)
+        .prefix(owner_addr)
```
ethanfrey, line 516: Very cool alternative and much clearer actually.

packages/storage-plus/src/indexes.rs
```diff
@@ -204,6 +185,17 @@ where
         self.sub_prefix(K::SubPrefix::new())
             .range(store, min, max, order)
     }
+
+    pub fn keys<'c>(
```
ethanfrey, line 189: Ahh... you added this method. I was wondering how I missed it before. Good call.

ethanfrey (approved):
Nice one

PR #356, opened 2021-07-27
Implemented expiration for cw1-subkeys contract

hashedone:
There is one edge case which is not clean - what to do, when previous allowance is expired, but no new allowance is set. There are two reasonable solutions:

* treat previous expired allowance as no allowance, so missing allowance in `IncreaseAllowance` in such case would just set expiration to `Expire::Never {}` as it is default expiration for new allowances. This feels natural, but might cause dangerous usecase: if transaction was scheduled before previous allowance expired, then probably he expected it not to change, but it would change, if transaction would be validated to late (and it would be never expiring)
* behave, as it would be setting allowance which is already expired - return an error - this seems to be way more reasonable

This obviously doesn't impact decreasing allowance, as decreasing non existing allowance is always an error (nothing to decrease).

### DIFF
--- a/contracts/cw1-subkeys/src/contract.rs
+++ b/contracts/cw1-subkeys/src/contract.rs
@@ -6,7 +6,7 @@ use std::ops::{AddAssign, Sub};
 use cosmwasm_std::entry_point;
 use cosmwasm_std::{
     attr, to_binary, BankMsg, Binary, Coin, CosmosMsg, Deps, DepsMut, DistributionMsg, Empty, Env,
-    MessageInfo, Order, Response, StakingMsg, StdError, StdResult, SubMsg,
+    MessageInfo, Order, Response, StakingMsg, StdResult, SubMsg,
 };
 use cw0::Expiration;
 use cw1::CanExecuteResponse;
@@ -76,7 +76,7 @@ pub fn execute(
 
 pub fn execute_execute<T>(
     deps: DepsMut,
-    _env: Env,
+    env: Env,
     info: MessageInfo,
     msgs: Vec<CosmosMsg<T>>,
 ) -> Result<Response<T>, ContractError>
@@ -107,6 +107,10 @@ where
                 }) => {
                     ALLOWANCES.update::<_, ContractError>(deps.storage, &info.sender, |allow| {
                         let mut allowance = allow.ok_or(ContractError::NoAllowance {})?;
+                        if is_expired(&env, allowance.expires) {
+                            return Err(ContractError::NoAllowance {});
+                        }
+
                         // Decrease allowance
                         allowance.balance = allowance.balance.sub(amount.clone())?;
                         Ok(allowance)
@@ -125,6 +129,14 @@ where
         ..Response::default()
     };
     Ok(res)
+}
+
+pub fn is_expired(env: &Env, expire: Expiration) -> bool {
+    match expire {
+        Expiration::AtTime(timestamp) => timestamp <= env.block.time,
+        Expiration::AtHeight(height) => height < env.block.height,
+        _ => false,
+    }
 }
 
 pub fn check_staking_permissions(
@@ -174,7 +186,7 @@ pub fn check_distribution_permissions(
 
 pub fn execute_increase_allowance<T>(
     deps: DepsMut,
-    _env: Env,
+    env: Env,
     info: MessageInfo,
     spender: String,
     amount: Coin,
@@ -193,11 +205,19 @@ where
         return Err(ContractError::CannotSetOwnAccount {});
     }
 
-    ALLOWANCES.update::<_, StdError>(deps.storage, &spender_addr, |allow| {
-        let mut allowance = allow.unwrap_or_default();
+    ALLOWANCES.update::<_, ContractError>(deps.storage, &spender_addr, |allow| {
+        let mut allowance = allow
+            .filter(|allow| !is_expired(&env, allow.expires))
+            .unwrap_or_default();
+
         if let Some(exp) = expires {
+            if is_expired(&env, exp) {
+                return Err(ContractError::SettingExpiredAllowance(exp));
+            }
+
             allowance.expires = exp;
         }
+
         allowance.balance.add_assign(amount.clone());
         Ok(allowance)
     })?;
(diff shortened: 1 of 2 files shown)


contracts/cw1-subkeys/src/contract.rs
```diff
@@ -107,6 +107,10 @@ where
                 }) => {
                     ALLOWANCES.update::<_, ContractError>(deps.storage, &info.sender, |allow| {
                         let mut allowance = allow.ok_or(ContractError::NoAllowance {})?;
+                        if is_expired(&env, allowance.expires) {
```
ethanfrey, line 110: Wow, super important bug here. 💯

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -127,6 +131,14 @@ where
     Ok(res)
 }
 
+pub fn is_expired(env: &Env, expire: Expiration) -> bool {
```
ethanfrey, line 134: No need to implement this, you can do:

`expire.is_expired(&env.block)`

https://github.com/CosmWasm/cosmwasm-plus/blob/main/packages/cw0/src/expiration.rs#L41-L47

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -193,11 +205,21 @@ where
         return Err(ContractError::CannotSetOwnAccount {});
     }
 
-    ALLOWANCES.update::<_, StdError>(deps.storage, &spender_addr, |allow| {
-        let mut allowance = allow.unwrap_or_default();
+    ALLOWANCES.update::<_, ContractError>(deps.storage, &spender_addr, |allow| {
+        let mut allowance = allow
+            .filter(|allow| !is_expired(&env, allow.expires))
```
ethanfrey, line 210: Good checks for expiration (and good error below).
I just request using the Expiration method

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -217,7 +239,7 @@ where
 
 pub fn execute_decrease_allowance<T>(
     deps: DepsMut,
-    _env: Env,
+    env: Env,
```
ethanfrey, line 240: Ditto.
(All these _env were good tips we were not checking for expiration.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -313,12 +346,14 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
 }
 
 // if the subkey has no allowance, return an empty struct (not an error)
-pub fn query_allowance(deps: Deps, spender: String) -> StdResult<Allowance> {
+pub fn query_allowance(deps: Deps, env: Env, spender: String) -> StdResult<Allowance> {
     // we can use unchecked here as it is a query - bad value means a miss, we never write it
     let spender = deps.api.addr_validate(&spender)?;
     let allow = ALLOWANCES
         .may_load(deps.storage, &spender)?
+        .filter(|allow| !is_expired(&env, allow.expires))
```
ethanfrey, line 352: makes sense

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -351,7 +391,9 @@ fn can_execute(deps: Deps, sender: String, msg: CosmosMsg) -> StdResult<bool> {
             let allowance = ALLOWANCES.may_load(deps.storage, &sender)?;
             match allowance {
                 // if there is an allowance, we subtract the requested amount to ensure it is covered (error on underflow)
-                Some(allow) => Ok(allow.balance.sub(amount).is_ok()),
+                Some(allow) => {
+                    Ok(!is_expired(&env, allow.expires) && allow.balance.sub(amount).is_ok())
```
ethanfrey, line 393: Another good catch

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -475,6 +526,14 @@ mod tests {
         withdraw: false,
     };
 
+    // Expiration constant working properly with default `mock_env`
+    const NON_EXPIRED_HEIGHT: Expiration = Expiration::AtHeight(22_222);
```
ethanfrey, line 528: nice to make these variables.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -660,25 +727,48 @@ mod tests {
...
 
             // Check allowances work for accounts with no balance
-            let allowance = query_allowance(deps.as_ref(), SPENDER3.to_string()).unwrap();
+            let allowance =
+                query_allowance(deps.as_ref(), mock_env(), SPENDER3.to_string()).unwrap();
             assert_eq!(allowance, Allowance::default());
         }
 
+        #[test]
+        fn query_expired() {
```
ethanfrey, line 734: :+1:

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -1142,13 +1233,118 @@ mod tests {
...
+                            expires: NON_EXPIRED_HEIGHT,
                         }
                     ]
                 }
                 .canonical(),
             );
         }
+
+        #[test]
+        fn previous_expired() {
```
ethanfrey, line 1243: Nice test, it removes previously expired when it add the new one.

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -1142,13 +1233,118 @@ mod tests {
...
+        #[test]
+        fn set_expired() {
+            let Suite {
+                mut deps, owner, ..
+            } = SuiteConfig::new()
+                .with_allowance(SPENDER1, coin(1, TOKEN1))
+                .expire_allowances(SPENDER1, NON_EXPIRED_HEIGHT)
+                .init();
+
+            execute(
```
ethanfrey, line 1291: I would usually asset which error it was. Something like:

````rust
let err = execute(...).unwrap_err();
assert!(matches!(err, ContractError::SettingExpiredAllowance(_)), "{}", err);
```

ethanfrey (approved):
Good fixes and good tests.
One change requested to not use the custom `is_expired` method (if the other one has an issue, then let's fix it).

Once you do that, feel free to merge.

hashedone:
I just somehow missed `is_expired` method - I have no idea how, I would replace all occurrences.

PR #357, opened 2021-07-27
Add transactional helper

### LINKED ISSUE #349: multi-test: cleaner use of transactions
We use `StorageTransaction::new()`, a number of places to give us:

1. ability to revert/commit changes written to a cache
2. read-only access to an underlying snapshot and write access to a cache, which is needed for wasm execution (at least)

This is also called a number of places. I would simplify this by calling it primarily one place: `Router.execute`, which works like the "ante handler" transaction wrapper from the SDK. This also means any sub messages from WasmKeeper (which call into that) also are automatically transaction wrapped. 

We can provide a helper method to make it cleaner. Something like:

```rust
as_transaction(&mut base_storage, |write_cache: &mut Storage, read_snapshot: &Storage| -> Result<T, String> {
});
```

It would wrap the base, return a cache and a read-only reference to base to the closure. Only on `Ok`, it will commit the cache to base, and it will return the Result from the closure to the caller.

ethanfrey:
Closes #349

### DIFF
--- a/packages/multi-test/src/transactions.rs
+++ b/packages/multi-test/src/transactions.rs
@@ -19,10 +19,10 @@ type BTreeMapPairRef<'a, T = Vec<u8>> = (&'a Vec<u8>, &'a T);
 
 pub fn transactional<F, T, E>(base: &mut dyn Storage, action: F) -> Result<T, E>
 where
-    F: FnOnce(&dyn Storage, &mut dyn Storage) -> Result<T, E>,
+    F: FnOnce(&mut dyn Storage, &dyn Storage) -> Result<T, E>,
 {
     let mut cache = StorageTransaction::new(base);
-    let res = action(base, &mut cache)?;
+    let res = action(&mut cache, base)?;
     cache.prepare().commit(base);
     Ok(res)
 }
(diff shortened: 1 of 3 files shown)


hashedone (approved):
I really like how this simplified things. I am concerned about `#[allow(dead_code)]` - it makes me think that something which should be publicly exposed is not, but I assume it is just because something which would later use those internally is in progress.

ethanfrey:
Yeah, those are just used in some tests. I guess I could remove or conditionally compile.

packages/multi-test/src/transactions.rs
```diff
@@ -17,6 +17,16 @@ use cosmwasm_std::{Order, Pair};
...
 
+pub fn transactional<F, T>(base: &mut dyn Storage, action: F) -> Result<T, String>
+where
+    F: FnOnce(&mut dyn Storage, &dyn Storage) -> Result<T, String>,
+{
+    let mut cache = StorageTransaction::new(base);
+    let res = action(&mut cache, base)?;
+    cache.prepare().commit(base);
+    Ok(res)
+}
```
maurolacy, line 28: Nice helper / pattern.

packages/multi-test/src/wasm.rs
```diff
@@ -513,23 +515,22 @@ where
...
+            let mut contract_storage = self.contract_storage(write_cache, &address);
+            let querier = RouterQuerier::new(router, read_store, block);
+            let env = self.get_env(address, block);
+
+            let deps = DepsMut {
+                storage: contract_storage.as_mut(),
+                api: self.api.deref(),
+                querier: QuerierWrapper::new(&querier),
+            };
+            action(handler, deps, env)
```
maurolacy, line 532: Whoa, one action inside an action.

PR #363, opened 2021-07-29
Testing sent founds visibility in multi-test

### LINKED ISSUE #347: multitest: Ensure Warm sent funds visible to querier
When we execute/instantiate Wasm message, we:

1. (optionally) move funds from sender to contract
2. execute contract
3. (optionally) dispatch sub messages that were returned (recursive)

Currently, if we move funds (1), those are not visible to the querier in (2). We need to ensure we use the proper setup for that.

hashedone:
closes #347

### DIFF
--- a/packages/multi-test/src/app.rs
+++ b/packages/multi-test/src/app.rs
@@ -269,7 +269,7 @@ mod test {
         attr, coin, coins, AllBalanceResponse, BankMsg, BankQuery, Event, Reply, SubMsg, WasmMsg,
     };
 
-    use crate::test_helpers::contracts::{payout, reflect};
+    use crate::test_helpers::contracts::{hackatom, payout, reflect};
     use crate::test_helpers::{CustomMsg, EmptyMsg};
     use crate::transactions::StorageTransaction;
     use crate::BankKeeper;
@@ -775,5 +775,46 @@ mod test {
 
         let committed = query_app(&app, &rcpt);
         assert_eq!(coins(37, "eth"), committed);
+    }
+
+    #[test]
+    fn sent_funds_properly_visible_on_execution() {
+        // Testing if funds on contract are properly visible on contract.
+        // Hackatom contract is initialized with 10btc. Then, the contract is executed, with
+        // additional 20btc. Then beneficiary balance is checked - expeced value is 30btc. 10btc
+        // would mean that sending tokens with message is not visible for this very message, and
+        // 20btc means, that only such just send funds are visible.
+        let mut app = mock_app();
+
+        let owner = Addr::unchecked("owner");
+        let beneficiary = Addr::unchecked("beneficiary");
+        app.init_bank_balance(&owner, coins(30, "btc")).unwrap();
+
+        let contract_id = app.store_code(hackatom::contract());
+        let contract = app
+            .instantiate_contract(
+                contract_id,
+                owner.clone(),
+                &hackatom::InitMsg {
+                    beneficiary: beneficiary.as_str().to_owned(),
+                },
+                &coins(10, "btc"),
+                "Hackatom",
+            )
+            .unwrap();
+
+        app.execute_contract(
+            owner.clone(),
+            contract.clone(),
+            &EmptyMsg {},
+            &coins(20, "btc"),
+        )
+        .unwrap();
+
+        // Check balance of all accounts to ensure no tokens where burned or created, and they are
+        // in correct places
+        assert_eq!(get_balance(&app, &owner), &[]);
+        assert_eq!(get_balance(&app, &contract), &[]);
+        assert_eq!(get_balance(&app, &beneficiary), coins(30, "btc"));
     }
 }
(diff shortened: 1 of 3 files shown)


packages/multi-test/src/test_helpers.rs
```diff
@@ -3,17 +3,24 @@ use schemars::JsonSchema;
...
 
+/// This is just a demo place so we can test custom message handling
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, PartialEq)]
+#[serde(rename = "snake_case")]
+pub enum CustomMsg {
+    SetName { name: String },
+    SetAge { age: u32 },
+}
+
 fn instantiate_error(
```
ethanfrey, line 24: Is there a reason you left error top level and didn't pull it into a separate file like payout and reflect?

I know it is smaller, but was there another reason?
(I like your reorg)

packages/multi-test/src/app.rs
```diff
@@ -492,9 +489,9 @@ mod test {
         assert_eq!(funds, coins(5, "eth"));
 
         // reflect count updated
-        let qres: PayoutCountResponse = app
+        let qres: payout::CountResponse = app
```
ethanfrey, line 492: All of these type names get better

packages/multi-test/src/test_helpers.rs
hashedone: I decided that those functions are just not part of specific contracts, they are generic to be used in other contracts. I actually used `query_error` this way in my new contract (not yet here).

hashedone:
I agree with merging those refactor in separate PR, but I prefer this one to be the feature one (better name, better tracking), so I just created separate one with those changes only: #365

packages/multi-test/src/app.rs
hashedone: My hint: if you have 3+ names in your file/module with exactly the same prefix, you want to have a submodule ;)

packages/multi-test/src/test_helpers.rs
hashedone: Forget it - I just missed there is whole contract for errors. I am blind.

ethanfrey:
Please ping me on discord when ready for the final review

packages/multi-test/src/test_helpers/contracts/hackatom.rs
```diff
@@ -0,0 +1,52 @@
...
+    Ok(Response::default())
+}
+
+fn execute(
+    deps: DepsMut,
+    env: Env,
+    _info: MessageInfo,
+    _msg: EmptyMsg,
+) -> Result<Response, StdError> {
+    let init = HACKATOM.load(deps.storage)?;
```
ethanfrey, line 34: Nice and simple. Anyone can pay out.
Does the job for the test and no other frills.
I like it.

packages/multi-test/src/test_helpers/contracts/hackatom.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    let resp = Response::new().add_message(BankMsg::Send {
+        to_address: init.beneficiary,
+        amount: balance,
+    });
+
+    Ok(resp)
+}
+
+fn query(_deps: Deps, _env: Env, msg: EmptyMsg) -> Result<Binary, StdError> {
```
ethanfrey, line 45: I would just return an error here if you don't support any queries.

I would actually add a query to return the beneficiary / InitMsg

packages/multi-test/src/test_helpers/contracts/hackatom.rs
```diff
@@ -0,0 +1,52 @@
...
+    to_binary(&msg)
+}
+
+pub fn contract() -> Box<dyn Contract<Empty>> {
```
ethanfrey, line 49: :+1:

packages/multi-test/src/app.rs
```diff
@@ -776,4 +776,45 @@ mod test {
...
+            owner.clone(),
+            contract.clone(),
+            &EmptyMsg {},
+            &coins(20, "btc"),
+        )
+        .unwrap();
+
+        // Check balance of all accounts to ensure no tokens where burned or created, and they are
+        // in correct places
+        assert_eq!(get_balance(&app, &owner), &[]);
```
ethanfrey, line 816: Awesome work.

I would usually add such asserts at the state between instantiate and execute, but you are correct, those are totally not needed for such a BDD test like this. If these failed for some reason, I would add some intermediary checks, otherwise it is just noise.

Looks good to me.

ethanfrey (approved):
Good stuff

packages/multi-test/src/app.rs
hashedone: I just try to find a balance between checking everything, and deciding what is relevant to case. Basically it is nice when you can understand the test just looking at it quickly. It would be ideal to test everything (like even to check if some side state didn't change - which there is no reason to happen, but for sake of completeness), but I found this checks should actually cover test case.

PR #364, opened 2021-07-29
Implement cw20-merkle-airdrop

orkunkl:
This contract implements merkle airdrops for efficient token distribution.


contracts/cw20-merkle-airdrop/src/contract.rs
```diff
@@ -0,0 +1,508 @@
...
+    let mut hash: [u8; 32] = sha3::Keccak256::digest(user_input.as_bytes())
+        .as_slice()
+        .try_into()
+        .map_err(|_| ContractError::WrongLength {})?;
+
+    for p in proof {
+        let mut proof_buf: [u8; 32] = [0; 32];
+        hex::decode_to_slice(p, &mut proof_buf)?;
+
+        hash = match bytes_cmp(hash, proof_buf) {
```
hashedone, line 156: I would convert it to something like:
```
let mut hashes = [hash, proof_buf];
hashes.sort();
sha3::Keccak256::digest(&hashes.concat())
  .as_slice().
  .try_into()
  .map_err(|_| ContractError::WrongLength {})?;
```

Honestly - I spend way too much time on looking for difference in those branches.

Also this whole loop looks very much as fold to me, which I find way more readable than mutating variable:

```
let hash = proof.into_iter().try_fold(hash, |hash, p| {
  let mut proof_buf = [0; 32];
  hex::decode_to_slice(p, &mut proof_buf)?;
  let mut hashes = [hash, proof_buf];
  hashes.sort();
  sha3::Keccak256::digest(&hashes.concat())
    .as_slice()
    .try_into()
    .map_err(|_| ConcrtactError::WrongLength {})
})?;
```

contracts/cw20-merkle-airdrop/src/contract.rs
```diff
@@ -0,0 +1,508 @@
...
+    let merkle_root = MERKLE_ROOT.load(deps.storage, stage.into())?;
+
+    let user_input: String = info.sender.to_string() + &amount.to_string();
+    let mut hash: [u8; 32] = sha3::Keccak256::digest(user_input.as_bytes())
+        .as_slice()
+        .try_into()
+        .map_err(|_| ContractError::WrongLength {})?;
+
+    for p in proof {
+        let mut proof_buf: [u8; 32] = [0; 32];
```
hashedone, line 153: Obsolete type hint.

contracts/cw20-merkle-airdrop/src/contract.rs
```diff
@@ -0,0 +1,508 @@
...
+            attr("stage", stage),
+            attr("address", info.sender),
+            attr("amount", amount),
+        ],
+        data: None,
+        events: vec![],
+    })
+}
+
+fn bytes_cmp(a: [u8; 32], b: [u8; 32]) -> std::cmp::Ordering {
```
hashedone, line 198: This is basically how is ordering defined for arrayst= in Rust std: https://doc.rust-lang.org/std/primitive.array.html#impl-Ord, there is no point to redefine it.

contracts/cw20-merkle-airdrop/src/contract.rs
```diff
@@ -0,0 +1,508 @@
...
+            vec![
+                attr("action", "claim"),
+                attr("stage", "1"),
+                attr("address", "wasm1qzy8rg0f406uvvl54dlww6ptlh30303xq2u3xu"),
+                attr("amount", "10220")
+            ]
+        );
+
+        /*
+        TODO enable after 0.16 release https://github.com/CosmWasm/cosmwasm/pull/989/files
```
hashedone, line 439: `TODO` => `TODO:` - will be recognised by most IDEs ;)

hashedone (commented):
Just left some Rust-idioms related comments, coupled with very general stuff (like files which I think shouldn't be there). I didn't review helper code at all as I have literally no knowledge of Node ecosystem (neither JS nor TS), so someone should probably take a look at this. Later I would also do a second round with attempt to understand the contract logic and verify if I can spot any bugs.

contracts/cw20-merkle-airdrop/.editorconfig
```diff
@@ -0,0 +1,11 @@
+root = true
```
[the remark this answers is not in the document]
orkunkl: Makes sense!

contracts/cw20-merkle-airdrop/src/contract.rs
orkunkl: Good one

contracts/cw20-merkle-airdrop/src/contract.rs
orkunkl: Good tip

orkunkl:
@hashedone here is an explanation on how merkle airdrop works for better context [Medium Merkle Airdrop: the Basics](https://medium.com/smartz-blog/merkle-airdrop-the-basics-9a0857fcc930)

contracts/cw20-merkle-airdrop/src/contract.rs
```diff
@@ -0,0 +1,508 @@
...
+) -> Result<Response, ContractError> {
+    // authorize owner
+    let cfg = CONFIG.load(deps.storage)?;
+    if info.sender != cfg.owner {
+        return Err(ContractError::Unauthorized {});
+    }
+
+    // validate owner change
+    let new_owner = owner
+        .ok_or(ContractError::InvalidInput {})
```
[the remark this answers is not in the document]
hashedone: I see you changed it to what I proposed, but I still don't get - why owner is optional in the message in the fist place? For me it looks like if it is not set it is an error, so why not to make it required?

contracts/cw20-merkle-airdrop/src/contract.rs
orkunkl: This is thrown by linter:
```
error: used `sort` on primitive type `array`
   --> contracts/cw20-merkle-airdrop/src/contract.rs:154:9
    |
154 |         hashes.sort();
    |         ^^^^^^^^^^^^^ help: try: `hashes.sort_unstable()`
    |
    = note: `-D clippy::stable-sort-primitive` implied by `-D warnings`
    = note: an unstable sort would perform faster without any observable difference for this data type
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#stable_sort_primitive
```

hashedone: Clippy is right here - `sort_unstable` is faster, and for primitive types order by definition doesn't matter for equal elements (as equal is "the same" for them). Just change `sort` to `sort_unstable`.

hashedone:
@ethanfrey please take additional look at contract details itself as we spoke yesterday. I do my best here, but I am not yet confident enough about smart contract logic to approve/merge it even if it is ok for me.

contracts/cw20-merkle-airdrop/src/state.rs
```diff
@@ -0,0 +1,23 @@
...
+}
+
+pub const CONFIG_KEY: &str = "config";
+pub const CONFIG: Item<Config> = Item::new(CONFIG_KEY);
+
+pub const STAGE_KEY: &str = "stage";
+pub const STAGE: Item<u8> = Item::new(STAGE_KEY);
+
+pub const MERKLE_ROOT_PREFIX: &str = "merkle_root";
+pub const MERKLE_ROOT: Map<U8Key, String> = Map::new(MERKLE_ROOT_PREFIX);
```
hashedone, line 20: I wonder - as we want root would be always hex-parsed to `[u8; 32]`, why not to keep it in the map as `[u8; 32]`? It will both simplify implementation, and make contract more space efficient, so probably slightly cheaper.

contracts/cw20-merkle-airdrop/src/contract.rs
```diff
@@ -0,0 +1,487 @@
...
+    let cfg = CONFIG.load(deps.storage)?;
+    if info.sender != cfg.owner {
+        return Err(ContractError::Unauthorized {});
+    }
+
+    // check merkle root length
+    let mut root_buf: [u8; 32] = [0; 32];
+    hex::decode_to_slice(merkle_root.to_string(), &mut root_buf)?;
+
+    let latest_stage: u8 = STAGE.load(deps.storage)?;
```
hashedone, line 108: There is `update` function for such cases:
```
let stage = STAGE.update(deps.storage, |stage| Ok(stage + 1))?;
```

Also - obsolete typehint.

contracts/cw20-merkle-airdrop/src/contract.rs
```diff
@@ -0,0 +1,487 @@
...
+    // verify not claimed
+    let claimed = CLAIM.may_load(deps.storage, (&info.sender, U8Key::from(stage)))?;
+    if claimed.is_some() {
+        return Err(ContractError::Claimed {});
+    }
+
+    let config = CONFIG.load(deps.storage)?;
+    let merkle_root = MERKLE_ROOT.load(deps.storage, stage.into())?;
+
+    let user_input: String = info.sender.to_string() + &amount.to_string();
```
hashedone, line 143: This is suboptimal, as it performs at least one (probably two) obsolete string allocations. Maybe not a big deal, but format should work better here, and be maybe even slightly cleaner:
```
let user_input = format!("{}{}", info.sender, amount);
```

Also again - obsolete type hint.

contracts/cw20-merkle-airdrop/src/msg.rs
```diff
@@ -0,0 +1,64 @@
...
+pub struct ConfigResponse {
+    pub owner: String,
+    pub cw20_token_address: String,
+}
+
+// We define a custom struct for each query response
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct MerkleRootResponse {
+    pub stage: u8,
+    pub merkle_root: String,
```
hashedone, line 48: I see, that sending hashes as `String` in message (in particular json) seems to be simple and clean, but the issue I have with that, is that it makes message bigger, and also requires additional parsing - all sums up to unnecessary gas cost increase. Also it increases class of invalid messages. I wonder if it isn't better to use `[u8; 32]` as this type (and also the same for `proof`). I guess that at the end noone would write those things by hand, so basically it should be no problem to have conversion from human-readable hash to just array of bytes.

However - I don't have strong opinion on this. @ethanfrey what is approach we prefer in such cases?

contracts/cw20-merkle-airdrop/src/state.rs
ethanfrey: Actually not so with Map :disappointed: 

The content of Map are not stored raw, but rather JSON-encoded. This is usually a good choice.
The library we use based on serde-json-core will encode [u8; N] or Vec<u8> like any other Vec<T>, that is `[1, 5, 23, 175, 216, 82, ...]` Not very efficient.

We introduced the Binary wrapper to ensure base64 encoding/decoding.

So, Binary or (manually hex-encoded) String is probably better. Maybe Binary here

ethanfrey: That said, if you want performance and you are just going from `&[u8] -> [u8; 32]`, you could use raw storage get and set (and prefixed storage if you want that). That is definitely more efficient (no encoding going on), but I think makes the code more confusing

contracts/cw20-merkle-airdrop/src/msg.rs
ethanfrey: I mention above the issues, since we are limited to JSON parsing here, not internal structs, or binary protocol.
All input-output over Msg or Storage layer will be inefficient as JSON is nowhere like writing raw bytes.

I prefer using `Binary` in general, which we can use like Vec<u8> (or &[u8]) in the contract and transparently handles base64 encoding/decoding.

If you want hex representation, so people can easier inspect it, Orkun has the write approach. We could also add a helper to cosmwasm-std HexBytes, which is like Binary but using a hex-encoding. Not sure there are so many cases, but people do seem to like to compare hashes in hex.

contracts/cw20-merkle-airdrop/src/contract.rs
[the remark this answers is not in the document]
ethanfrey: > I see you changed it to what I proposed, but I still don't get - why owner is optional in the message in the fist place? For me it looks like if it is not set it is an error, so why not to make it required?

I agree, the Msg types should capture requirements, and will enforce them in parsing without you writing any code. if we require a number to be present AND between 1 and 100, well, I would just use serde to enforce it is a value u64, and then add business logic for 1-100. But the types should be captured in the Msg classes and the rustdoc there

contracts/cw20-merkle-airdrop/NOTICE
```diff
@@ -0,0 +1,14 @@
+CW20-Merkle-Airdrop: A reference implementation for merkle airdrop on CosmWasm
+Copyright (C) 2020 Confio OÜ
```
ethanfrey, line 2: 2021

contracts/cw20-merkle-airdrop/README.md
```diff
@@ -0,0 +1,98 @@
+# CW20 Merkle Airdrop
```
ethanfrey, line 1: Thanks for this README

contracts/cw20-merkle-airdrop/README.md
```diff
@@ -0,0 +1,98 @@
...
+- Recipient claims the airdrop
+
+These two solutions are very ineffective when recipient list is big. First, costly because bank send cost for the
+distributor will be costly. Second, whole airdrop list stored in the state, again costly.
+
+Merkle Airdrop is very efficient even when recipient number is massive.
+
+This contract works with multiple airdrop rounds, meaning you can execute several airdrops using same instance.
+
+Uses **SHA3 Keccak 256** for merkle root tree construction.
```
ethanfrey, line 23: Why this choice?
To be compatible with Ethereum obviously, but do we need to?
Are there frontend tools we can reuse by doing so?

I would use sha2-256 or sha3-256 (non-keccak) or maybe blake2. Keccak is only used in Ethereum, the rest are more standard libs.

contracts/cw20-merkle-airdrop/README.md
```diff
@@ -0,0 +1,98 @@
...
+  pub cw20_token_address: String,
+}
+```
+
+#### ExecuteMsg
+
+```rust
+pub enum ExecuteMsg {
+  UpdateConfig {
+    owner: Option<String>,
```
ethanfrey, line 60: None would be to leave this contract without owner, no more updates allowed, right?
But that is rejected in the code?

contracts/cw20-merkle-airdrop/README.md
```diff
@@ -0,0 +1,98 @@
...
+  using full list.
+
+#### QueryMsg
+
+``` rust
+pub enum QueryMsg {
+    Config {},
+    MerkleRoot { stage: u8 },
+    LatestStage {},
+    IsClaimed { stage: u8, address: String },
```
ethanfrey, line 86: Good queries

contracts/cw20-merkle-airdrop/helpers/package.json
```diff
@@ -0,0 +1,60 @@
+{
```
ethanfrey: Did you build this from scratch or modify an existing project?

If it was based on other work, it would be good to check their license. MIT and BSD is more like "no what you want". Apache does require us to post notice of the prior work it builds on (give them credit/link to it). GPL is more problematic

contracts/cw20-merkle-airdrop/helpers/src/airdrop.ts
```diff
@@ -0,0 +1,45 @@
+import keccak256 from 'keccak256';
+import { MerkleTree } from 'merkletreejs';
+
+class Airdrop {
+  private tree: MerkleTree;
+
+  constructor(accounts: Array<{ address: string; amount: string }>) {
+    const leaves = accounts.map((a) => keccak256(a.address + a.amount));
+    this.tree = new MerkleTree(leaves, keccak256, { sort: true });
```
ethanfrey: Ah, `MerkleTree` is configurable, so we could use another algorithm that is well supported in Rust and JS.

Sha2 comes to mind, as this is what is used everywhere in Cosmos, and pretty much any other linux utility

contracts/cw20-merkle-airdrop/helpers/src/airdrop.ts
```diff
@@ -0,0 +1,45 @@
...
+  public verify(
+    proof: string[],
+    account: { address: string; amount: string }
+  ): boolean {
+    let hashBuf = keccak256(account.address + account.amount);
+
+    proof.forEach((proofElem) => {
+      const proofBuf = Buffer.from(proofElem, 'hex');
+
+      if (hashBuf < proofBuf) {
```
ethanfrey: I am a bit wondering on this logic, but I think I need to understand the internals of MerkleTree better first.

It doesn't really matter, but we need some test cases to ensure it matches Rust.

Rust probably does have such test cases (I will review that later).
You could write some code there to make some demo hashes and test vectors (matches and non-matches) from the Rust code.
Then write some TS tests that ensure these test vectors work in the JS code.
No need to call Rust from TS, but make sure they are the same.

contracts/cw20-merkle-airdrop/schema/config_response.json
```diff
@@ -0,0 +1,17 @@
+{
```
ethanfrey: Please remove msg.rs above (empty file), as you use src/msg.rs

contracts/cw20-merkle-airdrop/src/state.rs
```diff
@@ -0,0 +1,23 @@
...
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct Config {
+    pub owner: Addr,
+    pub cw20_token_address: Addr,
+}
+
+pub const CONFIG_KEY: &str = "config";
+pub const CONFIG: Item<Config> = Item::new(CONFIG_KEY);
+
+pub const STAGE_KEY: &str = "stage";
```
ethanfrey: 2 points:

1) we don't need to store the key separately with storage-plus, I would just combine these into one line. (same with others). `pub const STAGE: Item<u8> = Item::new("stage);` works well
2) I assume this is *latest* stage. Can you change the name or add rustdoc to clarify that

contracts/cw20-merkle-airdrop/src/state.rs
```diff
@@ -0,0 +1,23 @@
...
+pub const STAGE: Item<u8> = Item::new(STAGE_KEY);
+
+pub const MERKLE_ROOT_PREFIX: &str = "merkle_root";
+pub const MERKLE_ROOT: Map<U8Key, String> = Map::new(MERKLE_ROOT_PREFIX);
+
+pub const CLAIM_PREFIX: &str = "claim";
+pub const CLAIM: Map<(&Addr, U8Key), bool> = Map::new(CLAIM_PREFIX);
```
ethanfrey: Nice use of composite pk

contracts/cw20-merkle-airdrop/testdata/airdrop_stage_1.json
```diff
@@ -0,0 +1,9 @@
+[
```
ethanfrey: Ah, you have test vectors already. Let's share them with js land!

contracts/cw20-merkle-airdrop/src/contract.rs
```diff
@@ -0,0 +1,487 @@
...
+};
+
+use crate::state::{Config, CLAIM, CONFIG, MERKLE_ROOT, STAGE};
+
+use crate::error::ContractError;
+use crate::msg::{
+    ConfigResponse, ExecuteMsg, InstantiateMsg, IsClaimedResponse, LatestStageResponse,
+    MerkleRootResponse, MigrateMsg, QueryMsg,
+};
+use cw20::Cw20ExecuteMsg;
```
ethanfrey: Nit pick, I would move these `use` up higher (and run `cargo fmt` again). I like to list crate internal imports at the end

contracts/cw20-merkle-airdrop/src/contract.rs
```diff
@@ -0,0 +1,487 @@
...
+use cw_storage_plus::U8Key;
+use sha3::Digest;
+use std::convert::TryInto;
+
+// Version info, for migration info
+const _CONTRACT_NAME: &str = "crates.io:cw20-merkle-airdrop";
```
ethanfrey: unused. want to add the cw2 line to store them? I forget what it is, but a one-liner from most other cosmwasm-plus contracts

contracts/cw20-merkle-airdrop/src/contract.rs
```diff
@@ -0,0 +1,487 @@
...
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn instantiate(
+    deps: DepsMut,
+    _env: Env,
+    _info: MessageInfo,
+    msg: InstantiateMsg,
+) -> StdResult<Response> {
+    let config = Config {
+        owner: deps.api.addr_validate(&msg.owner)?,
```
ethanfrey: You could also:

1) Remove owner and use `info.sender` as the owner, as that often makes sense and is simpler
2) Make owner an Option, use it if provided, or default to `info.sender`.

Both seem slightly nicer for usability. But this is fine for correctness

contracts/cw20-merkle-airdrop/src/contract.rs
```diff
@@ -0,0 +1,487 @@
...
+    owner: Option<String>,
+) -> Result<Response, ContractError> {
+    // authorize owner
+    let cfg = CONFIG.load(deps.storage)?;
+    if info.sender != cfg.owner {
+        return Err(ContractError::Unauthorized {});
+    }
+
+    // validate owner change
+    let new_owner = owner.ok_or(ContractError::InvalidInput {})?;
```
ethanfrey: Same point as @hashedone 

If you want it optional (and remove all owner if set to None), then you cannot use `update`, but rather and switch between save and remove.

contracts/cw20-merkle-airdrop/src/contract.rs
```diff
@@ -0,0 +1,487 @@
...
+    let new_owner = deps.api.addr_validate(&new_owner)?;
+
+    CONFIG.update(deps.storage, |mut exists| -> StdResult<_> {
+        exists.owner = new_owner;
+        Ok(exists)
+    })?;
+
+    Ok(Response {
```
ethanfrey: This isn't on 0.16 yet is it?

`Ok(Response::new().add_attribute("action", "update_config"))`

is now a one-liner

contracts/cw20-merkle-airdrop/src/msg.rs
```diff
@@ -0,0 +1,64 @@
...
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum ExecuteMsg {
+    UpdateConfig {
+        owner: Option<String>,
+    },
+    RegisterMerkleRoot {
+        merkle_root: String,
```
ethanfrey: Please add rust doc that this is "hex-encoded merkle root" 
(and same with Claim.proof)

contracts/cw20-merkle-airdrop/src/contract.rs
```diff
@@ -0,0 +1,487 @@
...
+    info: MessageInfo,
+    merkle_root: String,
+) -> Result<Response, ContractError> {
+    // authorize owner
+    let cfg = CONFIG.load(deps.storage)?;
+    if info.sender != cfg.owner {
+        return Err(ContractError::Unauthorized {});
+    }
+
+    // check merkle root length
```
ethanfrey: This is a proper decode, but does it assert the length?
I would just manually assert the input is 64 chars and assert it explicitly

contracts/cw20-merkle-airdrop/src/contract.rs
```diff
@@ -0,0 +1,487 @@
...
+    let mut root_buf: [u8; 32] = [0; 32];
+    hex::decode_to_slice(merkle_root.to_string(), &mut root_buf)?;
+
+    let latest_stage: u8 = STAGE.load(deps.storage)?;
+    let stage = latest_stage + 1;
+
+    MERKLE_ROOT.save(deps.storage, U8Key::from(stage), &merkle_root)?;
+    STAGE.save(deps.storage, &stage)?;
+
+    Ok(Response {
```
ethanfrey: Again, this needs 0.16 Response format as above... just chain the `.add_attribute()` calls

contracts/cw20-merkle-airdrop/src/contract.rs
```diff
@@ -0,0 +1,487 @@
...
+pub fn execute_claim(
+    deps: DepsMut,
+    _env: Env,
+    info: MessageInfo,
+    stage: u8,
+    amount: Uint128,
+    proof: Vec<String>,
+) -> Result<Response, ContractError> {
+    // verify not claimed
+    let claimed = CLAIM.may_load(deps.storage, (&info.sender, U8Key::from(stage)))?;
```
ethanfrey: :+1:

contracts/cw20-merkle-airdrop/src/contract.rs
ethanfrey: I would add a test vector to ensure this matches the TS code 100%

ethanfrey (commented):
Was doing a review untl a big diff came in. Will continue later

ethanfrey:
Please remove node_modules (and include that in .gitignore) and use `git rebase -i` to splice out that commit from the git history (splitting it into 2 parts first if you want to save something).

node_modules is HUUUGGGEEE and should never be committed anywhere (there is already discussion on yarn.lock)

contracts/cw20-merkle-airdrop/helpers/package.json
orkunkl: Built from scratch.

contracts/cw20-merkle-airdrop/README.md
orkunkl: I pulled this code base from anchor protocol(Apache 2.0) licence.
I believe developers come from solidity background.
I thought of changing it but left it for your decision. I will apply it!

contracts/cw20-merkle-airdrop/schema/config_response.json
orkunkl: Oh left over code.

contracts/cw20-merkle-airdrop/src/state.rs
orkunkl: > we don't need to store the key separately with storage-plus, I would just combine these into one line. (same with others). pub const STAGE: Item<u8> = Item::new("stage); works well

Purpose of storing key separately is for making better raw query coding, when raw querying importing contract can used this variable to build a key. Thought this could be a good practice

contracts/cw20-merkle-airdrop/src/contract.rs
```diff
@@ -0,0 +1,486 @@
...
+    if claimed.is_some() {
+        return Err(ContractError::Claimed {});
+    }
+
+    let config = CONFIG.load(deps.storage)?;
+    let merkle_root = MERKLE_ROOT.load(deps.storage, stage.into())?;
+
+    let user_input = format!("{}{}", info.sender, amount);
+    let hash = sha3::Keccak256::digest(user_input.as_bytes())
```
ethanfrey, line 143: These lines (142/143-157) should be pulled out into a separate function to heavily unit test it and compare with TS test vectors. They are simple function on basic data structures and easier to heavily test than the execute methods

contracts/cw20-merkle-airdrop/src/contract.rs
```diff
@@ -0,0 +1,486 @@
...
+    let mut root_buf: [u8; 32] = [0; 32];
+    hex::decode_to_slice(merkle_root, &mut root_buf)?;
+    if root_buf != hash {
+        return Err(ContractError::VerificationFailed {});
+    }
+
+    // Update claim index to the current stage
+    CLAIM.save(deps.storage, (&info.sender, stage.into()), &true)?;
+
+    let msgs: Vec<SubMsg> = vec![SubMsg::new(WasmMsg::Execute {
```
ethanfrey, line 168: Note that the contract has no idea if there is enough funds for this.
I guess the funder must do ensure the funds are there before they set the merkle root, right?
This is fine and seems to be documented. Just noticing it.

contracts/cw20-merkle-airdrop/src/contract.rs
```diff
@@ -0,0 +1,486 @@
...
+pub fn query_latest_stage(deps: Deps) -> StdResult<LatestStageResponse> {
+    let latest_stage = STAGE.load(deps.storage)?;
+    let resp = LatestStageResponse { latest_stage };
+
+    Ok(resp)
+}
+
+pub fn query_is_claimed(deps: Deps, stage: u8, address: String) -> StdResult<IsClaimedResponse> {
+    let key: (&Addr, U8Key) = (&deps.api.addr_validate(&address)?, stage.into());
+    let is_claimed = CLAIM.load(deps.storage, key)?;
```
ethanfrey, line 225: This will return error if it is not claimed, when we want false.

Try `let is_claimed = CLAIM.may_load(deps.storage, key)?.unwrap_or(false);`

contracts/cw20-merkle-airdrop/src/contract.rs
```diff
@@ -0,0 +1,486 @@
...
+    let resp = IsClaimedResponse { is_claimed };
+
+    Ok(resp)
+}
+
+#[entry_point]
+pub fn migrate(_deps: DepsMut, _env: Env, _msg: MigrateMsg) -> StdResult<Response> {
+    Ok(Response::default())
```
ethanfrey, line 233: No-op is okay, but at least let's do the cw2 check here.

contracts/cw20-merkle-airdrop/src/contract.rs
```diff
@@ -0,0 +1,486 @@
...
+        let config: ConfigResponse = from_binary(&res).unwrap();
+        assert_eq!("owner0001", config.owner.as_str());
+
+        // Unauthorized err
+        let env = mock_env();
+        let info = mock_info("owner0000", &[]);
+        let msg = ExecuteMsg::UpdateConfig { owner: None };
+
+        let res = execute(deps.as_mut(), env, info, msg);
+        match res {
```
ethanfrey, line 302: Modern rust allows `matches!` for this.
Also, you may be able to use PartialEq.

```rust
let err = execute(deps.as_mut(), env, info, msg).unwrap_err();
assert_eq!(err, ContractError::Unauthorized{});
```

contracts/cw20-merkle-airdrop/src/contract.rs
```diff
@@ -0,0 +1,486 @@
...
+
+        let env = mock_env();
+        let info = mock_info("owner0000", &[]);
+        let msg = ExecuteMsg::RegisterMerkleRoot {
+            merkle_root: "5d4f48f147cb6cb742b376dce5626b2a036f69faec10cd73631c791780e150fc"
+                .to_string(),
+        };
+        let _res = execute(deps.as_mut(), env, info, msg).unwrap();
+
+        let msg = ExecuteMsg::Claim {
```
ethanfrey, line 382: It would be good to document where these come from. Or even better, have a script that generates many of these examples, so we can use testvectors.

You can store as json files and include in the tests with `include_bytes!()`

contracts/cw20-merkle-airdrop/src/state.rs
ethanfrey: I avoid supporting raw queries, except in a few special cases.
cw4 interface is one and has a very well define raw query support for ONE use case - is_member.

But if you prefer them, your style is not wrong, just a bit more verbose. And yes, does allow raw queries.

ethanfrey (commented):
Finished looking through it.
The actual hashing and verify logic should be in a separate function and heavily tested for compatibility with TS. Otherwise, looks good.

contracts/cw20-merkle-airdrop/README.md
orkunkl: Makes a lot of sense.

contracts/cw20-merkle-airdrop/Cargo.toml
```diff
@@ -19,15 +19,17 @@ backtraces = ["cosmwasm-std/backtraces"]
 library = []
 
 [dependencies]
-cw20 = { path = "../../packages/cw20", version = "0.7.0" }
-cosmwasm-std = { version = "0.15.0", features = ["iterator"] }
-cw-storage-plus = { path = "../../packages/storage-plus", version = "0.7.0", features = ["iterator"] }
+cw0 = { path = "../../packages/cw0", version = "0.8.0-rc1" }
```
ethanfrey, line 22: nice

ethanfrey (commented):
Good steps

contracts/cw20-merkle-airdrop/src/state.rs
orkunkl: > I avoid supporting raw queries, except in a few special cases.
Hmm why avoid?
I see many times contracts calling each others config items. IMO advocating item raw query usage could reduce gas cost chain wide.

contracts/cw20-merkle-airdrop/src/contract.rs
orkunkl:
```rust
let mut root_buf: [u8; 32] = [0; 32];
```
I though `32` runs assertion

contracts/cw20-merkle-airdrop/README.md
ethanfrey: Let's leave it.

ethanfrey (approved):
I will merge this in, so it doesn't rot as we prepare a release.

The Rust contract code is fine. However, I really would like to see some unit tests in TS land showing compatibility with the Rust hashing algorithm. That can be a follow up PR (and maybe by a JS dev if you can convince one to do so)

orkunkl:
I've finished sha256 refactor was writing tests. I will make it another PR

PR #365, opened 2021-07-29
Reorganizations of contracts in `multi-test::test_utils`

hashedone:


packages/multi-test/src/test_helpers/contracts/error.rs
```diff
@@ -0,0 +1,44 @@
...
+    Err(StdError::generic_err("Query failed"))
+}
+
+pub fn contract() -> Box<dyn Contract<Empty>> {
+    let contract: ContractWrapper<_, _, _, _, _, _, _, _, _> =
+        ContractWrapper::new(execute, instantiate, query);
+    Box::new(contract)
+}
+
+#[allow(dead_code)]
```
ethanfrey, line 36: good for completeness

ethanfrey: I wonder if we could just merge with the above? It is less efficient to add a transform from Empty -> Empty but this is only for test code? Does smell funny though

packages/multi-test/src/test_helpers/contracts/payout.rs
```diff
@@ -0,0 +1,94 @@
...
+pub fn contract() -> Box<dyn Contract<Empty>> {
+    let contract = ContractWrapper::new(execute, instantiate, query).with_sudo(sudo);
+    Box::new(contract)
+}
+
+pub fn contract_custom<C>() -> Box<dyn Contract<C>>
+where
+    C: Clone + fmt::Debug + PartialEq + JsonSchema + 'static,
+{
+    let contract = ContractWrapper::new_with_empty(execute, instantiate, query);
```
ethanfrey, line 92: This is missing the `with_sudo` line

ethanfrey (approved):
Good stuff. Let's get this in and rebase #363

packages/multi-test/src/test_helpers/contracts/error.rs
hashedone: Actually I didn't want to make this for now just because I was still thinking about this, but I see literally no reason to split those functions.

Also while we are at this point - I see no reason to make those returning `Box<dyn Contract<C>>` instead of `impl Contract<C>` - it may always be boxed when needed. I see we are in general overusing dynamic dispatch - I find them not working so well with typesystem in general.

packages/multi-test/src/test_helpers/contracts/payout.rs
hashedone: True! I did just moving around and it was missing in original code, but should be there.

packages/multi-test/src/test_helpers/contracts/error.rs
ethanfrey: >  I see we are in general overusing dynamic dispatch - I find them not working so well with typesystem in general.

I used to have 2-4 generics on almost every function. This was replaced with eg `&dyn Storage` and I found it made everything more readable and usable in general.

We can add in some more generics as you see fit, but I would like to keep the number relatively low.

ethanfrey: > I see no reason to make those returning Box<dyn Contract<C>> instead of impl Contract<C> 

You can make a commit doing that change, and if it makes code simpler, then let's use it.

packages/multi-test/src/test_helpers/contracts/error.rs
```diff
@@ -27,18 +27,10 @@ fn query(_deps: Deps, _env: Env, _msg: EmptyMsg) -> Result<Binary, StdError> {
...
 
-pub fn contract() -> Box<dyn Contract<Empty>> {
-    let contract: ContractWrapper<_, _, _, _, _, _, _, _, _> =
-        ContractWrapper::new(execute, instantiate, query);
-    Box::new(contract)
-}
-
-#[allow(dead_code)]
-pub fn contract_custom<C>() -> Box<dyn Contract<C>>
+pub fn contract<C>() -> Box<dyn Contract<C>>
```
ethanfrey, line 30: cool

packages/multi-test/src/test_helpers/contracts/payout.rs
```diff
@@ -61,7 +61,10 @@ fn execute(
         .add_attribute("action", "payout"))
 }
 
-fn sudo(deps: DepsMut, _env: Env, msg: SudoMsg) -> Result<Response, StdError> {
+fn sudo<C>(deps: DepsMut, _env: Env, msg: SudoMsg) -> Result<Response<C>, StdError>
```
ethanfrey, line 64: This is a nice way to make it more generic.
For test contracts we can do it a lot.

(In real contracts we need to specify C in the function signature, so we can compile down to a concrete wasm `#[entry_point]`. Good to show we can handle ones that don't return that as well.

ethanfrey (commented):
Good stuff

PR #368, opened 2021-08-02
Abstracting API out of tests internals so it is clearly owned by `App`

### LINKED ISSUE #353: multi-test: Pass in API access to BankKeeper
See TODO: https://github.com/CosmWasm/cosmwasm-plus/blob/15685ab98ad7a6785357ea4341dcb2a3b7999e44/packages/multi-test/src/bank.rs#L143

The ability to validate addresses is in `Api` and bank should have that as well (and any future keeper as well).

Find an easy way to pass this into BankKeeper as well. We either want to clone Api to provide multiple copies for each keeper (which is not implemented on the trait), or just pass in &dyn Api to each Keeper and hold `Box<dyn Api>` in the App. You only need `&Api` as it exposes pure functions, but I am not sure how these references play with lifetimes.

After getting Api into bank, then we can properly validate any input and reject sends to invalid addresses

hashedone:
closes #353

The reason behind whole change is that `Router` (or `WasmKeeper`) is no longer the only user of `Api`.

The easy way is to keep `Api` in `Rc`, and besides my worries about that, I tried this solution. Unfortunately `BankKeeper` which is eager to use `Api` is passed to `App` after it is configured, so it would require exposing information that things are `Rc` internally to the world, which I don't like, as it allows creator of `App` to keep his copy of `App`, and I don't like idea that `App` is nevermore owner of `Api`. I would probably go this way if only I would be able to keep `App::new` just taking `Api` with ownership - I even found a way, but it overcomplicates internals.

Another shot is instead of keeping `Api` internal for `App`, just take it there as a borrow, and deal this borrow to another users. Unfortunately it doesn't have too much upsides - app still is nevermore clean owner of `Api`, and lifetimes bounds everywhere makes thinks slightly less clean, which is not needed to happen in tests.

Also it is possible to just clone `Api` around, as it is pretty much trivial, but it would make it less flexible to extend it in future.

I took a way, where `Api` is just passed to relevant calls, just like `Storage` and other utilities. This isolates everything nicely, doesn't affect usage, and is very consistent with other interfaces.

### DIFF
--- a/packages/multi-test/src/app.rs
+++ b/packages/multi-test/src/app.rs
@@ -30,7 +30,8 @@ pub struct App<C = Empty>
 where
     C: Clone + fmt::Debug + PartialEq + JsonSchema + 'static,
 {
-    pub router: Router<C>,
+    router: Router<C>,
+    api: Box<dyn Api>,
     storage: Box<dyn Storage>,
     block: BlockInfo,
 }
@@ -41,7 +42,7 @@ where
 {
     fn raw_query(&self, bin_request: &[u8]) -> QuerierResult {
         self.router
-            .querier(self.storage.as_ref(), &self.block)
+            .querier(self.api.as_ref(), self.storage.as_ref(), &self.block)
             .raw_query(bin_request)
     }
 }
@@ -62,14 +63,15 @@ where
     C: Clone + fmt::Debug + PartialEq + JsonSchema + 'static,
 {
     pub fn new(
-        api: Box<dyn Api>,
+        api: impl Api + 'static,
         block: BlockInfo,
         bank: impl Bank + 'static,
-        storage: Box<dyn Storage>,
+        storage: impl Storage + 'static,
     ) -> Self {
         App {
-            router: Router::new(api, bank),
-            storage,
+            router: Router::new(bank),
+            api: Box::new(api),
+            storage: Box::new(storage),
             block,
         }
     }
@@ -106,10 +108,16 @@ where
         // meaning, wrap current state, all writes go to a cache, only when execute
         // returns a success do we flush it (otherwise drop it)
 
-        let (block, router) = (&self.block, &self.router);
-        transactional(self.storage.as_mut(), |write_cache, _| {
+        let Self {
+            block,
+            router,
+            api,
+            storage,
+        } = self;
+
+        transactional(storage.as_mut(), |write_cache, _| {
             msgs.into_iter()
-                .map(|msg| router.execute(write_cache, block, sender.clone(), msg))
+                .map(|msg| router.execute(api.as_ref(), write_cache, block, sender.clone(), msg))
                 .collect()
         })
     }
@@ -144,6 +152,7 @@ where
     ) -> Result<AppResponse, String> {
         let msg = to_vec(msg).map_err(|e| e.to_string())?;
         self.router.wasm.sudo(
+            self.api.as_ref(),
             contract_addr.into(),
             self.storage.as_mut(),
             &self.router,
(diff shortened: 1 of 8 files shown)


packages/multi-test/src/app.rs
```diff
@@ -758,11 +780,17 @@ mod test {
...
+                &mut cache,
+                &app.block,
+                owner.clone(),
+                msg.into(),
+            )
             .unwrap();
 
         // shows up in cache
-        let cached_rcpt = query_router(&app.router, &cache, &rcpt);
+        let cached_rcpt = query_router(&app.router, app.api.as_ref(), &cache, &rcpt);
```
maurolacy, line 793:
```suggestion
        let cached_rcpt = query_router(&app.router, &*app.api, &cache, &rcpt);
```
here and elsewhere?

packages/multi-test/src/bank.rs
```diff
@@ -136,18 +141,23 @@ impl Bank for BankKeeper {
...
+        api: &dyn Api,
+        storage: &dyn Storage,
+        request: BankQuery,
+    ) -> Result<Binary, String> {
         let bank_storage = prefixed_read(storage, NAMESPACE_BANK);
         match request {
             BankQuery::AllBalances { address } => {
-                // TODO: shall we pass in Api to make this safer?
-                let amount = self.get_balance(&bank_storage, &Addr::unchecked(address))?;
+                let address = api.addr_validate(&address).map_err(|err| err.to_string())?;
```
maurolacy, line 153: Nice.

maurolacy (approved):
lgtm.

packages/multi-test/src/app.rs
hashedone: Honestly I personally find `as_ref()` more explicit, and what is more important - wider used also in this project. That is why I would keep it here for consistency.

maurolacy: It's OK. It just seems nice to me to have all parameters referenced in more or less the same way (with `&`). But it's the same.

hashedone: Then I will change it, I somehow agree with you. Today I was working on other think at afternoon, but I will do it as first thing at the morning an merge this (as you approved I assume there are no other issues).

packages/multi-test/src/app.rs
```diff
@@ -189,29 +197,31 @@ where
     /// QuerierWrapper to interact with
     pub fn query(
         &self,
+        api: &dyn Api,
```
ethanfrey, line 200: Great solution to this problem... pass it in!

packages/multi-test/src/app.rs
```diff
@@ -62,14 +63,15 @@ where
     C: Clone + fmt::Debug + PartialEq + JsonSchema + 'static,
 {
     pub fn new(
-        api: Box<dyn Api>,
```
ethanfrey, line 65: Nice to remove Box from the public API. Better this way.

(And on reflection from you, I think maybe generics for eg. Bank implementation makes sense.. but that is a different refactor).

packages/multi-test/src/app.rs
ethanfrey: Either way. I'll defer to both your judgement for handling boxes (feel free to update other usages if desired... maybe as a separate PR)

ethanfrey (approved):
nice

packages/multi-test/src/app.rs
hashedone: I would actually prefer to be generic over all those things (like `Api` and `Storage`), as internal optimizer tells me there is additional overhead of v-call here, but... come on, those are tests. This makes usage easier, as additional type hints are never an issue. I would not do that unless there is an actual need (I found sometimes dyn traits doesn't play well with typesystem, but I don't think here is a case).

ethanfrey: If you look at `Deps` and `DepsMut` we use `dyn` for `Storage`, `Api` and `Querier`. That change was made in 0.10 or 0.11. This made writing contracts much, much easier, especially for Rust newcomers.

I am very happy with that and would avoid re-introducing generics there. But for all other types, I am open. I just got in the habit after seeing how much that simplified the code.

PR #369, opened 2021-08-02
Properly handling data in submessages in multi-test

### LINKED ISSUE #352: multi-test: data handling with replies
See the comment here: https://github.com/CosmWasm/cosmwasm-plus/blob/05a05f895b27f116e5ea999ee1c4bdc34ad74433/packages/multi-test/src/wasm.rs#L344-L386

In [SEMANTICS.md](https://github.com/CosmWasm/cosmwasm/blob/main/SEMANTICS.md#handling-the-reply) we define when a reply can change the data field of the parent response. This is not implemented. Let's implement and test it.

The event semantics can be dealt with in #348 this just needs to deal with data

hashedone:
Closes #352 

I am not entirely sure - the issue mentions only data, but I am pretty sure that events from sub messages should also be merged (however I can misunderstand something) - as this would be similar and simple change, I think I can add it to the same PR.

### DIFF
--- a/packages/multi-test/src/app.rs
+++ b/packages/multi-test/src/app.rs
@@ -307,14 +307,6 @@ mod test {
         App::new(api, env.block, bank, MockStorage::new())
     }
 
-    fn binary_app() -> App<Binary> {
-        let env = mock_env();
-        let api = MockApi::default();
-        let bank = BankKeeper::new();
-
-        App::new(api, env.block, bank, MockStorage::new())
-    }
-
     fn get_balance<C>(app: &App<C>, addr: &Addr) -> Vec<Coin>
     where
         C: Clone + fmt::Debug + PartialEq + JsonSchema,
@@ -942,8 +934,8 @@ mod test {
         fn make_echo_submsg(
             contract: Addr,
             data: impl Into<Option<&'static str>>,
-            sub_msg: Vec<SubMsg<Binary>>,
-        ) -> SubMsg<Binary> {
+            sub_msg: Vec<SubMsg>,
+        ) -> SubMsg {
             let data = data.into().map(|s| s.to_owned());
             SubMsg::new(CosmosMsg::Wasm(WasmMsg::Execute {
                 contract_addr: contract.into(),
@@ -954,7 +946,7 @@ mod test {
 
         #[test]
         fn no_submsg() {
-            let mut app = binary_app();
+            let mut app = mock_app();
 
             let owner = Addr::unchecked("owner");
 
@@ -975,12 +967,12 @@ mod test {
                 )
                 .unwrap();
 
-            assert_eq!(response.data, Some("Data".to_owned().into_bytes().into()));
+            assert_eq!(response.data, Some("Data".as_bytes().into()));
         }
 
         #[test]
         fn single_submsg() {
-            let mut app = binary_app();
+            let mut app = mock_app();
 
             let owner = Addr::unchecked("owner");
 
@@ -1001,12 +993,12 @@ mod test {
                 )
                 .unwrap();
 
-            assert_eq!(response.data, Some("Second".to_owned().into_bytes().into()));
+            assert_eq!(response.data, Some("Second".as_bytes().into()));
         }
 
         #[test]
         fn single_no_data() {
-            let mut app = binary_app();
+            let mut app = mock_app();
 
             let owner = Addr::unchecked("owner");
 
@@ -1027,12 +1019,12 @@ mod test {
                 )
                 .unwrap();
 
-            assert_eq!(response.data, Some("First".to_owned().into_bytes().into()));
+            assert_eq!(response.data, Some("First".as_bytes().into()));
         }
 
         #[test]
         fn multiple_submsg() {
-            let mut app = binary_app();
+            let mut app = mock_app();
 
             let owner = Addr::unchecked("owner");
 
@@ -1058,12 +1050,12 @@ mod test {
                 )
                 .unwrap();
 
-            assert_eq!(response.data, Some("Second".to_owned().into_bytes().into()));
+            assert_eq!(response.data, Some("Second".as_bytes().into()));
         }
 
         #[test]
         fn nested_submsg() {
-            let mut app = binary_app();
+            let mut app = mock_app();
 
             let owner = Addr::unchecked("owner");
 
@@ -1096,7 +1088,7 @@ mod test {
                 )
                 .unwrap();
 
-            assert_eq!(response.data, Some("Second".to_owned().into_bytes().into()));
+            assert_eq!(response.data, Some("Second".as_bytes().into()));
         }
     }
 }
--- a/packages/multi-test/src/app.rs
+++ b/packages/multi-test/src/app.rs
@@ -284,14 +284,14 @@ mod test {
         attr, coin, coins, AllBalanceResponse, BankMsg, BankQuery, Event, Reply, SubMsg, WasmMsg,
     };
 
-    use crate::test_helpers::contracts::{hackatom, payout, reflect};
+    use crate::test_helpers::contracts::{echo, hackatom, payout, reflect};
     use crate::test_helpers::{CustomMsg, EmptyMsg};
     use crate::transactions::StorageTransaction;
     use crate::BankKeeper;
 
     use super::*;
 
-    fn mock_app() -> App {
+    fn mock_app() -> App<Empty> {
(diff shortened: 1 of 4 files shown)


ethanfrey:
Events get appended (they should already!).

Data *may* get overwritten if set to Some(_). Only the Response of `reply` can overwrite data. All events from all executions are emitted and appended. (I actually need to clarify this with Alex this week, but for now, just throw them all together). The last time I looked at events in tests, I think they already worked like that?

packages/multi-test/src/app.rs
```diff
@@ -903,4 +911,172 @@ mod test {
...
+
+        use crate::test_helpers::EmptyMsg;
+
+        use super::*;
+
+        fn make_echo_submsg(
+            contract: Addr,
+            data: impl Into<Option<&'static str>>,
+            sub_msg: Vec<SubMsg<Binary>>,
+        ) -> SubMsg<Binary> {
```
ethanfrey, line 926: Why do you need `SubMsg<Binary>`?

They type `T` here just controls the type of CosmosMsg::Custom variant, so we can compile with code that uses it. We return an error if anyone actually returns those variants in multi-test (this is a future TODO to support it)

packages/multi-test/src/test_helpers/contracts/echo.rs
```diff
@@ -0,0 +1,62 @@
...
+    pub data: Option<String>,
+    pub sub_msg: Vec<SubMsg<Binary>>,
+}
+
+fn instantiate(
+    _deps: DepsMut,
+    _env: Env,
+    _info: MessageInfo,
+    _msg: EmptyMsg,
+) -> Result<Response<Binary>, StdError> {
```
ethanfrey, line 23: Commented above, but I am quite sure there is no need for `Response<Binary>` unless you do something like:

```rust
let msg = CosmosMsg::Custom(Binary::from(b"foo"));
res.add_message(msg);
```

The type here is unused unless you use some blockchain-specific custom return types. Which is only the case for tgrade-specific contracts (PoE), terra-specific contracts, etc. Not for any tests. Multi-test does not support those messages when returned

packages/multi-test/src/test_helpers/contracts/echo.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+fn execute(
+    _deps: DepsMut,
+    _env: Env,
+    _info: MessageInfo,
+    msg: Message,
+) -> Result<Response<Binary>, StdError> {
+    let mut resp = Response::new();
+    if let Some(data) = msg.data {
```
ethanfrey, line 34: nice logic

packages/multi-test/src/test_helpers/contracts/echo.rs
```diff
@@ -0,0 +1,62 @@
...
+        resp = resp.set_data(data.into_bytes());
+    }
+    Ok(resp.add_submessages(msg.sub_msg))
+}
+
+fn query(_deps: Deps, _env: Env, msg: EmptyMsg) -> Result<Binary, StdError> {
+    to_binary(&msg)
+}
+
+fn reply(_deps: DepsMut, _env: Env, msg: Reply) -> Result<Response<Binary>, StdError> {
```
ethanfrey, line 44: And this will re-play whatever reply data came from the submsg. Nice for tests

packages/multi-test/src/wasm.rs
```diff
@@ -425,14 +425,17 @@ where
...
-            events.extend_from_slice(&subres.events);
-        }
-        Ok(AppResponse {
-            events,
-            data: response.data,
-        })
+        let data = response
+            .messages
+            .into_iter()
+            .try_fold(response.data, |data, resend| {
```
ethanfrey, line 431: Interesting rework of loop to functional, but looks good

packages/multi-test/src/wasm.rs
```diff
@@ -425,14 +425,17 @@ where
...
+                let subres =
+                    self.execute_submsg(router, storage, block, contract.clone(), resend)?;
+                events.extend_from_slice(&subres.events);
+                Ok::<_, String>(subres.data.or(data))
```
ethanfrey, line 435: :100: great idiomatic Rust

packages/multi-test/src/app.rs
```diff
@@ -903,4 +911,172 @@ mod test {
...
+    mod replay_data_overwrite {
+        use cosmwasm_std::to_binary;
+
+        use crate::test_helpers::EmptyMsg;
+
+        use super::*;
+
+        fn make_echo_submsg(
+            contract: Addr,
+            data: impl Into<Option<&'static str>>,
```
ethanfrey, line 924: Woah, so `"name"` will implement `Into<Option<&str>>` ?

Meaning, I can use `"data"` or `None` rather than `Some("data")` / `None`?

Cool trick

packages/multi-test/src/app.rs
```diff
@@ -903,4 +911,172 @@ mod test {
...
+                    contract,
+                    &echo::Message {
+                        data: Some("Data".to_owned()),
+                        sub_msg: vec![],
+                    },
+                    &[],
+                )
+                .unwrap();
+
+            assert_eq!(response.data, Some("Data".to_owned().into_bytes().into()));
```
ethanfrey, line 958: That is some chain... Doesn't it work without the `to_owned`?

Or `"Data".to_bytes().into()`?

packages/multi-test/src/app.rs
```diff
@@ -903,4 +911,172 @@ mod test {
...
+                    contract.clone(),
+                    &echo::Message {
+                        data: Some("First".to_owned()),
+                        sub_msg: vec![make_echo_submsg(contract, "Second", vec![])],
+                    },
+                    &[],
+                )
+                .unwrap();
+
+            assert_eq!(response.data, Some("Second".to_owned().into_bytes().into()));
```
ethanfrey, line 984: Nice test case. Very correct

packages/multi-test/src/app.rs
```diff
@@ -903,4 +911,172 @@ mod test {
...
+                    contract.clone(),
+                    &echo::Message {
+                        data: Some("First".to_owned()),
+                        sub_msg: vec![make_echo_submsg(contract, None, vec![])],
+                    },
+                    &[],
+                )
+                .unwrap();
+
+            assert_eq!(response.data, Some("First".to_owned().into_bytes().into()));
```
ethanfrey, line 1010: No overwrite on None. :+1:

packages/multi-test/src/app.rs
```diff
@@ -903,4 +911,172 @@ mod test {
...
+                .unwrap();
+
+            let response = app
+                .execute_contract(
+                    owner,
+                    contract.clone(),
+                    &echo::Message {
+                        data: Some("Orig".to_owned()),
+                        sub_msg: vec![
+                            make_echo_submsg(contract.clone(), None, vec![]),
```
ethanfrey, line 1031: Very cool test case for the fold

packages/multi-test/src/app.rs
```diff
@@ -903,4 +911,172 @@ mod test {
...
+                .instantiate_contract(contract_id, owner.clone(), &EmptyMsg {}, &[], "Echo", None)
+                .unwrap();
+
+            let response = app
+                .execute_contract(
+                    owner,
+                    contract.clone(),
+                    &echo::Message {
+                        data: Some("Orig".to_owned()),
+                        sub_msg: vec![make_echo_submsg(
```
ethanfrey, line 1061: Another great test case.
I had to read it twice to ensure the final result made sense, but yes, very correct

ethanfrey (approved):
Great stuff.

I'd just ask you to remove the unneeded `Binary` in `Response<Binary>`, `SubMsg<Binary>` for echo. We may need better docs to explain when you need that T.

packages/multi-test/src/wasm.rs
hashedone: I could do that or introduce mutable variable which is updated. I find folding way more readable that mutating, as if you understand folding, nothing surprises you. Analyze mutation always requires you to carefully read the loop. Otherwise I won't bother changing it.

packages/multi-test/src/app.rs
hashedone: Yes, and if you are really hard working you can still use `Some("data")`.

ethanfrey:
Nice updates

PR #370, opened 2021-08-02
Cw20 logo spec

ethanfrey:
Add a new optional extension to the cw20 spec to allow for managing more marketing-related metadata on-chain. Including links to the projects site as well as token logo.

Please review this if there is more info needed here, to suggest better wording, to point out some security issue, or to comment on DownloadLogoResponse compatibility with React/Vue.


packages/cw20/src/logo.rs
```diff
@@ -0,0 +1,39 @@
+use cosmwasm_std::Binary;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+/// This is used for uploading logo data, or setting it in InstantiateData
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
+#[serde(rename_all = "snake_case")]
+pub enum Logo {
+    /// A reference to an externally hosted logo. Must be a valid HTTP or HTTPS URL.
+    /// PLEASE REVIEW: is this dangerous from a security point of view?
```
hashedone, line 10: I don't think it is insecure in anyway from blockchain POV - we are not using it internally. The question is if it is possible to use it on external side, and I am not aware about remote execution method using embedded images. And even if it is a case, I think it is webpage who is using this URL responsibility to ensure this image is trusted. My concern is, that keeping images in blockchain/external servers allows abuses with usage of logo, as it is more difficult to moderate this if it lies in blockchain than if it is on owner server, but this relates to both methods of uploading image.

Anyway - whatever final decision is, pleas remember to remove this comment before merging.

ethanfrey:
> I don't see any actual implementation on storing/retrieving those data - are those meant to be just interfaces?

Yes, this is just interfaces. The next task #371 is to actually implement them for cw20-base. (And maybe add helpers.rs methods in cw20)

hashedone (approved):
Then looks good. I don't merge it myself in case you want anyone else to check for those potential security issues.

PR #373, opened 2021-08-03
Responses validation in multi-test

### LINKED ISSUE #341: multitest: Enforce validity checks for returned items
wasmd enforced some constraints on the data we return. Let's assert them in multitest as well to help shake out bugs earlier. Some items:

- Label in WasmMsg::Instantiate cannot be empty ("")
- Returned attributes (also inside events) cannot have empty key or empty value
- Returned attributes (also inside events) cannot start with `_` (wasmd 0.18)
- Returned events must have a type of length >= 2

Please add any more constraints here as you notice them in wasmd

Related to #348  (maybe both resolved in one PR?)

hashedone:
closes #341 

I don't like checking the errors in test by comparing strings - I would really prefer to switch to proper error types (we have `thiserror` anyway), and if needed converting them to string at the end. If for simplicity we want to unify error types, then we could just go with handling them with `anyhow` so if needed, error types can be retrieved by downcasting.

### DIFF
--- a/packages/multi-test/src/test_helpers/contracts/echo.rs
+++ b/packages/multi-test/src/test_helpers/contracts/echo.rs
@@ -1,9 +1,11 @@
 //! Very simple echoing contract which just returns incomming string if any, but performming subcall of
-//! given message to test response
+//! given message to test response.
+//!
+//! Additionally it bypass all events and attributes send to it
 
 use cosmwasm_std::{
-    to_binary, Binary, ContractResult, Deps, DepsMut, Empty, Env, MessageInfo, Reply, Response,
-    StdError, SubMsg, SubMsgExecutionResponse,
+    to_binary, Attribute, Binary, ContractResult, Deps, DepsMut, Empty, Env, Event, MessageInfo,
+    Reply, Response, StdError, SubMsg, SubMsgExecutionResponse,
 };
 use serde::{Deserialize, Serialize};
 
@@ -13,6 +15,8 @@ use crate::{test_helpers::EmptyMsg, Contract, ContractWrapper};
 pub struct Message {
     pub data: Option<String>,
     pub sub_msg: Vec<SubMsg>,
+    pub attributes: Vec<Attribute>,
+    pub events: Vec<Event>,
 }
 
 #[allow(clippy::unnecessary_wraps)]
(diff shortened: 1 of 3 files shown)


packages/multi-test/src/wasm.rs
```diff
@@ -720,6 +724,43 @@ where
         let storage = ReadonlyPrefixedStorage::multilevel(storage, &[NAMESPACE_WASM, &namespace]);
         Box::new(storage)
     }
+
+    fn verify_attributes(attributes: &[Attribute]) -> Result<(), String> {
```
ethanfrey, line 728: Looks good

packages/multi-test/src/wasm.rs
```diff
@@ -268,6 +268,10 @@ where
                 funds,
                 label,
             } => {
+                if label.is_empty() {
```
ethanfrey, line 271: 👍

packages/multi-test/src/test_helpers/contracts/echo.rs
```diff
@@ -48,12 +57,14 @@ fn reply(_deps: DepsMut, _env: Env, msg: Reply) -> Result<Response, StdError> {
...
-                data: Some(data), ..
+                data: Some(data),
+                events,
+                ..
             }),
         ..
     } = msg
     {
-        Ok(Response::new().set_data(data))
+        Ok(Response::new().set_data(data).add_events(events))
```
ethanfrey, line 67: I would actually not add these to the `reply` call. These should be appended by the runtime when there is a SubMsg executed without the reply blob doing anything

packages/multi-test/src/wasm.rs
```diff
@@ -720,6 +724,43 @@ where
         let storage = ReadonlyPrefixedStorage::multilevel(storage, &[NAMESPACE_WASM, &namespace]);
         Box::new(storage)
     }
+
+    fn verify_attributes(attributes: &[Attribute]) -> Result<(), String> {
+        for attr in attributes {
+            if attr.key.is_empty() {
```
ethanfrey, line 730: Note in the real wasmd implementation, we trim off whitespace from the ends before checking empty, so "  " is also invalid.

The relevant Go code changes are are https://github.com/CosmWasm/wasmd/pull/562/files maybe you can use them as reference

packages/multi-test/src/wasm.rs
```diff
@@ -720,6 +724,43 @@ where
...
+
+    fn verify_response<T>(response: Response<T>) -> Result<Response<T>, String>
+    where
+        T: Clone + fmt::Debug + PartialEq + JsonSchema,
+    {
+        Self::verify_attributes(&response.attributes)?;
+
+        for event in &response.events {
+            Self::verify_attributes(&event.attributes)?;
+            if event.ty.len() < 2 {
```
ethanfrey, line 757: Same thing, I think we trim whitespace before checking length

ethanfrey (commented):
Looks good.

I would remove the events echo in reply and add some trim spaces to the length checks. Then good to merge

packages/multi-test/src/wasm.rs
hashedone: Missed it, corrected.

ethanfrey (approved):
lgtm

PR #375, opened 2021-08-04
Marketing info for cw20-base contract

### LINKED ISSUE #371: Implement cw20 logo spec for cw20-base
Once #370 is approved and merged. Match that spec and also allow setting initial marketing info / logo in InstantiateMsg (API is not defined here). No need to implement this for other tokens than cw20-base.

hashedone:
Closes #371 
Implementation of #370

### DIFF
--- a/contracts/cw20-base/src/contract.rs
+++ b/contracts/cw20-base/src/contract.rs
@@ -33,13 +33,13 @@ fn verify_xml_preamble(data: &[u8]) -> Result<(), ContractError> {
     let preamble = data
         .split_inclusive(|c| *c == b'>')
         .next()
-        .ok_or(ContractError::InvalidXMLPreamble {})?;
+        .ok_or(ContractError::InvalidXmlPreamble {})?;
 
     const PREFIX: &[u8] = b"<?xml ";
     const POSTFIX: &[u8] = b"?>";
 
     if !(preamble.starts_with(PREFIX) && preamble.ends_with(POSTFIX)) {
-        Err(ContractError::InvalidXMLPreamble {})
+        Err(ContractError::InvalidXmlPreamble {})
     } else {
         Ok(())
     }
@@ -71,7 +71,7 @@ fn verify_png_logo(logo: &[u8]) -> Result<(), ContractError> {
     if logo.len() > LOGO_SIZE_CAP {
         Err(ContractError::LogoTooBig {})
     } else if !logo.starts_with(&HEADER) {
-        Err(ContractError::InvalidPNGHeader {})
+        Err(ContractError::InvalidPngHeader {})
     } else {
         Ok(())
     }
@@ -1847,7 +1847,7 @@ mod tests {
             )
             .unwrap_err();
 
-            assert_eq!(err, ContractError::InvalidPNGHeader {});
+            assert_eq!(err, ContractError::InvalidPngHeader {});
 
             assert_eq!(
                 query_marketing_info(deps.as_ref()).unwrap(),
@@ -1898,7 +1898,7 @@ mod tests {
             )
             .unwrap_err();
 
-            assert_eq!(err, ContractError::InvalidXMLPreamble {});
+            assert_eq!(err, ContractError::InvalidXmlPreamble {});
 
             assert_eq!(
                 query_marketing_info(deps.as_ref()).unwrap(),
--- a/contracts/cw20-base/src/contract.rs
+++ b/contracts/cw20-base/src/contract.rs
@@ -506,9 +506,10 @@ pub fn query_download_logo(deps: Deps) -> Result<DownloadLogoResponse, ContractE
 #[cfg(test)]
 mod tests {
     use cosmwasm_std::testing::{mock_dependencies, mock_env, mock_info};
-    use cosmwasm_std::{coins, from_binary, CosmosMsg, StdError, SubMsg, WasmMsg};
+    use cosmwasm_std::{coins, from_binary, Addr, CosmosMsg, StdError, SubMsg, WasmMsg};
 
     use super::*;
+    use crate::msg::InstantiateMarketingInfo;
 
     fn get_balance<T: Into<String>>(deps: Deps, address: T) -> Uint128 {
         query_balance(deps, address.into()).unwrap().balance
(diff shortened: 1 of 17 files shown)


ethanfrey (commented):
First comments. In general looks good.

I need to review contracts.rs, but need to leave the keyboard now for a bit

contracts/cw20-base/src/msg.rs
```diff
@@ -107,6 +116,19 @@ pub enum ExecuteMsg {
     },
     /// Only with "approval" extension. Destroys tokens forever
     BurnFrom { owner: String, amount: Uint128 },
+    /// Only with the "marketing" extension. If authorised, updates marketing metadata.
```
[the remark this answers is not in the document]
hashedone: I think it is better solution, I was surprised those types are separated. However I see downside of this solution - you cannot update `cw20` without updating contract itself (as matches arm would not be fully covered). However it can be easly solved, just adding additional arm returning `"Err::Std("Not yet implemented")` or even just panicking with `unimplemented!()`.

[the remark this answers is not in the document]
ethanfrey: Yeah, for cw20-base, which should be a reference implementation, I think it does make sense to error if we don't implement everything (or implement anything out of spec). Only instantiation is custom.

contracts/cw20-base/src/state.rs
```diff
@@ -23,6 +24,14 @@ pub struct MinterData {
     pub cap: Option<Uint128>,
 }
 
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
+pub struct MarketingInfo {
+    pub project: Option<String>,
+    pub description: Option<String>,
+    pub marketing: Option<Addr>,
+    pub logo: Option<Logo>,
```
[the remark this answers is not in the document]
ethanfrey: On reflection, I think TokenInfo is updated on minting and consulted a number of places.
It may make sense to store TokenInfo and MarketingInfo separately. And Logo in a 3rd section (which may well be empty, use `may_load` to check it)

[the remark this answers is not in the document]
ethanfrey: But pulling out Logo is the main issue

contracts/cw20-base/src/state.rs
```diff
@@ -23,12 +24,21 @@ pub struct MinterData {
...
+}
+
 impl TokenInfo {
     pub fn get_cap(&self) -> Option<Uint128> {
         self.mint.as_ref().and_then(|v| v.cap)
     }
 }
 
 pub const TOKEN_INFO: Item<TokenInfo> = Item::new("token_info");
+pub const LOGO: Item<Logo> = Item::new("logo");
```
ethanfrey, line 42: 👍

contracts/cw20-base/src/contract.rs
```diff
@@ -14,19 +17,34 @@ use crate::allowances::{
...
 // version info for migration info
 const CONTRACT_NAME: &str = "crates.io:cw20-base";
 const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
 
+const LOGO_SIZE_CAP: usize = 5 * 1024;
+
+fn verify_logo(logo: &Logo) -> Result<(), ContractError> {
+    if matches!(
+        logo,
+              Logo::Embedded(EmbeddedLogo::Svg(logo))
```
ethanfrey, line 31: It would good to also verify the format. At least somewhat...

eg. svg starts with `<?xml` and ends with `</svg>` (or a slightly less restrictive contains search on this). Maybe some other items

png should at least have the magic number (4 bytes prefix). 

We don't need to protect against people seeing malformed data, but rather check if they tried to shove a gif as a png. Better to error on upload.

contracts/cw20-base/src/contract.rs
```diff
@@ -47,13 +65,27 @@ pub fn instantiate(
         None => None,
     };
 
+    let marketing = match msg.marketing {
```
ethanfrey, line 68: Logo cannot be set on instantiate, correct, only later?

ethanfrey: This belongs in a Rustdoc (so the js devs see it as well)

contracts/cw20-base/src/contract.rs
```diff
@@ -261,6 +299,104 @@ pub fn execute_send(
...
+    info: MessageInfo,
+    project: Option<String>,
+    description: Option<String>,
+    marketing: Option<String>,
+) -> Result<Response, ContractError> {
+    let mut config = TOKEN_INFO.load(deps.storage)?;
+
+    // If there is no marketing info, there is also noone who can update it, therefore it cannot be
+    // authorised
+    let marketing_info = config
```
ethanfrey, line 314: This method is proper code. I wonder if there is a simpler way to handle it?

contracts/cw20-base/src/contract.rs
```diff
@@ -261,6 +299,104 @@ pub fn execute_send(
...
+    }
+
+    match description {
+        Some(empty) if empty.trim().is_empty() => marketing_info.description = None,
+        Some(description) => marketing_info.description = Some(description),
+        None => (),
+    }
+
+    match marketing {
+        Some(empty) if empty.trim().is_empty() => marketing_info.marketing = None,
```
ethanfrey, line 341: I never saw if statements on the branches of a match.
I had to look twice, but this is an interesting approach.

I would have just matched on Some(marketing) and done the if inside the branch. But this works too and is a bit shorter

contracts/cw20-base/src/contract.rs
```diff
@@ -312,12 +450,48 @@ pub fn query_minter(deps: Deps) -> StdResult<Option<MinterResponse>> {
...
+            description: None,
+            marketing: None,
+            logo: None,
+        },
+    };
+
+    Ok(info)
+}
+
+pub fn query_download_logo(deps: Deps) -> StdResult<DownloadLogoResponse> {
```
ethanfrey, line 473: :+1:

contracts/cw20-base/src/contract.rs
```diff
@@ -312,12 +450,48 @@ pub fn query_minter(deps: Deps) -> StdResult<Option<MinterResponse>> {
...
+pub fn query_marketing_info(deps: Deps) -> StdResult<MarketingInfoResponse> {
+    let meta = TOKEN_INFO.load(deps.storage)?;
+    let info = match meta.marketing {
+        Some(marketing) => MarketingInfoResponse {
+            project: marketing.project,
+            description: marketing.description,
+            marketing: marketing.marketing,
+            logo: marketing.logo,
+        },
+        None => MarketingInfoResponse {
```
ethanfrey, line 462: `MarketingInfoResponse::default()` ?

ethanfrey (approved):
Nice stuff. A few more comments

contracts/cw20-base/src/contract.rs
hashedone: I was considering it, but abandoned as I decided that I am not able to perform validation in reasonable time (in terms of performance, not development time), and don't see point of such non-whole checks. In particular for svg to have reasonable check we should parse xml, as at the and `</svg>` doesn't need to be at the end (even comments may follow it), on the other hand - searching for `</svg>` may end up finding one which is specifically in the comment. Basically if someone would like to use malformed image as attack vector on someone using logo API, then we cannot protect it. The most we can actually do in reasonable time is checking xml preamble (as it shouldn't contain anything before it including comments), and probably header for png. I can add it.

contracts/cw20-base/src/contract.rs
hashedone: Yea, after all I am not sure why I did remove logo from instantiation. I though about not making the data big, but it would be big only if someone sends such logo so I don't see actual issue. I think I will bring it back in instantiation.

PR #378, opened 2021-08-05
Upgrade to cosmwasm 0.16

### LINKED ISSUE #377: Upgrade CosmWasm to 0.16.0
https://github.com/CosmWasm/cosmwasm/commit/c14937849cedf7c70f85c15b90053bdcf01e8a1c

ethanfrey:
Closes #377

### DIFF
--- a/packages/multi-test/src/app.rs
+++ b/packages/multi-test/src/app.rs
@@ -282,7 +282,7 @@ mod test {
     use cosmwasm_std::testing::MockStorage;
     use cosmwasm_std::{
         attr, coin, coins, to_binary, AllBalanceResponse, Attribute, BankMsg, BankQuery, Event,
-        Reply, SubMsg, WasmMsg,
+        Reply, StdResult, SubMsg, WasmMsg,
     };
 
     use crate::test_helpers::contracts::{echo, hackatom, payout, reflect};
@@ -672,9 +672,8 @@ mod test {
 
         // no reply writen beforehand
         let query = reflect::QueryMsg::Reply { id: 123 };
-        app.wrap()
-            .query_wasm_smart::<Reply, _, _>(&reflect_addr, &query)
-            .unwrap_err();
+        let res: StdResult<Reply> = app.wrap().query_wasm_smart(&reflect_addr, &query);
+        res.unwrap_err();
 
         // reflect sends 7 eth, success
         let msg = SubMsg::reply_always(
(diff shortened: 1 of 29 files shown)


packages/multi-test/src/app.rs
```diff
@@ -672,9 +672,8 @@ mod test {
 
         // no reply writen beforehand
         let query = reflect::QueryMsg::Reply { id: 123 };
-        app.wrap()
-            .query_wasm_smart::<Reply, _, _>(&reflect_addr, &query)
-            .unwrap_err();
+        let res: StdResult<Reply> = app.wrap().query_wasm_smart(&reflect_addr, &query);
```
ethanfrey, line 675: @uint I was forced to change this with the impl changes. I first changed `<Reply, _, _>` to simply `<Reply>`, but I got:

```
error[E0632]: cannot provide explicit generic arguments when `impl Trait` is used in argument position
   --> packages/multi-test/src/app.rs:676:33
    |
676 |             .query_wasm_smart::<Reply>(&reflect_addr, &query)
    |                                 ^^^^^ explicit generic argument not allowed
```

This is how I got it to compile.  Maybe you have another idea?

(This is only a bit of a pain with unwrap_err(), so not a big deal, mainly curious)

ethanfrey (commented):
Works, one open question

packages/multi-test/src/app.rs
uint: @ethanfrey Oh, huh. I had no idea that's a limitation. I think your change is perfectly fine, but I also think we should just switch back all the fns in the std that have type parameters alongside `impl` blocks. We (arguably) lose a little cosmetically, but if users get to use turbofish syntax again, that's worth it.

uint: Then again, something like `::<Reply, _, _>` is not pretty. No one sane will ever want to type out the last two types. Hm.

ethanfrey: No need to change here.
Since this is only needed for `unwrap_err()` (normally we just do `let res: Reply = ... ?`, I would not change the API here. Just something to be brought up if there are cases where we have impl and generics mixed and the generic cannot be easily inferred by the input type or the return value. I believe this is the case some places where we take closures as arguments.

ethanfrey: But let's not change the std lib now. I just wanted to flag it for reflection. One place in ~20 contract with a simple work-around doesn't make it a bad API decision.

PR #380, opened 2021-08-09
Multitest events match wasmd

### LINKED ISSUE #348: multi-test: ensure event handling matches wasmd 0.18 implementation
There is a lot of logic, for renames and for rejecting that is present here:

https://github.com/CosmWasm/wasmd/blob/master/x/wasm/keeper/events.go
(and include return error where the FIXME is, rather than continue)

We should implement the equivalent in multi-test

BLOCKED: Let us merge https://github.com/CosmWasm/wasmd/issues/440 first so we have final event status

ethanfrey:
Closes #348

Based on design from https://github.com/CosmWasm/wasmd/pull/581

Ensure multi-test emits the same events as documented for x/wasm

- [x] Only emit `wasm` events if attribute is not-empty
- [x] Add custom `execute`, `instantiate`, `migrate`, `reply` events
- [x] Update all tests broken by refactoring
- [x] Test that reply callback gets expected events
- [x] Update README

### DIFF
--- a/contracts/cw20-escrow/src/integration_test.rs
+++ b/contracts/cw20-escrow/src/integration_test.rs
@@ -105,12 +105,16 @@ fn escrow_happy_path_cw20_tokens() {
     let res = router
         .execute_contract(owner.clone(), cash_addr.clone(), &send_msg, &[])
         .unwrap();
-    assert_eq!(2, res.events.len());
+    assert_eq!(4, res.events.len());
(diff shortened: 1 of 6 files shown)


ethanfrey:
Alex, I just added you FYI as we now provide an environment for the multi-test framework to match wasmd (based on the spec)

alpe:
very nice! 🌷

Issue #36, opened 2020-08-12
Convert all existing code to apache license


ethanfrey:
Leave a comment that future packages may be under AGPL-3.0 later, (on a package-by-package basis), but all existing code is and will remain under Apache-2.0

PR #381, opened 2021-08-09
Properly fold results

ethanfrey:
This shows an issue coming up with multi-test

If the top-level has a reply statement, it may return a new data to over-ride what was previously set with execute / instantiate / migrate.

However, if the submessage returns a result, but we don't set the data in reply, it *must not* override the top level reply.

This has lead to improper contract data being returned in some cases (one instantiate calls another instantiate, the second data is returned on the top level).

My "quick fix" here broke the reply-set-data functionality completely, and it seems a bit tricky to address. There is a broken test case however that should show a working solution when it passes.

@hashedone I'd be happy if you take a look and push to this branch if you see fit.

### DIFF
--- a/packages/multi-test/src/app.rs
+++ b/packages/multi-test/src/app.rs
@@ -744,11 +744,14 @@ mod test {
         let top_msg = echo::Message {
             data: Some("outside".into()),
             events: vec![Event::new("topmsg").add_attribute("called", "true")],
-            sub_msg: vec![SubMsg::new(WasmMsg::Execute {
-                contract_addr: echo_addr.to_string(),
-                msg: to_binary(&sub_msg).unwrap(),
-                funds: vec![],
-            })],
+            sub_msg: vec![SubMsg::reply_always(
+                WasmMsg::Execute {
+                    contract_addr: echo_addr.to_string(),
+                    msg: to_binary(&sub_msg).unwrap(),
+                    funds: vec![],
+                },
+                1,
+            )],
             attributes: vec![],
         };
 
@@ -759,11 +762,12 @@ mod test {
         // ensure data is set via reply
         assert_eq!(res.data.unwrap().as_slice(), b"inside");
         // ensure expected events
-        assert_eq!(res.events.len(), 4, "{:?}", res.events);
+        assert_eq!(res.events.len(), 5, "{:?}", res.events);
         assert_eq!("execute", &res.events[0].ty);
         assert_eq!("wasm-topmsg", &res.events[1].ty);
         assert_eq!("execute", &res.events[2].ty);
         assert_eq!("wasm-submsg", &res.events[3].ty);
+        assert_eq!("reply", &res.events[4].ty);
     }
 
     #[test]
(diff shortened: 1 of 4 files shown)


hashedone:
I tested also #36 with this change, seems to be working.

hashedone (approved):

packages/multi-test/src/wasm.rs
```diff
@@ -512,7 +512,12 @@ where
             let subres =
                 self.execute_submsg(api, router, storage, block, contract.clone(), resend)?;
             events.extend_from_slice(&subres.events);
-            Ok::<_, String>(subres.data.or(data))
+
+            if data.is_some() {
```
ethanfrey, line 516: This may cover the test cases, but is not the general solution we need.
It still makes no distinction between data returned in the sub message execute and in the reply processing submessage execute

packages/multi-test/src/app.rs
```diff
@@ -1140,6 +1166,33 @@ mod test {
...
+                .instantiate_contract(contract_id, owner.clone(), &EmptyMsg {}, &[], "Echo", None)
+                .unwrap();
+
+            let response = app
+                .execute_contract(
+                    owner,
+                    contract.clone(),
+                    &echo::Message {
+                        data: Some("First".to_owned()),
+                        sub_msg: vec![make_echo_submsg_no_reply(contract, "Second", vec![])],
```
ethanfrey, line 1186: This does look like it passes. Hmm...

packages/multi-test/src/wasm.rs
hashedone: It does, just not in this very place. `execute_submsg` is modified, so it doesn't return data at all if reply was not executed, as in such a case those data are never to be used. At some point I even added checking for `ReplyOn` here, but when went deeper in this I realized it doesn't make sense - the logic behind deciding if reply is to be executed doesn't belong here, instead if there is no reply, then what is visible from `execute_submsg` should actually match what we expect - no data is set (as in such a case also wasmd implementation does not process this result in any way).

packages/multi-test/src/wasm.rs
```diff
@@ -389,28 +389,26 @@ where
         });
 
         // call reply if meaningful
-        if let Ok(r) = res {
+        if let Ok(mut r) = res {
```
hashedone, line 392: This if-let implementation was actually root cause of the problem. Previously when there was a submsg response received, but no reply was called, replay was returned including data received from submsg execution. It does not make sense, as submsg execution result is not propagated directly, it is only send to reply handled and forget, those in case when there is response from submsg, but no reply is called, `Ok(None)` is now returned here.

ethanfrey:
I think there is still errors here.

Please look at #384. It is a minimal change (building off lots of tests from other PRs). It fixes all tests and solves the issue. But I think it is a bit ugly.

hashedone:
This is trict equivalent of my change. I am pretty sure that you missed that I also altered `execute_submsg` function, but except of adding additional side result (which I agree is bad solution), I altered `data` to be None if reply is not called. I don't say that this is 100% bug free, but with this claim if you think there is a scenario it doesnt cover, please explain me it and I will write test to prove it covers it (or I will find a bug and correct the bug, but I am sure that in such case #348 would also fail as well as it is equivalent).

ethanfrey:
Closing in favour of #385 and #383

PR #382, opened 2021-08-09
Adaptors for all contracts and entry points from Empty -> C

ethanfrey:
Some little missing pieces I found when working on another PR.

Pulling them out here, so they can be reviewed and merged as one conceptual unit.


hashedone (approved):
Couple utilities, all looks fine.

PR #383, opened 2021-08-09
Document submsg data failures and fix them

ethanfrey:
Merge #382 first

This takes some of the work from #381 and extends it to generate a proper test suite for the reply data tests. 

4 tests fail here. Do not fix them on this branch. Rather make a PR from another branch onto this which fixes those tests and we can look at tests and solutions separately.


hashedone (approved):
Tests looks ok for me.

PR #384, opened 2021-08-09
Reply data fix, take 2

ethanfrey:
One solution to close #383

Very simple change, but a bit ugly. Happy for a cleaner, equivalent solution.


ethanfrey:
Closing in favour of the much nicer #385

PR #385, opened 2021-08-10
Corrected submessage data response handling in multi-test

hashedone:


packages/multi-test/src/wasm.rs
```diff
@@ -389,28 +392,27 @@ where
...
                         data: r.data,
                     }),
                 };
                 // do reply and combine it with the original response
-                let res2 = self._reply(api, router, storage, block, contract, reply)?;
-                // override data if set
-                if let Some(data) = res2.data {
-                    orig.data = Some(data);
-                }
+                let reply_res = self._reply(api, router, storage, block, contract, reply)?;
```
ethanfrey, line 405: Very good changes. We return the data from reply, not from the original sub message (which is only needed to fill into reply)

ethanfrey (approved):

PR #386, opened 2021-08-10
Added some missing traits on messages of cw20-base

hashedone:


ethanfrey (approved):
Always nice to have.

ethanfrey:
Will make it for 0.8.0. As well as any other little fixes you find today

PR #394, opened 2021-08-26
Use QuerierWrapper not Querier in helpers

### LINKED ISSUE #390: Accept &QuerierWrapper not &Querier in helpers
This has come up a few times and it seems to not be so usable currently. Most code has `Deps` or `DepsMut`, which contains `&QuerierWrapper`. However, the helpers [want a `Querier`](https://github.com/CosmWasm/cw-plus/blob/main/packages/cw721/src/helpers.rs#L52) only to later [cast it to a `QuerierWrapper`](https://github.com/CosmWasm/cw-plus/blob/main/packages/cw721/src/helpers.rs#L45).

That plus the generic `Q: Querier` condition makes it very hard to use. Let's make these easier to use. (All of them, not just the cw721 in the example)

ethanfrey:
Closes #390

- [x] cw20
- [x] cw721

Other ones were already updated (like cw4)

Note that this is API breaking as I had to update multitests that actually use it, so I cannot release this in 0.8.1...

### DIFF
--- a/contracts/cw20-escrow/src/integration_test.rs
+++ b/contracts/cw20-escrow/src/integration_test.rs
@@ -80,9 +80,9 @@ fn escrow_happy_path_cw20_tokens() {
     let cash = Cw20Contract(cash_addr.clone());
 
     // ensure our balances
-    let owner_balance = cash.balance(&router, owner.clone()).unwrap();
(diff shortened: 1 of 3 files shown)


hashedone (approved):
Just proper change.

PR #396, opened 2021-08-30
Remove svg preamble validation

hashedone:
Problem occurred for technically proper svg file with no preamble - turns out that the only proper way to validate svg file is full xml parse and ensuring it is proper xml with only svg related tags (but all the cdata sections and comments are allowed). It is too complicated and to costly in gas, so for now it might be better just to check file size.

### DIFF
--- a/contracts/cw20-base/src/contract.rs
+++ b/contracts/cw20-base/src/contract.rs
@@ -25,33 +25,8 @@ const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
 
 const LOGO_SIZE_CAP: usize = 5 * 1024;
 
-/// Checks if data starts with XML preamble
-fn verify_xml_preamble(data: &[u8]) -> Result<(), ContractError> {
-    // The easiest way to perform this check would be just match on regex, however regex
-    // compilation is heavy and probably not worth it.
-
-    let preamble = data
-        .split_inclusive(|c| *c == b'>')
-        .next()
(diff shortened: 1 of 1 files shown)


contracts/cw20-base/src/contract.rs
```diff
@@ -25,33 +25,8 @@ const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
...
-
-    // Additionally attributes format could be validated as they are well defined, as well as
-    // comments presence inside of preable, but it is probably not worth it.
-}
-
 /// Validates XML logo
 fn verify_xml_logo(logo: &[u8]) -> Result<(), ContractError> {
-    verify_xml_preamble(logo)?;
-
     if logo.len() > LOGO_SIZE_CAP {
```
maurolacy, line 30:
```suggestion
    if logo.is_empty() || logo.len() > LOGO_SIZE_CAP {
```
or a min size?

maurolacy (approved):
Less is more. Fine with me.

contracts/cw20-base/src/contract.rs
hashedone: Why? If `logo.is_empty()`, then `.len()` by definition returns `0` for slice, it is not an UB...

maurolacy: Nothing to do with UB. Just to add a check for minimum / non-empty too.

PR #397, opened 2021-08-31
Make cw721-base queries public

shanev:
This is so contracts that extend cw721-base can call these queries directly, especially useful in unit tests.


orkunkl:
@shanev Thanks! can you fix conflicting files? then we can merge

maurolacy:
> @shanev Thanks! can you rebase/fix conflicting files? then we can merge

Hmm, not sure if he would be able to do that properly, because we screwed `main`. Better thing would be to close this, and do it again against a rebased `main`.

Or, we dot it ourselve, as they are simple / small changes.

maurolacy:
Replaced by #399.

PR #398, opened 2021-09-01
Implementing all messages handling in mutlitest App

### LINKED ISSUE #388: Use builder pattern for App init
Extending #387 

With the generic modules, we no longer have generic initialisation functions (like setting the balance).

I propose the following builder pattern.

`Router::new()` // this will set with BankKeeper, WasmKeeper, and the rest unimplemented
`.with_staking(stakingKeeper)` // so we can extend these later
`.build_app()` //  this will create an app with MockApi and MockStorage

The main question I have is how to call the setup/init methods on custom keepers that go beyond the default module types (eg. init_balance). Here are some examples:

```rust
    // Add a new contract. Must be done on the base object, when no contracts running
    fn store_code(&mut self, code: Box<dyn Contract<C>>) -> usize;

    // Admin interface
    fn init_balance(
        &self,
        storage: &mut dyn Storage,
        account: &Addr,
        amount: Vec<Coin>,
    ) -> Result<(), String>;
```

We can either just set these on the router when building (and then pass the storage into the build_app constructor:

```rust
let mut router = Router::new();
let mut storage = MockStorage::new();
router.bank.init_balance(&mut storage, &owner, coins(1000, "etc"));
let mut app = router.build_app(storage);
```

Or we can expose these with some closure in the app.

```rust
let mut app = Router::new().build_app();
app.setup(|router: &Router, storage: &mut dyn Storage| {
  router.bank.init_balance(storage, &owner, coins(1000, "etc"));
  router.bank.init_balance(storage, &trader, coins(10, "btc"));
})
```

In either case, we must make the different keepers public on the router interface. I lean towards the second one.

hashedone:
Closes #388

### DIFF
--- a/packages/multi-test/src/app.rs
+++ b/packages/multi-test/src/app.rs
@@ -122,6 +122,14 @@ where
         self
     }
 
+    /// Overwrites default custom messages handler
+    #[track_caller]
+    pub fn with_custom(mut self, custom: impl CustomHandler<ExecC, QueryC> + 'static) -> Self {
+        assert!(self.custom.is_none(), "Custom handler already overwritten");
+        self.custom = Some(Box::new(custom));
+        self
+    }
+
     /// Overwrites default initial block
     #[track_caller]
     pub fn with_block(mut self, block: BlockInfo) -> Self {
@@ -1077,7 +1085,7 @@ mod test {
                 .execute_contract(
                     owner,
                     contract,
-                    &echo::Message {
+                    &echo::Message::<Empty> {
                         data: Some("Data".to_owned()),
                         ..echo::Message::default()
                     },
@@ -1225,7 +1233,7 @@ mod test {
             // reflect will call echo
             // echo will set the data
             // top-level app will not display the data
-            let echo_msg = echo::Message {
+            let echo_msg = echo::Message::<Empty> {
                 data: Some("my echo".into()),
                 events: vec![Event::new("echo").add_attribute("called", "true")],
                 ..echo::Message::default()
@@ -1408,7 +1416,7 @@ mod test {
                 .execute_contract(
                     owner,
                     contract,
-                    &echo::Message {
+                    &echo::Message::<Empty> {
                         data: None,
                         attributes: vec![
                             Attribute::new("   ", "value"),
@@ -1438,7 +1446,7 @@ mod test {
                 .execute_contract(
                     owner,
                     contract,
-                    &echo::Message {
+                    &echo::Message::<Empty> {
                         data: None,
                         attributes: vec![
                             Attribute::new("key", "   "),
@@ -1468,7 +1476,7 @@ mod test {
                 .execute_contract(
                     owner,
                     contract,
-                    &echo::Message {
+                    &echo::Message::<Empty> {
                         data: None,
                         events: vec![Event::new("event")
                             .add_attribute("   ", "value")
@@ -1497,7 +1505,7 @@ mod test {
                 .execute_contract(
                     owner,
                     contract,
-                    &echo::Message {
+                    &echo::Message::<Empty> {
                         data: None,
                         events: vec![Event::new("event")
                             .add_attribute("key", "   ")
@@ -1526,7 +1534,7 @@ mod test {
                 .execute_contract(
                     owner,
                     contract,
-                    &echo::Message {
+                    &echo::Message::<Empty> {
                         data: None,
                         events: vec![Event::new(" e "), Event::new("event")],
                         ..echo::Message::default()
@@ -1536,6 +1544,56 @@ mod test {
                 .unwrap_err();
 
             assert_eq!(Error::event_type_too_short("e"), err.downcast().unwrap());
+        }
+    }
+
+    mod custom_messages {
+        use super::*;
+        use mockall as ma;
+
+        #[test]
+        fn triggering_custom_msg() {
+            let api = MockApi::default();
+            let sender = api.addr_validate("sender").unwrap();
+            let owner = api.addr_validate("owner").unwrap();
+
+            let mut custom_handler = MockSimpleCustomHandler::<CustomMsg, Empty>::new();
+            custom_handler
+                .expect_execute()
+                .with(
+                    ma::predicate::always(),
+                    ma::predicate::always(),
+                    ma::predicate::eq(CustomMsg::SetAge { age: 20 }),
+                )
+                .returning(|_, _, _| {
+                    Ok(AppResponse {
+                        data: None,
+                        events: vec![],
+                    })
+                });
+
+            let mut app = AppBuilder::new()
+                .with_api(api)
+                .with_custom(custom_handler)
+                .build();
+
+            let contract_id = app.store_code(echo::custom_contract());
+            let contract = app
+                .instantiate_contract(contract_id, owner, &EmptyMsg {}, &[], "Echo", None)
+                .unwrap();
+
+            app.execute_contract(
+                sender,
+                contract,
+                &echo::Message {
+                    sub_msg: vec![SubMsg::new(CosmosMsg::Custom(CustomMsg::SetAge {
+                        age: 20,
+                    }))],
+                    ..Default::default()
+                },
+                &[],
+            )
+            .unwrap();
         }
     }
 }
--- a/packages/multi-test/src/app.rs
+++ b/packages/multi-test/src/app.rs
@@ -1,7 +1,6 @@
-use std::fmt;
+use std::fmt::{self, Debug};
 
-#[cfg(test)]
-use cosmwasm_std::testing::{mock_env, MockApi};
+use cosmwasm_std::testing::{mock_env, MockApi, MockStorage};
 use cosmwasm_std::{
     from_slice, to_vec, Addr, Api, Binary, BlockInfo, Coin, ContractResult, CosmosMsg, Empty,
     Querier, QuerierResult, QuerierWrapper, QueryRequest, Storage, SystemError, SystemResult,
@@ -14,8 +13,10 @@ use crate::contracts::Contract;
 use crate::executor::{AppResponse, Executor};
 use crate::transactions::transactional;
 use crate::wasm::{ContractData, Wasm, WasmKeeper};
+use crate::BankKeeper;
 
 use anyhow::Result as AnyResult;
+use derivative::Derivative;
 
 pub fn next_block(block: &mut BlockInfo) {
     block.time = block.time.plus_seconds(5);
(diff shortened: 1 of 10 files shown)


packages/multi-test/src/wasm.rs
```diff
@@ -861,8 +868,12 @@ mod test {
...
     }
 
-    fn mock_router() -> Router<Empty> {
-        Router::new(BankKeeper {})
+    fn mock_router() -> Router<Empty, Empty> {
+        Router {
+            wasm: Box::new(WasmKeeper::new()),
+            bank: Box::new(BankKeeper::new()),
+            custom: Box::new(MockSimpleCustomHandler::new()),
+        }
```
maurolacy, line 876: Does this calls for using the builder pattern for `Router` too?

maurolacy (commented):
Looks good.

Handling this into two separate MRs (builder pattern for `App` and custom message handling) would have made reviewing easy for me.

packages/multi-test/src/app.rs
```diff
@@ -26,21 +30,26 @@ pub fn next_block(block: &mut BlockInfo) {
...
-/// C is the custom message returned init, handle, sudo (Response<C>).
-/// All contracts must return Response<C> or Response<Empty>
-pub struct App<C = Empty>
+/// ExecC is the custom message returned init, handle, sudo (Response<C>).
+/// All contracts must return Response<C> or Response<Empty>.
+///
+/// Also `ExecC` is the custom message which is handled by custom message handler.
+///
+/// `QueryC` is custom query message handled by custom message handler.
+pub struct App<ExecC = Empty, QueryC = Empty>
```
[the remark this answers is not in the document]
hashedone: I am basically enemy of single letter names, they are semanticless. Sometimes in very obvious cases `T` makes its jobs, but besides I find it wasting more time reading, that saving writing.

packages/multi-test/src/test_helpers/contracts/echo.rs
```diff
@@ -7,37 +7,49 @@ use cosmwasm_std::{
...
-use std::fmt;
+use std::fmt::Debug;
 
-#[derive(Debug, Clone, Serialize, Deserialize, Default)]
-pub struct Message {
+use derivative::Derivative;
+
+#[derive(Debug, Clone, Serialize, Deserialize, Derivative)]
+#[derivative(Default(bound = "", new = "true"))]
+pub struct Message<ExecC>
```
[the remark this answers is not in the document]
hashedone: It is possible for sure and I think it is even used. Empty message is still a message, good enough for testing purposes.

packages/multi-test/src/wasm.rs
hashedone: I don't think it makes sense, as Router is basically internal tool. Here it is the only place it is constructed besides app, and it is for very specific test, but in general router should never be constructed on its own so I don't see too much profit from moving those 5 lines to external builder (which I don't want to expose publicly).

hashedone:
Yea, it went bigger than I expected :/ I wanted to start with all other messages (stargate and so), but I decided that I will wait for this to be reviewed as it would be too much spam.

packages/multi-test/src/custom_handler.rs
```diff
@@ -0,0 +1,60 @@
...
+        api: &dyn Api,
+        storage: &dyn Storage,
+        block: &BlockInfo,
+        msg: QueryC,
+    ) -> AnyResult<Binary>;
+}
+
+/// Simplified version of `CustomHandler` having only arguments which are not app internals - they
+/// are just discarded. Usefull for simpler mocking.
+#[automock(type QueryResult = Binary;)]
```
ethanfrey, line 31: I would feel happier removing this and just implementing the 2 needed methods, then mockall can be a dev-dependency only.

ethanfrey: In reality, we will likely have much larger blockchain specific logic in the custom handlers and just applying mocks here doesn't make too much sense usually.

packages/multi-test/src/app.rs
```diff
@@ -49,35 +58,115 @@ where
...
+    pub fn with_block(mut self, block: BlockInfo) -> Self {
+        assert!(
+            self.block.is_none(),
+            "Initial block info already overwritten"
+        );
+        self.block = Some(block);
+        self
+    }
+
+    pub fn build(self) -> App<ExecC, QueryC> {
```
ethanfrey, line 144: Nice way to provide defaults that can be overridden

packages/multi-test/src/app.rs
```diff
@@ -49,35 +58,115 @@ where
...
-    pub fn new(
-        api: impl Api + 'static,
-        block: BlockInfo,
-        bank: impl Bank + 'static,
-        storage: impl Storage + 'static,
-    ) -> Self {
+    /// Overwrites default wasm executor. Panic if already set.
+    #[track_caller]
+    pub fn with_wasm(mut self, wasm: impl Wasm<ExecC, QueryC> + 'static) -> Self {
+        assert!(self.wasm.is_none(), "Wasm executor already overwritten");
```
ethanfrey, line 93: We don't allow overriding it twice? I guess no need, but seems odd to enforce.

packages/multi-test/src/app.rs
```diff
@@ -162,28 +251,23 @@ where
     }
 }
 
-pub struct Router<C> {
-    pub wasm: Box<dyn Wasm<C>>,
-    pub bank: Box<dyn Bank>,
+pub struct Router<ExecC, QueryC> {
+    pub(crate) wasm: Box<dyn Wasm<ExecC, QueryC>>,
```
ethanfrey, line 255: I would actually leave these public. But that can be a different PR when there is a need.

packages/multi-test/src/custom_handler.rs
```diff
@@ -0,0 +1,60 @@
...
+pub trait SimpleCustomHandler<ExecC = Empty, QueryC = Empty> {
+    type QueryResult;
+
+    fn execute(&self, block: &BlockInfo, sender: Addr, msg: ExecC) -> AnyResult<AppResponse>;
+    fn query(&self, block: &BlockInfo, msg: QueryC) -> AnyResult<Binary>;
+}
+
+impl<ExecC, QueryC, T: SimpleCustomHandler<ExecC, QueryC>> CustomHandler<ExecC, QueryC> for T {
```
ethanfrey, line 39: I would just have this implementation return error (so you can test it works). For any real usage, we will have time to build a module to define the semantics of this. We will want to mock out the real blockchain behaviour, so I would tend against one-off mocks.

ethanfrey: Or maybe a FailingCustomHandler, which returns a pre-defined error (set in new?)
And SuccessCustomHandler, which returns a fixed response (set in new?)

One or both just to help in your testing in other packages, but those don't even have to be public outside of this crate.

packages/multi-test/src/wasm.rs
```diff
@@ -82,29 +83,33 @@ where
...
     ) -> AnyResult<AppResponse>;
 }
 
-pub struct WasmKeeper<C> {
+pub struct WasmKeeper<ExecC = Empty, QueryC = Empty> {
     /// code is in-memory lookup that stands in for wasm code
     /// this can only be edited on the WasmRouter, and just read in caches
-    codes: HashMap<usize, Box<dyn Contract<C>>>,
+    codes: HashMap<usize, Box<dyn Contract<ExecC>>>,
+    /// Just marker to make type elision fork when using it as `Wasm` trait
```
ethanfrey, line 96: Actually, the contract should have QueryC in it's type.

Update: I looked at this code (also in cosmwasm-std) and it seems the query types are not enforced in DepsMut (which is where they are used). This is proper here, but maybe we can look at making cosmwasm-std more type safe

(TODO: @ethanfrey make an issue on cosmwasm-std)

ethanfrey (commented):
In general, this looks good. I like the builder pattern and thanks for adding some Custom support.

I am rather strongly against using mockall here, as we will not need that in any real usage of multitest (a proper mock object should be provided by one chain). I support a minimal implementation of a custom handler (without mockall) that is just `pub(crate)` and used for the internal tests. 

Each blockchain can provide a proper implementation along with the bindings where they define CosmosMsg and QueryRequest extensions.

packages/multi-test/src/app.rs
hashedone: If you overwrite something twice in your builder then you made some mistake. Better to catch early.

packages/multi-test/src/app.rs
hashedone: Why? Any reason why user of this crate should in any point get access to internals of structure?

packages/multi-test/src/custom_handler.rs
hashedone: Basically this is the idea - specific blockchain impl just adds its own implementation. Auto-generated mock is just to make internal testing easier, but I will get rid of this in this place.

packages/multi-test/src/app.rs
ethanfrey: Not now, I have an idea, but the need will show up in a future PR.
Out of scope here. For now pub(crate) is good.

packages/multi-test/src/custom_handler.rs
```diff
@@ -58,3 +58,72 @@ impl<ExecC, QueryC, T: SimpleCustomHandler<ExecC, QueryC>> CustomHandler<ExecC,
         self.query(block, msg)
     }
 }
+
+/// Custom handler implementation panicking on each call. Assuming, that unless specific behavior
+/// is implemented, custom messages should not be send.
+pub struct PanickingCustomHandler;
+
+impl<ExecC, QueryC> SimpleCustomHandler<ExecC, QueryC> for PanickingCustomHandler
```
ethanfrey, line 66: Why even have the `SimpleCustomHandler` trait? Why not just implement `CustomHandler` directly?

Also, I think `CustomHandler` is the only type in this file that should be `pub` (rather than `pub(crate)`). These other types are only meaningful for us to test multi-test itself (not for blockchains that actually use custom types)

packages/multi-test/src/app.rs
```diff
@@ -1594,6 +1582,13 @@ mod test {
                 &[],
             )
             .unwrap();
+
+            assert_eq!(
```
ethanfrey, line 1586: This looks nice.

ethanfrey (approved):
Looks better.
I would still hide some more types (make them `pub(crate)`) to ensure/signal they are only for internal testing of multi-test

packages/multi-test/src/custom_handler.rs
hashedone: Yea, probably in contract we should not use in contracts tests. `SimpleCustomHandler` is here to make implementations kind of simpler as I assume that `app` is somehow accessible on the side, but it may be an overkill.

PR #399, opened 2021-09-01
Make cw721-base queries public

maurolacy:
On behalf of @shanev.

Closes / replaces #397.

### DIFF
--- a/contracts/cw721-base/src/contract.rs
+++ b/contracts/cw721-base/src/contract.rs
@@ -388,23 +388,23 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
     }
 }
 
-fn query_minter(deps: Deps) -> StdResult<MinterResponse> {
+pub fn query_minter(deps: Deps) -> StdResult<MinterResponse> {
     let minter_addr = MINTER.load(deps.storage)?;
     Ok(MinterResponse {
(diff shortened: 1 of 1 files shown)


maurolacy:
`vw-plus` contracts are more or less split in that they make these internal query functions public:

```bash
$ grep "fn query_" contracts/*/src/contract.rs | grep ":pub" | wc -l
28
$ grep "fn query_" contracts/*/src/contract.rs | grep -v ":pub" | wc -l
26
$ 
```

So, I would say, this is perhaps not best practice, but it's OK / allowed.

shanev:
@maurolacy What's the case against making all query functions public?

maurolacy:
> @maurolacy What's the case against making all query functions public?

None, really. It's just that it's supposed these functions must be called through the `query` handler. But as you said, it's pretty convenient for tests. And making them public is not a risk, because they cannot be called directly in a production environment anyway.

ethanfrey:
> > @maurolacy What's the case against making all query functions public?
> 
> None, really. It's just that it's supposed these functions must be called through the `query` handler. But as you said, it's pretty convenient for tests. And making them public is not a risk, because they cannot be called directly in a production environment anyway.

Nice fix.

I'm fine to make them all public. I was trying not to expose everything  but you do need all the execute* and query* functions to be public to extend a contract, so why not do that everywhere just in case someone wants to extend it?

Issue #400, opened 2021-09-03
Test execute on cw1-whitelist contract


hashedone:
Exec messages are not covered in cw1-whitelist contract, which should be improved. This cannot be tested it UT, as it involves calling another contract, so multitest should be used.

PR #402, opened 2021-09-06
CW1-whitelist execute multitest

ueco-jb:
closes https://github.com/CosmWasm/cw-plus/issues/400


contracts/cw1-whitelist/src/multitest.rs
```diff
@@ -0,0 +1,116 @@
...
+    let contract = ContractWrapper::new(
+        crate::contract::execute,
+        crate::contract::instantiate,
+        crate::contract::query,
+    );
+    Box::new(contract)
+}
+
+/// Configuration of single member
+struct MemberConfig {
```
hashedone, line 27: As it has just single field, I would just remove this proxy struct and store `String` as members.

contracts/cw1-whitelist/src/multitest.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    pub fn with_member(mut self, addr: &str) -> Self {
+        self.members.push(MemberConfig {
+            addr: addr.to_owned(),
+        });
+
+        self
+    }
+
+        pub fn init(self, admins: Vec<String>) -> Result<Suite> {
```
hashedone, line 65: `cargo fmt`

ueco-jb:
I'm opening this for review. There's actually some stuff that I could change further (for example to not call `store_code` a second time in test), but I'll do it tommorow if necessary.

Changes outside of `cw1-whitelist` directory are not part of this PR and are here because for some reason last two merges from main dissapeared.

contracts/cw1-whitelist/src/integration_tests.rs
```diff
@@ -0,0 +1,129 @@
...
+use assert_matches::assert_matches;
+use cosmwasm_std::testing::{mock_env, MockApi, MockStorage};
+use cosmwasm_std::{to_binary, Addr, CosmosMsg, Empty, QueryRequest, WasmMsg, WasmQuery};
+use cw1::Cw1Contract;
+use cw_multi_test::{App, AppResponse, BankKeeper, Contract, ContractWrapper, Executor};
+use derivative::Derivative;
+use lazy_static::lazy_static;
+
+lazy_static! {
+    static ref OWNER: String = "owner".to_string();
```
hashedone, line 12: Why to involve static var instead of instantiate it in `Suite`?

contracts/cw1-whitelist/src/integration_tests.rs
```diff
@@ -0,0 +1,129 @@
...
+            )
+            .unwrap();
+
+        Ok(Suite {
+            app,
+            whitelist: Cw1Contract(whitelist),
+        })
+    }
+
+    pub fn freeze(&mut self, contract_addr: &Addr) -> Result<AppResponse> {
```
hashedone, line 68: I find this signatre missleading. This doesn't call freeze directly, this calls `execute`, and freeze is just message to execute to ensure that execution works correctly. Better would be to go with signature like:

```
fn execute<Msg: serde::DeserializeOwned>(&mut self, contract_addr: &Addr, msg: Msg) -> Result<AppResponse>
```

And then call it:

```
execute(address, ExecuteMsg::Freeze {})?;
```

contracts/cw1-whitelist/src/integration_tests.rs
```diff
@@ -0,0 +1,129 @@
...
+                self.whitelist.addr(),
+                &execute,
+                &[],
+            )
+            .map_err(|err| anyhow!(err))
+    }
+}
+
+#[test]
+fn execute_freeze() {
```
hashedone, line 89: Also missliding name, the clue is not to execute freeze, it is to proxy freeze message. Having this test name, and no single comment on it, I assume flow:
* Have single mutable cw1 contract
* Send `freeze`
* Ensure contract is frozen

My proposal is `proxy_freeze_message`.

contracts/cw1-whitelist/src/integration_tests.rs
```diff
@@ -0,0 +1,129 @@
...
+    let mut suite = Suite::init(true).unwrap();
+
+    let cw1_id = suite.app.store_code(contract_cw1());
+    let second_contract = suite
```
hashedone, line 93: This whole instantiation could also have utility function (and then `OWNER` would not need to be exposed outside of suite).

contracts/cw1-whitelist/src/integration_tests.rs
ueco-jb: It doesn't actually matter I think.
I did it as you say earlier, but it was bit annoying to have same unused-but-required string instantiated in 3 different places. (unused because it has no use in this case)
But yeah, maybe I've overdone it. I'm looking at this now and `Owner` inside `Suite` looks fine as well.

contracts/cw1-whitelist/src/integration_tests.rs
```diff
@@ -0,0 +1,129 @@
...
+                mutable: true,
+            },
+            &[],
+            "Whitelist",
+            None,
+        )
+        .unwrap();
+
+    assert_ne!(second_contract, suite.whitelist.0);
+    assert_matches!(suite.freeze(&second_contract), Ok(_));
```
hashedone, line 109: Why not `suite.freeze(...).unwrap()`?

contracts/cw1-whitelist/src/integration_tests.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    let query_msg: QueryMsg = QueryMsg::AdminList {};
+    assert_matches!(
+        suite
```
hashedone, line 113: Also here you could do helper for this query

contracts/cw1-whitelist/src/integration_tests.rs
```diff
@@ -0,0 +1,129 @@
...
+            .wrap()
+            .query(&QueryRequest::Wasm(WasmQuery::Smart {
+                contract_addr: second_contract.to_string(),
+                msg: to_binary(&query_msg).unwrap(),
+            })
+        ),
+        Ok(
+            AdminListResponse {
+                admins: _,
+                mutable
```
hashedone, line 124: You can simplify this to:
```
AdminListResponse {
  mutable,
  ..
}
```

contracts/cw1-whitelist/src/integration_tests.rs
```diff
@@ -0,0 +1,129 @@
...
+            "Whitelist",
+            None,
+        )
+        .unwrap();
+
+    assert_ne!(second_contract, suite.whitelist.0);
+    assert_matches!(suite.freeze(&second_contract), Ok(_));
+
+    let query_msg: QueryMsg = QueryMsg::AdminList {};
+    assert_matches!(
```
hashedone, line 112: I  like using `assert_matches!` here.

contracts/cw1-whitelist/src/integration_tests.rs
hashedone: Technically you are right, but basically I find global statics code smell unless they are well argued. Here it is testing code, so it basically doesn't matter, as in test code we often allow for simplifications we wont have in production, but this is not this much a simplification, and also there is nothing wrong with having it in Suite, and additionally - it introduces dependency (a heavy one) which is not such a big deal, but in this case I don't see a reason.

hashedone (changes requested):
Mostly good changes, some stylistic notes.

contracts/cw1-whitelist/src/integration_tests.rs
ueco-jb: Just personal preference, I like having visual confirmation with result being `Ok(_)` instead of one of many unwraps in test.

contracts/cw1-whitelist/src/integration_tests.rs
ueco-jb: Done.

contracts/cw1-whitelist/src/integration_tests.rs
ueco-jb: Done.

contracts/cw1-whitelist/src/integration_tests.rs
ueco-jb: Agree, done.

contracts/cw1-whitelist/src/integration_tests.rs
hashedone: Don't care this much about this, as you have `assert_matches` here anyway (usage in other place is very nice), so you can leave it.

contracts/cw1-whitelist/src/integration_tests.rs
ueco-jb: Removed.

contracts/cw1-whitelist/src/integration_tests.rs
ueco-jb: I've put it in helper function.

contracts/cw1-whitelist/src/integration_tests.rs
ueco-jb: Done.

contracts/cw1-whitelist/src/integration_tests.rs
```diff
@@ -0,0 +1,128 @@
...
+    );
+    Box::new(contract)
+}
+
+#[derive(Derivative)]
+#[derivative(Debug)]
+pub struct Suite {
+    /// Application mock
+    #[derivative(Debug = "ignore")]
+    pub app: App,
```
hashedone, line 34: I am pretty sure you don't need this to be public.

hashedone (approved):
One single detail to be corrected, just everything else ok. Merge yourself when remove obsolete pub.

PR #409, opened 2021-09-09
Snapshot item

### LINKED ISSUE #193: Add SnapshotItem to storage-plus
Let's build the equivalent type to `SnapshotMap`, but for `Item`. We can query this at any snapshotted height. Can use the same logic for selected checkpoints (+1 for code reuse/refactor)

In particular, I want this, so I can query a snapshot of `total_weight` at the beginning of the block the proposal was created on . This is a vulnerability mentioned in one of the cw3-flex-multisig snapshoting prs/issues (can't find it now)

You can use this for snapshotting total weight in cw4-* and exposing history to cw3 callers in the same PR or in a separate.

maurolacy:
Closes #193.

Straightforward impl adapting from `SnapshotMap`. 

Still need to review / validate tests a little more.


packages/storage-plus/src/snapshot_item.rs
```diff
@@ -0,0 +1,354 @@
...
+    ) -> Self {
+        SnapshotItem {
+            primary: Item::new(storage_key),
+            checkpoints: Map::new(checkpoints),
+            changelog: Map::new(changelog),
+            strategy,
+        }
+    }
+
+    pub fn add_checkpoint(&self, store: &mut dyn Storage, height: u64) -> StdResult<()> {
```
ethanfrey, line 46: We could remove the checkpoint/strategy logic here and just make it "always".
Or refactor this out to share the same logic in both item and map.

It is too complex for me to feel comfortable with copy-and-paste implementation.

ethanfrey (commented):
In general, looks good.

I would either just hardcode "always checkpoint" to simplify the SnapshotItem (preferred) or figure out a way to reuse logic between both, as those concepts are non-trivial to code and test.

packages/storage-plus/src/snapshot_item.rs
maurolacy: I'll put this helper in the `snapshot` module, and use it from both structs.

ethanfrey:
I'm doing a deeper dig through the code now. The diffs don't help, I need to check it out.

The first comment is with snapshot.rs, snapshot_map.rs and snapshot_item.rs, maybe they deserve a submodule. This package is growing and could use some structure. Like `snapshot` and below that `mod.rs` `map.rs` and `item.rs`. And then re-exported on the top level. 

Maybe index code could also be grouped like this as well, but that is a different PR.

packages/storage-plus/src/keys.rs
```diff
@@ -19,6 +19,16 @@ pub trait PrimaryKey<'a>: Clone {
     }
 }
 
+// Empty / no primary key
+impl<'a> PrimaryKey<'a> for () {
```
ethanfrey, line 23: Now I get it...

`Item<T>` is just another name for `Map<(), T>` (but with more clear methods).
Now I see where re-use comes in. Nice idea.

packages/storage-plus/src/snapshot_item.rs
```diff
@@ -0,0 +1,290 @@
...
+use crate::keys::U64Key;
+use crate::snapshot::{ChangeSet, Snapshot};
+use crate::{Bound, Item, Strategy};
+
+/// Item that maintains a snapshot of one or more checkpoints.
+/// We can query historical data as well as current state.
+/// What data is snapshotted depends on the Strategy.
+pub struct SnapshotItem<'a, T> {
+    primary: Item<'a, T>,
+    snapshots: Snapshot<'a, (), T>,
```
ethanfrey, line 17: Very nice way to unify this.

packages/storage-plus/src/snapshot_item.rs
```diff
@@ -0,0 +1,290 @@
...
+            .snapshots
+            .changelog
+            .may_load(store, ((), U64Key::from(height)))?
+            .is_some()
+        {
+            return Ok(());
+        }
+        // otherwise, store the previous value
+        let old = self.primary.may_load(store)?;
+        self.snapshots
```
ethanfrey, line 60: Could we expose a helper on `self.snapshots` to do this?

`self.snapshots.write_changelog(store, (): K, height: u64, old: Option<T>)`

It makes the usage clearer

packages/storage-plus/src/snapshot_item.rs
```diff
@@ -0,0 +1,290 @@
...
+}
+
+impl<'a, T> SnapshotItem<'a, T>
+where
+    T: Serialize + DeserializeOwned + Clone,
+{
+    /// load old value and store changelog
+    fn write_change(&self, store: &mut dyn Storage, height: u64) -> StdResult<()> {
+        // if there is already data in the changelog for this block, do not write more
+        if self
```
ethanfrey, line 50: likewise, a helper here, like:

`self.snapshots.has_changelog(store, (), height)` would be nice.

ethanfrey: And maybe expose `write_change` on `self.snapshots`, as this same basic logic is used for Map

packages/storage-plus/src/snapshot_item.rs
```diff
@@ -0,0 +1,290 @@
...
+    pub fn may_load(&self, store: &dyn Storage) -> StdResult<Option<T>> {
+        self.primary.may_load(store)
+    }
+
+    // may_load_at_height reads historical data from given checkpoints.
+    // Only returns `Ok` if we have the data to be able to give the correct answer
+    // (Strategy::EveryBlock or Strategy::Selected and h is registered as checkpoint)
+    //
+    // If there is no checkpoint for that height, then we return StdError::NotFound
+    pub fn may_load_at_height(&self, store: &dyn Storage, height: u64) -> StdResult<Option<T>> {
```
ethanfrey, line 96: This also looks a lot like the `snapshot_map` implementation and most of that could be moved into `Snapshot` (sure, we pass a placeholder (), but saves code duplication of the most important logic.

packages/storage-plus/src/snapshot_map.rs
```diff
@@ -0,0 +1,386 @@
...
+    pub fn prefix(&self, p: K::Prefix) -> Prefix<T> {
+        self.primary.prefix(p)
+    }
+
+    pub fn sub_prefix(&self, p: K::SubPrefix) -> Prefix<T> {
+        self.primary.sub_prefix(p)
+    }
+
+    /// load old value and store changelog
+    fn write_change(&self, store: &mut dyn Storage, k: K, height: u64) -> StdResult<()> {
```
ethanfrey, line 70: I would refactor this and `may_load_at_height` to be in Snapshot, as I mentioned under Item

ethanfrey (commented):
Very nice design and well done pulling out the Snapshot.

I propose 2 more methods to put into Snapshot.
And it would be nice to have a few basic unit tests in snapshot.rs (I trust it as it works in higher level, but to demo the API and hit a few edge cases, it would be nice).

Other than that, good to merge IMO

packages/storage-plus/src/snapshot_item.rs
maurolacy: Thank you! I like it too.

ethanfrey:
I just checked out the code and read it. It looked good. No need to break it up.

Just saw these 2 non-trivial methods I figured could move into snapshot, then they all look great.

Moving the files and even adding more unit tests could be done in a separate PR. Happy to have this.

packages/storage-plus/src/snapshot_item.rs
maurolacy: This was non trivial to refactor, because of the coupling with the `primary` map.

Did something that works using nested `Option`. Passing `primary` to `snapshots` is worse I think.

packages/storage-plus/src/snapshot_map.rs
maurolacy: Here again, the need for `primary` makes this unelegant to refactor. I would leave the current version.

maurolacy:
Please take a look. I would merge  this as it is, and then do a follow-up with the paths refactoring, and adding some `Snapshot` tests.

packages/storage-plus/src/snapshot_item.rs
```diff
@@ -57,9 +57,7 @@ where
         }
         // otherwise, store the previous value
         let old = self.primary.may_load(store)?;
-        self.snapshots
-            .changelog
-            .save(store, ((), U64Key::from(height)), &ChangeSet { old })
+        self.snapshots.write_changelog(store, (), height, old)
```
ethanfrey, line 54: nice

packages/storage-plus/src/snapshot_item.rs
```diff
@@ -47,12 +47,7 @@ where
     /// load old value and store changelog
     fn write_change(&self, store: &mut dyn Storage, height: u64) -> StdResult<()> {
         // if there is already data in the changelog for this block, do not write more
-        if self
-            .snapshots
-            .changelog
-            .may_load(store, ((), U64Key::from(height)))?
-            .is_some()
-        {
+        if self.snapshots.has_changelog(store, (), height)? {
```
ethanfrey, line 49: much cleaner. thanks

packages/storage-plus/src/snapshot.rs
```diff
@@ -120,6 +120,36 @@ where
...
+            .changelog
+            .prefix(key)
+            .range(store, Some(start), None, Order::Ascending)
+            .next();
+
+        if let Some(r) = first {
+            // if we found a match, return this last one
+            r.map(|(_, v)| Some(v.old))
+        } else {
+            Ok(None)
```
ethanfrey, line 150: Nice decision here

packages/storage-plus/src/snapshot_item.rs
```diff
@@ -81,26 +80,11 @@ where
...
-            .changelog
-            .range(store, Some(start), None, Order::Ascending)
-            .next();
-
-        if let Some(r) = first {
-            // if we found a match, return this last one
-            r.map(|(_, v)| v.old)
+        let snapshot = self.snapshots.may_load_at_height(store, (), height)?;
+
+        if let Some(r) = snapshot {
```
ethanfrey, line 86: Yes, this looks nice here, keeping the primary load in this function, and checking an option to switch.

ethanfrey (approved):
Very nice. Thanks for this cleanup. Will merge

Issue #362, opened 2021-07-29
Refactor `multilib::wasm::WasmKeeper` tests to use `mulitest::wasm::Wasm` instead of internal functions


hashedone:
Some `multi-lib` entities are unit tested on their internal levels, instead of their actual exposed API, which could be improved to capture additional orchestration logic, and avoid depending on internals stability.

- [ ] `mutlitest::wasm::WasmKeeper` should use `multitest::wasm::Wasm` in UTs

ethanfrey:
I think we can depend on WamsKeeper, as we have generic in App/Router and do call them sometimes.

I do agree we can avoid calling private methods and try only pub ones. Any way to get the compiler to enforce that?

ethanfrey:
Also find this as not urgent... but a nice cleanup

PR #410, opened 2021-09-09
Removing dyn from Router

### LINKED ISSUE #404: Remove use of dyn in multitest Router
for Storage, Api, Querier, we should stick with dyn, as that is what we pass into Deps and we have no real need for more details. For the other, it can be useful to get the original types when setting up test code.

I am talking about:

```rust
pub struct Router<ExecC, QueryC> {
    pub(crate) wasm: Box<dyn Wasm<ExecC, QueryC>>,
    pub(crate) bank: Box<dyn Bank>,
    pub(crate) custom: Box<dyn CustomHandler<ExecC, QueryC>>,
}
```

https://github.com/CosmWasm/cw-plus/blob/mt-msg-handling/packages/multi-test/src/app.rs#L254-L258

I would propose something like the follow:

```rust
pub struct Router<ExecC, QueryC, W: Wasm<ExecC, QueryC>, B: Bank, C: CustomHandler<ExecC, QueryC>> {
    pub(crate) wasm: Box<W>,
    pub(crate) bank: Box<B>,
    pub(crate) custom: Box<C>,
}
```

It would require some changes to the builder pattern (but we have a similar setup in ContractWrapper where we change types during the builder.

hashedone:
Closes #404 

This is basically first iteration, however it makes maintaining App difficult - it requires almost always providing types for router/app which makes it unusable. I have ideas how to work it around, but still in progress.

Also I would like to remove dyn from entire App in the process (from internals, not interfaces) for consistency.

As a minor upgrade I want to include here is to provide helper proxy traits to make bounds way simpler (and don't have everywhere repeated the same compliacted bounds).

### DIFF
--- a/packages/multi-test/src/app.rs
+++ b/packages/multi-test/src/app.rs
@@ -27,18 +27,19 @@ pub fn next_block(block: &mut BlockInfo) {
 
 /// Type alias for default build `App` to make its storing simpler in typical scenario
 pub type BasicApp<ExecC = Empty, QueryC = Empty> =
-    App<BankKeeper, PanickingCustomHandler<ExecC, QueryC>, WasmKeeper<ExecC, QueryC>>;
+    App<BankKeeper, MockApi, PanickingCustomHandler<ExecC, QueryC>, WasmKeeper<ExecC, QueryC>>;
 
 /// Router is a persisted state. You can query this.
 /// Execution generally happens on the RouterCache, which then can be atomically committed or rolled back.
 /// We offer .execute() as a wrapper around cache, execute, commit/rollback process.
 pub struct App<
     Bank = BankKeeper,
+    Api = MockApi,
     Custom = PanickingCustomHandler<Empty, Empty>,
     Wasm = WasmKeeper<Empty, Empty>,
 > {
     router: Router<Bank, Custom, Wasm>,
-    api: Box<dyn Api>,
+    api: Api,
     storage: Box<dyn Storage>,
     block: BlockInfo,
 }
@@ -60,27 +61,29 @@ where
     AppBuilder::new_custom().build()
 }
 
-impl<BankT, CustomT, WasmT> Querier for App<BankT, CustomT, WasmT>
+impl<BankT, ApiT, CustomT, WasmT> Querier for App<BankT, ApiT, CustomT, WasmT>
 where
     CustomT::ExecC: Clone + fmt::Debug + PartialEq + JsonSchema + DeserializeOwned + 'static,
     CustomT::QueryC: CustomQuery + DeserializeOwned + 'static,
     WasmT: Wasm<CustomT::ExecC, CustomT::QueryC>,
     BankT: Bank,
+    ApiT: Api,
     CustomT: CustomHandler,
 {
     fn raw_query(&self, bin_request: &[u8]) -> QuerierResult {
         self.router
-            .querier(&*self.api, &*self.storage, &self.block)
+            .querier(&self.api, &*self.storage, &self.block)
             .raw_query(bin_request)
     }
 }
 
-impl<BankT, CustomT, WasmT> Executor<CustomT::ExecC> for App<BankT, CustomT, WasmT>
+impl<BankT, ApiT, CustomT, WasmT> Executor<CustomT::ExecC> for App<BankT, ApiT, CustomT, WasmT>
 where
     CustomT::ExecC: Clone + fmt::Debug + PartialEq + JsonSchema + DeserializeOwned + 'static,
     CustomT::QueryC: CustomQuery + DeserializeOwned + 'static,
     WasmT: Wasm<CustomT::ExecC, CustomT::QueryC>,
     BankT: Bank,
+    ApiT: Api,
     CustomT: CustomHandler,
 {
     fn execute(&mut self, sender: Addr, msg: CosmosMsg<CustomT::ExecC>) -> AnyResult<AppResponse> {
@@ -91,22 +94,24 @@ where
 }
 
 /// Utility to build App in stages. If particular items wont be set, defaults would be used
-pub struct AppBuilder<Bank, Custom, Wasm> {
+pub struct AppBuilder<Bank, Api, Custom, Wasm> {
     wasm: Wasm,
     bank: Bank,
-    api: Option<Box<dyn Api>>,
+    api: Api,
     storage: Option<Box<dyn Storage>>,
     custom: Custom,
     block: Option<BlockInfo>,
 }
 
-impl AppBuilder<BankKeeper, PanickingCustomHandler<Empty, Empty>, WasmKeeper<Empty, Empty>> {
+impl
+    AppBuilder<BankKeeper, MockApi, PanickingCustomHandler<Empty, Empty>, WasmKeeper<Empty, Empty>>
+{
     /// Creates builder with default components working with empty exec and query messages.
     pub fn new() -> Self {
         AppBuilder {
             wasm: WasmKeeper::new(),
             bank: BankKeeper::new(),
-            api: None,
+            api: MockApi::default(),
             storage: None,
             custom: PanickingCustomHandler::new(),
             block: None,
@@ -115,7 +120,12 @@ impl AppBuilder<BankKeeper, PanickingCustomHandler<Empty, Empty>, WasmKeeper<Emp
 }
 
 impl<ExecC, QueryC>
-    AppBuilder<BankKeeper, PanickingCustomHandler<ExecC, QueryC>, WasmKeeper<ExecC, QueryC>>
+    AppBuilder<
+        BankKeeper,
+        MockApi,
+        PanickingCustomHandler<ExecC, QueryC>,
+        WasmKeeper<ExecC, QueryC>,
+    >
 where
     ExecC: Debug + Clone + PartialEq + JsonSchema + DeserializeOwned + 'static,
     QueryC: Debug + CustomQuery + DeserializeOwned + 'static,
@@ -126,7 +136,7 @@ where
         AppBuilder {
             wasm: WasmKeeper::new(),
             bank: BankKeeper::new(),
-            api: None,
+            api: MockApi::default(),
             storage: None,
             custom: PanickingCustomHandler::new(),
             block: None,
@@ -134,7 +144,7 @@ where
     }
 }
 
-impl<BankT, CustomT, WasmT> AppBuilder<BankT, CustomT, WasmT> {
+impl<BankT, ApiT, CustomT, WasmT> AppBuilder<BankT, ApiT, CustomT, WasmT> {
     /// Overwrites default wasm executor.
     ///
     /// At this point it is needed that new wasm implements some `Wasm` trait, but it doesn't need
@@ -147,7 +157,7 @@ impl<BankT, CustomT, WasmT> AppBuilder<BankT, CustomT, WasmT> {
     pub fn with_wasm<B, C: CustomHandler, NewWasm: Wasm<C::ExecC, C::QueryC>>(
         self,
         wasm: NewWasm,
-    ) -> AppBuilder<BankT, CustomT, NewWasm> {
+    ) -> AppBuilder<BankT, ApiT, CustomT, NewWasm> {
         let AppBuilder {
             bank,
             api,
@@ -168,7 +178,10 @@ impl<BankT, CustomT, WasmT> AppBuilder<BankT, CustomT, WasmT> {
     }
 
     /// Overwrites default bank interface
-    pub fn with_bank<NewBank: Bank>(self, bank: NewBank) -> AppBuilder<NewBank, CustomT, WasmT> {
+    pub fn with_bank<NewBank: Bank>(
+        self,
+        bank: NewBank,
+    ) -> AppBuilder<NewBank, ApiT, CustomT, WasmT> {
         let AppBuilder {
             wasm,
             api,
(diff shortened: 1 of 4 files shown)


packages/multi-test/src/custom_handler.rs
```diff
@@ -1,50 +1,72 @@
...
 use std::rc::Rc;
 
 use crate::AppResponse;
 
 /// Custom message handler trait. Implementor of this trait is mocking environment behavior on
 /// given custom message.
-pub trait CustomHandler<ExecC = Empty, QueryC = Empty> {
+pub trait CustomHandler {
+    /// Custom exec message for this handler
+    type ExecC;
```
ethanfrey, line 16: Huh, I have seen such in eg. Iterator, but I am not sure how they are different in practice than generic traits.

This is needed for the Sudoku?

packages/multi-test/src/wasm.rs
```diff
@@ -61,15 +59,19 @@ where
     ) -> AnyResult<Binary>;
 
     /// Handles all WasmMsg messages
-    fn execute(
+    fn execute<BankT, CustomT>(
```
ethanfrey, line 62: Shouldn't we enforce that CustomT makes sense? Such that CustomT is related to ExecC and QueryC?

ethanfrey: Ah, I just saw the where clause. Perfect.

packages/multi-test/src/wasm.rs
```diff
@@ -78,38 +80,42 @@ where
...
 
-pub struct WasmKeeper<ExecC = Empty, QueryC = Empty> {
+pub struct WasmKeeper<ExecC, QueryC> {
     /// code is in-memory lookup that stands in for wasm code
     /// this can only be edited on the WasmRouter, and just read in caches
     codes: HashMap<usize, Box<dyn Contract<ExecC>>>,
-    /// Just marker to make type elision fork when using it as `Wasm` trait
-    _q: std::marker::PhantomData<QueryC>,
+    /// Just markers to make type elision fork when using it as `Wasm` trait
+    _p: std::marker::PhantomData<(ExecC, QueryC)>,
```
ethanfrey, line 103: Rust question:

Why we need them both here when `ExecC` is fixed by codes type already?

packages/multi-test/src/app.rs
```diff
@@ -19,153 +19,345 @@ use crate::wasm::{ContractData, Wasm, WasmKeeper};
...
 
 pub fn next_block(block: &mut BlockInfo) {
     block.time = block.time.plus_seconds(5);
     block.height += 1;
 }
 
+/// Type alias for default build `App` to make its storing simpler in typical scenario
+pub type BasicApp<ExecC = Empty, QueryC = Empty> = App<
+    BankKeeper,
+    MockApi,
```
ethanfrey, line 31: I would say MockApi and MockStorage could remain dyn if this helps simplify.

However, in practice is seems Wasm and custom types are the hardest part to deal with.

packages/multi-test/src/app.rs
```diff
@@ -19,153 +19,345 @@ use crate::wasm::{ContractData, Wasm, WasmKeeper};
...
-    block: BlockInfo,
+    AppBuilder::new_custom().build()
 }
 
-impl<ExecC, QueryC> Querier for App<ExecC, QueryC>
+impl<BankT, ApiT, StorageT, CustomT, WasmT> Querier for App<BankT, ApiT, StorageT, CustomT, WasmT>
 where
-    ExecC: Clone + fmt::Debug + PartialEq + JsonSchema + 'static,
-    QueryC: CustomQuery + DeserializeOwned,
+    CustomT::ExecC: Clone + fmt::Debug + PartialEq + JsonSchema + DeserializeOwned + 'static,
```
ethanfrey, line 78: Eyes starting to hurt....

You are a true soldier to make it to the end 🎖️

packages/multi-test/src/app.rs
```diff
@@ -835,14 +1035,17 @@ mod test {
         // TODO: check error?
     }
 
-    fn query_router<ExecC, QueryC>(
-        router: &Router<ExecC, QueryC>,
+    fn query_router<BankT, CustomT, WasmT>(
+        router: &Router<BankT, CustomT, WasmT>,
         api: &dyn Api,
```
ethanfrey, line 1040: Huh, there is still &dyn Api, but you fix the type in App?

Okay, so Router doesn't care about it as an argument, but App does when storing it?

packages/multi-test/src/app.rs
```diff
@@ -19,153 +19,345 @@ use crate::wasm::{ContractData, Wasm, WasmKeeper};
...
+    /// Overwrites default wasm executor.
+    ///
+    /// At this point it is needed that new wasm implements some `Wasm` trait, but it doesn't need
+    /// to be bound to Bank or Custom yet - as those may change. The cross-components validation is
+    /// done on final building.
+    ///
+    /// Also it is possible to completely abandon trait bounding here which would not be bad idea,
+    /// however it might make the message on build creepy in many cases, so as for properly build
+    /// `App` we always want `Wasm` to be `Wasm`, some checks are done early.
+    pub fn with_wasm<B, C: CustomHandler, NewWasm: Wasm<C::ExecC, C::QueryC>>(
```
ethanfrey, line 193: Huh.
I don't like so much disconnecting the Wasm<C::ExecC, C::QueryC> from they types used in CustomT (which is already defined).

So, you need to specify the proper (final) ExecC/QueryC when creating the (empty) builder, then can only install a new WasmKeeper that uses the same custom types.

That doesn't seem to be enforced here. But maybe I am just not following this.

packages/multi-test/src/app.rs
```diff
@@ -19,153 +19,345 @@ use crate::wasm::{ContractData, Wasm, WasmKeeper};
...
-        self
+    ///
+    /// At this point it is needed that new custom implements some `CustomHandler` trait, but it doesn't need
+    /// to be bound to ExecC or QueryC yet - as those may change. The cross-components validation is
+    /// done on final building.
+    ///
+    /// Also it is possible to completely abandon trait bounding here which would not be bad idea,
+    /// however it might make the message on build creepy in many cases, so as for properly build
+    /// `App` we always want `Wasm` to be `Wasm`, some checks are done early.
+    pub fn with_custom<NewCustom: CustomHandler>(
```
ethanfrey, line 297: Same as above. There is an implicit connection between the ExecC/QueryC types in Query and Wasm and it would be great if the builder enforced they are compatible.

It seems this will then cause some issue later on when trying to use the resulting app that some method is not available (as impl requires a match). But ideally the compile error would appear on the line that caused the issue (this one when setting an incompatible Custom handler)

ethanfrey (commented):
First pass on this.

My impression is that you spent quite some time fighting with the Rust compiler, but the types do seem to fit together somehow.

I raised a few points about stronger connection between the WasmHandler and the CustomHandler types when building, as these are required to be aligned. Otherwise, it looks good to me.

The biggest open question is how this affects usage in the contracts. Hopefully, we don't need to define 10 generics to use it and can just eg. use `BasicApp` or such. It seems the contracts were not updated yet, but that would be my only other take... to ensure the API still makes sense (is easy to work with) when using it. That probably could be fixed with a few helper types/functions that make more assumptions for you (as you seem to have started doing)

ethanfrey:
Does this also close https://github.com/CosmWasm/cw-plus/issues/362 in the refactor?
Or is that a separate issue still?

ethanfrey:
I checked this out and ran the contract tests that use multi-test and was amazed they worked without changes.

Such an overhaul of types and it didn't break the typical API usage. 👍

packages/multi-test/src/custom_handler.rs
hashedone: So difference is on the one hand subtle, on the other - fundamental.

When you have a generic trait, then it basically means, that this trait can be implemented on type multiple times, for different generics argument. So if `CustomHandler` is generic trait, you can make your very same type to be handler for multiple types. But what it also means is, that when you call a function on this trait, the if the function doesn't have all its generics in signature, type deduction my fail. Consider:

```rust
// Assume MyCustomHandler implements:
// * `CustomHandler<Empty, Empty>`
// * `CustomHandler<CustomMsg, Empty>`
let handler = MyCustomHandler::new();
// Now: which to call? Both `CustomHandler` implementations are valid here
handler.query(&api, &storage, &block, Empty);
```
 And the problem is even worse than hinting for type in normal types - as when for types you basically typehint once and the type is known for whole context, for traits it is not true. Even if here you call `query` from particular implementation, it doesn't mean that in next line, you don't want to call the other `query` - therefore you need to typehint all your calls and it looks like: `CustomHandler::<Empty, _>::query(&handler, &storage, &block, Empty)`.

On the other hand, associated types are induced from trait - so there is always one trait implementation for type, and types are delivered by this implementation. It is more limiting, but also way simpler for typesystem - types don't need to be ever elided. You see I often put type bounds on those types (`ExecT = ExecT`), but what it means is "if you already check that `CustomHandler` is implemented, make sure, that it is the same that other generic type" (which can also help type elision in case `ExecT` cannot be elided from other context, which is basically often case here).

Perfect example of difference between two approaches are `AsRef` and `Deref` traits which both give returns a borrow to underlying data, but they are different - `AsRef` is generic, `Deref` has associated `Deref::Target`. And they differ on semantics - `AsRef` means, that type can be used in place of borrow to other type (which makes sens to have multiple implementations - `String` can be used both as `&String` and `&str`, and there are more complex scenario, when type can have even more `AsRef impls). `Deref` is designing for dereferencing and it means that it always may be dereferenced to single type, and it never makes sense to have multiple implementation of it. This is a reason, why you always can use `*` on smart pointer (which uses deref internally), but to use `.as_ref` you always have to specify type you are wan to convert to (either directly by type hint, or by using in on unequivocal context).

Now an ancient question - when to use generic traits and when to use associated types? I often see rule of thumb: use generics unless they make you troubles, try associated when generics doesn't work. It is based on idea, that generics are somehow easier to think about I think, but I don't like this rule. My rule is: use generics if it makes sense to implement this trait with different types, otherwise use associated types to avoid type elision problems. Either way - in this very case, the generic implementation I took on first take proofed itself being problematic - because if in your whole test you chose to use only queries or execs, then type elision went crazy and spam of type hints were needed.

packages/multi-test/src/wasm.rs
hashedone: We don't overlook (I changed this type like all the times in the universe).

packages/multi-test/src/app.rs
hashedone: `MockApi` and `MockStorage` took me maybe 2h in total. It was straightforward substitution in all places, no wired dependencies.

packages/multi-test/src/app.rs
hashedone: This actually can be simplified - even if simply introducing marker trait:
```rust
trait CustomExecMsg: Clone + fmt::Debug + PartialEq + JsonSchema + DeserializeOwned + 'static {}

impl<C> CustomExecMsg for C where C: Clone + fmt::Debug + PartialEq + JsonSchema + DeserializeOwned + 'static {}
```

And then it can be used in all those long bounds. Also it is a trick for reducing all those app-traits to one, so instead of:
```rust
where
  BankT: Bank,
  CustomT: CustomHandler<ExecC = ExecC, QueryC = QueryC>,
  WasmT: Wasm<ExecC, QueryC>,
```

you would have:

```rust
where
  Self: App<BankT, CustomT, WasmT>
```

By extracting everything from `impl App` to separate trait `App` (and renaming `App` to some `AppImpl`) - bounds would be only in place of this trait impl. Same could be done for router (which is often bound in `Wasm`).

However I didn't want to include it in this review - it is big enough, and I didn't want to risk ruining something with those things.

packages/multi-test/src/app.rs
hashedone: Yes, you are wrong. It took me the most time to go around this. The only solution I found which actually connects wasm with other types requires, and plays well with building app is requiring GATs, so there are to reasons I didn't went for it:
* It requires GATs so it is an opposite of being straightforward
* It requires GATs, and GATs are not yet in stable.

I would love to find good solution for properly connecting those two. And I am sure it is one. I think the problem is, that those types are so entangled that gradually simplifying bounds could help, but I don't see the solution yet.

hashedone: Basically the problem is, that you have to disconnect those types here. If you wont, then how would you even build the app for non-standard custom msgs? If you set custom handler first, it fails - your new custom handler types doesn't match wasm types. Same if you set wasm first, the again, you are in the same place (unless wasm can handle both types which is true for probably most cases, but in some edge it may not - at the and wasm may be substituted with custom one). The relation between those types is defined in build, as then both components have to be final.

packages/multi-test/src/app.rs
hashedone: And as before, it is in `build` by bound `WasmT: Wasm<CustomT::ExecC, CustomT::QueryC>`.

hashedone:
I don't think it closes #362 - I didn't checked tests. And I am not sure if all tests would not need change (I think some if not all contracts are relating to multitest via cargo, not path), but I am sure there should not be drastic changes. `App` is still the same type (thanks to default - unless you use custom messages, when you should switch to `BasicApp`, but not the case for tests we already have). One place where things may have is building App, but it is minimal change (and basically comes to change `mock_app` implementation to `App::new()`).

packages/multi-test/src/custom_handler.rs
ethanfrey: Thanks for the explanation

packages/multi-test/src/app.rs
ethanfrey: Let's discuss this in a future issue. It is good (and possibly complete correct) as it is. Maybe I try a small PR later on to address this, but definitely not a blocker to merge.

ethanfrey (approved):
Good work.

Please rebase and merge it in.

PR #412, opened 2021-09-10
Add query helpers to Item and Map and use them in cw4 helpers

### LINKED ISSUE #376: Idea: item query helper storage helper
## Background 

Cross contract configuration `Item` queries is a common execution. This is usually done using `SmartQuery` which costs quite gas for retrieving simple object from storage. Instead `RawQuery` can be used for lower costs.
 
## Proposal

Develop new generic storage helper that generates raw query wrapper for given data struct.

## Usage example
On source contract
```rust
pub const CONTRACT_CONFIG: CONFIG<State> = CONFIG::new("config");
```

On external contract 
```rust
let EXT_CONFIG = CONTRACT_CONFIG::query(&deps.querier, external_contract_addr)?;
```

`CONTRACT_CONFIG::query(&deps.querier, external_contract_addr)`  
raw query with **config** key internally, and parses `State` struct.

Please look at cw4 helpers.rs for member query to see an example of reading remote state with raw query https://github.com/CosmWasm/cw-plus/blob/main/packages/cw4/src/helpers.rs#L87-L112

ethanfrey:
Closes #376 
 
And also adds this functionality to Map. I didn't add any unit tests to this, but updated cw4 helpers to use this, which are covered in `cw3-flex-multitest` and more. (A real use case).

### DIFF
--- a/packages/cw4/src/helpers.rs
+++ b/packages/cw4/src/helpers.rs
@@ -2,15 +2,15 @@ use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
 use cosmwasm_std::{
-    from_slice, to_binary, to_vec, Addr, Binary, ContractResult, CosmosMsg, Empty, QuerierWrapper,
-    QueryRequest, StdError, StdResult, SystemResult, WasmMsg, WasmQuery,
+    to_binary, Addr, CosmosMsg, Empty, QuerierWrapper, QueryRequest, StdResult, WasmMsg, WasmQuery,
 };
 
 use crate::msg::Cw4ExecuteMsg;
 use crate::query::HooksResponse;
 use crate::{
-    member_key, AdminResponse, Cw4QueryMsg, Member, MemberListResponse, MemberResponse, TOTAL_KEY,
+    AdminResponse, Cw4QueryMsg, Member, MemberListResponse, MemberResponse, MEMBERS_KEY, TOTAL_KEY,
 };
+use cw_storage_plus::{Item, Map};
 
 /// Cw4Contract is a wrapper around Addr that provides a lot of helpers
 /// for working with cw4 contracts
@@ -62,14 +62,6 @@ impl Cw4Contract {
         .into())
     }
 
-    fn encode_raw_query<T: Into<Binary>>(&self, key: T) -> QueryRequest<Empty> {
-        WasmQuery::Raw {
-            contract_addr: self.addr().into(),
-            key: key.into(),
-        }
-        .into()
-    }
-
     /// Show the hooks
     pub fn hooks(&self, querier: &QuerierWrapper) -> StdResult<Vec<String>> {
         let query = self.encode_smart_query(Cw4QueryMsg::Hooks {})?;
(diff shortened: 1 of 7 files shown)


packages/cw4/src/helpers.rs
```diff
@@ -79,36 +71,14 @@ impl Cw4Contract {
 
     /// Read the total weight
     pub fn total_weight(&self, querier: &QuerierWrapper) -> StdResult<u64> {
-        let query = self.encode_raw_query(TOTAL_KEY.as_bytes());
-        querier.query(&query)
+        const TOTAL: Item<u64> = Item::new(TOTAL_KEY);
+        TOTAL.query(querier, self.addr())
```
ueco-jb, line 75: Is there a point in creating local `const TOTAL` when similar
`let total = Item::new(TOTAL_KEY);`
is immutable as well? 
Also IMO function names are here quite self-explanatory, so you could go even shorter:
`Item::new(TOTAL_KEY).query(querier, self.addr())`

packages/cw4/src/helpers.rs
```diff
@@ -79,36 +71,14 @@ impl Cw4Contract {
...
-                // This is the only place we customize
-                if value.is_empty() {
-                    Ok(None)
-                } else {
-                    from_slice(&value)
-                }
-            }
-        }
+        const MEMBERS: Map<&Addr, u64> = Map::new(MEMBERS_KEY);
+        MEMBERS.query(querier, self.addr(), addr)
```
ueco-jb, line 81: As above.

ueco-jb (approved):

packages/cw4/src/helpers.rs
ethanfrey: We need to define the type Item returns, so: 

`Item::new::<u64>(TOTAL_KEY).query(querier, self.addr())`

But wait, I guess this is set in the wrapping function signature. Good idea

packages/cw4/src/helpers.rs
ethanfrey: Yup, this works. Nice simplification

packages/storage-plus/src/map.rs
```diff
@@ -83,6 +86,41 @@ where
     {
         self.key(k).update(store, action)
     }
+
+    /// If you import the proper Map from the remote contract, this will let you read the data
+    /// from a remote contract in a type-safe way using WasmQuery::RawQuery
+    pub fn query(
```
maurolacy, line 92: I'm OK with this but, wouldn't be better in terms of encapsulation, specificity and whatnot, to have a derived struct from `Map` for implementing this? Someting like `QueryMap`?

Perhaps better (or at least, simpler): Achieve this with a helper (`query_map`) function? Passing the map, extracting the namespace, building the query, etc. Helper can live in the same module / crate than the struct; `pub(module)` can be used for accessibility if needed.

maurolacy (approved):

packages/storage-plus/src/map.rs
ethanfrey: I can make another struct. But curious as to why that would be cleaner.

Maybe we can discuss this (Monday) on a short call and then decide.

maurolacy: > I can make another struct. But curious as to why that would be cleaner.
> 
> Maybe we can discuss this (Monday) on a short call and then decide.

Sure. I just feel that adding that to `Map`, which is such a fundamental container, is a little loaded. How many times you'd use a `Map`, vs. this `query` method?

ethanfrey: I reflected on this a bit more, and I see the option of `query_map(map: &Map<K, T>, k: K) -> StdResult<Option<T>>` as a reasonable alternative, also for item.

however, it seems to only really serve to hide it. One more import and `query_map(&BALANCES, &addr)?` over `BALANCES.query(&addr)?` is a little uglier.

I think the request from Orkun was that these be easy to use, and I would go for adding them here. I will also try to reduce the size of the code, as most of this has nothing to do with map, but overriding `QuerierWrapper.custom_query` to not parse the data (and assume it is non-empty). I will lower the size of code in these methods via a helper, but keep them exposed as methods for usability

packages/storage-plus/src/map.rs
```diff
@@ -123,6 +105,34 @@ where
     }
 }
 
+// TODO: move this to a better helpers location
```
ueco-jb, line 108: Do you have something specific in mind? Maybe create an issue along the way?

ethanfrey: I will do this before merge.
Can't merge with a TODO (gotta make an issue for it first)

PR #413, opened 2021-09-10
Add doc entry about key usage in maps

### LINKED ISSUE #295: Why you use `Addr` as keys in Maps?
Hey guys, thanks for your work! 
Have a little question :)
Why you use `Addr`(which is a String actually) as a key in Maps? [Here](https://github.com/CosmWasm/cosmwasm-plus/blob/65649f7f61ccae8504cb6cbb0c7cbd2c6a6ff3c2/contracts/cw20-base/src/state.rs#L33) for example?

[Here](https://docs.cosmwasm.com/0.14/architecture/addresses.html#optional-canonical-addresses) CosmWasm docs saying (and I agree):
```
This means that if message.signer is always the string address that signed
 the transaction and I use it to look up your account balance, if this encoding
 ever changed, then you lose access to your account. We clearly need
 a stable identifier to work with internally.
```

It is possible for address string representation to change, so ideally we all (smart contracts dev) should use `CanonicalAddr` as a key for Maps.

So, Why?

maurolacy:
Closes #295.

### DIFF
--- a/packages/storage-plus/README.md
+++ b/packages/storage-plus/README.md
@@ -188,6 +188,32 @@ fn demo() -> StdResult<()> {
 }
 ```
 
+### Key types
+
+A `Map` key can be anything that implements the `PrimaryKey` trait. There are a series of implementations of
+`PrimaryKey` already provided (see `packages/storage-plus/src/keys.rs`):
+
+ - `impl<'a> PrimaryKey<'a> for &'a [u8]`
+ - `impl<'a> PrimaryKey<'a> for &'a str`
+ - `impl<'a> PrimaryKey<'a> for Vec<u8>`
+ - `impl<'a> PrimaryKey<'a> for String`
+ - `impl<'a> PrimaryKey<'a> for Addr`
+ - `impl<'a> PrimaryKey<'a> for &'a Addr`
+ - `impl<'a, T: PrimaryKey<'a> + Prefixer<'a>, U: PrimaryKey<'a>> PrimaryKey<'a> for (T, U)`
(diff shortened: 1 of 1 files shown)


packages/storage-plus/README.md
```diff
@@ -188,6 +188,32 @@ fn demo() -> StdResult<()> {
...
+ - `impl<'a> PrimaryKey<'a> for &'a Addr`
+ - `impl<'a, T: PrimaryKey<'a> + Prefixer<'a>, U: PrimaryKey<'a>> PrimaryKey<'a> for (T, U)`
+ - `impl<'a, T: PrimaryKey<'a> + Prefixer<'a>, U: PrimaryKey<'a> + Prefixer<'a>, V: PrimaryKey<'a>> PrimaryKey<'a> for (T, U, V)`
+ - `impl<'a, T: Endian + Clone> PrimaryKey<'a> for IntKey<T>`
+
+That means that byte and string slices, byte vectors, and strings, can be conveniently used as keys.
+Moreover, some other types can be used as well, like addresses and addresses references, tuples and triples, and
+integer types.
+
+We suggest using `Addr` for keys in storage, instead of `String` or string slices. This implies doing address validation
```
ethanfrey, line 210: I would prefix this a bit:

"If the key represents and address, we suggest using &Addr..."

as I have used eg. escrow ID (string) as a key, where Addr wouldn't make sense.

also, generally &Addr is cheaper unless you need the other for eg. secondary indexes

ethanfrey (approved):
good doc.one minor comment

PR #414, opened 2021-09-10
Add CW-2981

the-frey:
Appreciate this might need some refining and/or further discussion.

Wanted essentially to have a bare minimum example for the [EIP-2981 spec](https://eips.ethereum.org/EIPS/eip-2981) implemented in CW-land. This is my attempt at that.

In essence, it's a very light wrapper/extension for CW721 compliant contracts that adds two additional query messages:
- `RoyaltyInfo` to be called when selling an NFT, to find out what royalties are owed
- `CheckRoyalties` to be called on the contract, to see if it implements CW-2981

There's an example contract implemented that adds royalty information at token mint time, although it is also possible to add that as contract metadata at instantiation time (that instantiation message format is in the package).

The main thing as I understand it here is basically the shape of the queries, that can be called when a marketplace sells an NFT. The actual implementation might differ on a per contract basis (the example contract here just does a straight percentage).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -540,6 +540,34 @@ dependencies = [
 ]
 
 [[package]]
+name = "cw2981"
+version = "0.8.1"
+dependencies = [
+ "cosmwasm-schema",
+ "cosmwasm-std",
+ "schemars",
+ "serde",
+]
+
+[[package]]
+name = "cw2981-base"
+version = "0.8.1"
+dependencies = [
+ "cosmwasm-schema",
+ "cosmwasm-std",
+ "cw-storage-plus 0.8.1",
+ "cw0 0.8.1",
+ "cw2",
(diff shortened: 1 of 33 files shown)


ethanfrey:
Thank you for submitting this and nice to see more NFT ideas ported to CosmWasm and shared.

I am a bit busy now, but it will get a review this week.

the-frey:
Cool! Like I say, anticipating it will need a bit more polish, but I'm at my desk this week so will keep an eye out for messages on this thread. Cheers!

ethanfrey:
Sorry not to get to this earlier.
A whole lot of NFT issues came up and I felt ill-equipped to deal with them.

I think it would be great if you could close this PR and open a new one with the same design on https://github.com/CosmWasm/cw-nfts That should get more comments and a faster feedback

PR #417, opened 2021-09-13
admin and hooks return Response<C> in execute_*

### LINKED ISSUE #408: Generalize controllers
`cw-controllers` are using `CosmosMsg<Empty>` (and relatives) when it comes to messaging. This is an issue, as it makes them unusable in chains which have their specific custom message (and there is no way to simply map those). The solution is simple - as those utilities are never using `Custom`, instead of returning `cosmwasm_std::Response`, they should return generic `cosmwasm_std::Response<T>`, for eg the `cw_controllers::Admin::execute_udpate_admin` may look like:

```rust
pub fn execute_update_admin<C>(...) -> Result<Response<C> { ... }
```

C should be easy to infer, as those calls are used to directly return them.

ethanfrey:
Closes #408

### DIFF
--- a/packages/controllers/src/admin.rs
+++ b/packages/controllers/src/admin.rs
@@ -1,5 +1,6 @@
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
+use std::fmt;
 use thiserror::Error;
 
 use cosmwasm_std::{attr, Addr, Deps, DepsMut, MessageInfo, Response, StdError, StdResult};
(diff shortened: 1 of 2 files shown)


packages/controllers/src/admin.rs
```diff
@@ -1,5 +1,6 @@
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
+use std::fmt;
```
ueco-jb, line 3: Nit: Personally I would add whole `std::fmt::Debug` here and second file..

ueco-jb (approved):
lgmt

packages/controllers/src/admin.rs
ethanfrey: I think I use this other places as well.
Let's change them all in another PR if this seems more clear. (feel free to make an issue)

PR #418, opened 2021-09-14
Snapshot item 2

maurolacy:
#409 follow-up. Refactors into a `snapshot` module, and adds  `Snapshot` unit tests.


packages/storage-plus/src/snapshot/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+        key: K,
+        height: u64,
+        old: Option<T>,
+    ) -> StdResult<()> {
+        self.changelog
+            .save(store, (key, U64Key::from(height)), &ChangeSet { old })
+    }
+
+    // may_load_at_height reads historical data from given checkpoints.
+    // Only returns `Ok` if we have the data to be able to give the correct answer
```
ethanfrey, line 132: This comment is incorrect.

Returns `StdError::NotFound` if we have no checkpoint (and can give no data)
Returns `Ok(None)` if there is a checkpoint, but no cached data (no changes since the checkpoint, caller should query current state)
Return `Ok(Some(x))`if there is a checkpoint and data written to changelog, returning the state at that time

packages/storage-plus/src/snapshot/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+    #[test]
+    fn should_checkpoint() {
+        let storage = MockStorage::new();
+
+        assert_eq!(NEVER.should_checkpoint(&storage, &DUMMY_KEY), Ok(false));
+        assert_eq!(EVERY.should_checkpoint(&storage, &DUMMY_KEY), Ok(true));
+        assert_eq!(SELECT.should_checkpoint(&storage, &DUMMY_KEY), Ok(false));
+    }
+
+    #[test]
```
ethanfrey, line 203: nice unit test here

packages/storage-plus/src/snapshot/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+        assert_eq!(
+            NEVER.may_load_at_height(&storage, DUMMY_KEY, 3),
+            Err(StdError::not_found("checkpoint"))
+        );
+        assert_eq!(EVERY.may_load_at_height(&storage, DUMMY_KEY, 3), Ok(None));
+        assert_eq!(SELECT.may_load_at_height(&storage, DUMMY_KEY, 3), Ok(None));
+
+        // Write a changelog at 3
+        NEVER
+            .write_changelog(&mut storage, DUMMY_KEY, 3, Some(100))
```
ethanfrey, line 313: I would store different values here, so if you mix something up in a test, it is clear which is which.

packages/storage-plus/src/snapshot/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+        assert_eq!(
+            EVERY.may_load_at_height(&storage, DUMMY_KEY, 3),
+            Ok(Some(Some(100)))
+        );
+        assert_eq!(
+            SELECT.may_load_at_height(&storage, DUMMY_KEY, 3),
+            Ok(Some(Some(100)))
+        );
+
+        // Write a changelog at 4, removing the value
```
ethanfrey, line 335: removing the value? that doesn't make sense to me

packages/storage-plus/src/snapshot/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+
+        assert_eq!(
+            NEVER.may_load_at_height(&storage, DUMMY_KEY, 3),
+            Err(StdError::not_found("checkpoint"))
+        );
+        assert_eq!(
+            EVERY.may_load_at_height(&storage, DUMMY_KEY, 3),
+            Ok(Some(Some(100)))
+        );
+        assert_eq!(
```
ethanfrey, line 330: It would also be nice to test (only with EVERY) that `may_load_at_height(2)` will return the value set at 3 (the first change after that). That is important to assert the iterator works and not just exact matches.

ethanfrey (approved):
Reorg looks good.

Minor comment on a comment text, and a small addition to test cases.
Feel free to merge once you consider these resolved.

packages/storage-plus/src/snapshot/mod.rs
maurolacy: Why not? It would be tantamount to removing the key. It's supported, so I think it must be tested. Also I like seeing `Some(None)` in the results.

PR #422, opened 2021-09-14
Simplify router args

ethanfrey:
I tried to simplify some of these complex args and self types.

But I ended up with:

```
error[E0038]: the trait `CosmosExecutor` cannot be made into an object
   --> packages/multi-test/src/wasm.rs:684:17
    |
684 |         router: &dyn CosmosExecutor<ExecC = ExecC>,
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `CosmosExecutor` cannot be made into an object
    |
note: for a trait to be "object safe" it needs to allow building a vtable to allow the call to be resolvable dynamically; for more information visit <https://doc.rust-lang.org/reference/items/traits.html#object-safety>
   --> packages/multi-test/src/app.rs:498:38
    |
497 | pub trait CosmosExecutor {
    |           -------------- this trait cannot be made into an object...
498 |     type ExecC: Clone + fmt::Debug + PartialEq + JsonSchema;
    |                                      ^^^^^^^^^ ...because it uses `Self` as a type parameter
```

I don't quite follow why it is not proper. Maybe you tried this path already and failed. Maybe you know how to fix it.

I think it will make Wasm cleaner as we add more types into Router (like Staking, etc). But maybe this approach is not possible.


ethanfrey:
If you have a solution or idea, please just push it on this branch

ethanfrey:
Update: this is compiling and passing tests.
Usage of the Router in wasm.rs is much simpler.

The biggest issue encountered was that `struct CosmosMsg<T>` requires `T: Clone + Debug + PartialEq + JsonSchema + ...` and it seems the PartialEq could not be used as a bound on the trait attribute/type. Ugh.

I demo a solution, I made a fake CosmosMsg, where the struct doesn't have any bounds on T (just the implementations). This solved the issue. We really should upstream this so as not to have multiple CosmosMsg types floating around.

Other than that, I think this does make usage of Router nicer and wasm.rs simpler. I would like some feedback

ethanfrey:
See https://github.com/CosmWasm/cosmwasm/pull/1098 to simplify the types upstream.

PR #425, opened 2021-09-15
Generic module types

### LINKED ISSUE #387: Define generic multitest module
Related to #266

Something like `Bank = Module<BankMsg, BankQuery>`.

We can have a default NotImplemented implementation that we can use to handle all the other unsupported types of CosmosMsg, but include them in the switch. All have the same interface, which allows us to easy add support for other modules (Custom, Staking, IBC) without changing everything else.

Router should include multiple modules for these main types, but `::new()` will just set as `NotImplemented`.

ethanfrey:
Closes #387 

Start work on generic modules.

- [x] Define a `Module` type generic of any Msg, Query
- [x] Implement this for Bank
- [x] Do all initialisation on App::build()
- [x] Simplify Apis, eg. using Bank = Module<BankMsg, BankQuery> alias
- [x] Contract tests pass
- [x] Use this for CustomHandler
- [x] Add Panicking implementations for Staking, Distribution

TODO in a future PR:

- Use Module for WasmHandler (this is a bit too central and too many special entry-points - we need to expose Sudo somehow)
- Clean up APIs some more (with usage)

### DIFF
--- a/contracts/cw20-escrow/src/integration_test.rs
+++ b/contracts/cw20-escrow/src/integration_test.rs
@@ -2,13 +2,9 @@
 
 use cosmwasm_std::{coins, to_binary, Addr, Empty, Uint128};
 use cw20::{Cw20Coin, Cw20Contract, Cw20ExecuteMsg};
-use cw_multi_test::{App, AppBuilder, Contract, ContractWrapper, Executor};
+use cw_multi_test::{App, Contract, ContractWrapper, Executor};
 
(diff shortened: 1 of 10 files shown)


hashedone (approved):
Seems good, except now we should implement proper staking and distribution modules, as they probably should not fail by default for wasm.

ethanfrey:
> Seems good, except now we should implement proper staking and distribution modules, as they probably should not fail by default for wasm.

Yeah, that can be another PR. In 0.10.1 or later.

Honestly, I would add that in a contract that needs it and then port it to cw-plus. But this is more important for Terra than us I believe. Now they have hooks to add it.

PR #428, opened 2021-09-17
Update changelog 2

maurolacy:
Add a couple parameters to `update_changelog.sh`, to pass a tag since when to start processing, and to auto-detect latest version tag.

Didn't find an option to replace the re-processed part, so, doing it manually.


ethanfrey (approved):
looks good, but I haven't run it.

Is this linked from the README already?

PR #429, opened 2021-09-17
Unintuitive bound handling on multiindex - failing test

hashedone:
Prove test, the simplest I managed to provide.

The test scenario logic:

Have a multi index map indexed with some arbitrary index (`&str` in the example), and mapped to some value (`u64` in the test). The map can be however indexed by value itself - the secondary multiindex is created for this case. On range call I want to take all items with secondary items up to particular value, but it won't work - as it requires to provide second part of key, which is also a part of a bound. In practice if I call range this way, I don't know the "upper bound" of the primary index (it would be basically the highest index in the map), and it makes `Bound::inclusive` unuseable with many multi-index cases. Obviously in this very test I can fix it by changing `vec![]` to `b"one".to_owned()`, or even `b"z".to_owned()`, but it is a trap, as there is no guarantee in real world, that there is no item of index `"Zumba"` which should be inlcuded in range.

Currently the only workaround I see is never use `Bound::inclusive` for multi-index, and always use `Bound::exclusive(key+1)` but it is completely unintuitive, and even possibly not easy for more complex multi-keys.

### DIFF
--- a/packages/storage-plus/src/indexed_map.rs
+++ b/packages/storage-plus/src/indexed_map.rs
@@ -697,4 +697,53 @@ mod test {
         assert_eq!(datas[0], marias[0].1);
         assert_eq!(datas[1], marias[1].1);
     }
+
+    mod inclusive_bound {
+        use super::*;
+        use crate::U64Key;
+
+        struct Indexes<'a> {
+            secondary: MultiIndex<'a, (U64Key, Vec<u8>), u64>,
+        }
+
+        impl<'a> IndexList<u64> for Indexes<'a> {
+            fn get_indexes(&'_ self) -> Box<dyn Iterator<Item = &'_ dyn Index<u64>> + '_> {
+                let v: Vec<&dyn Index<u64>> = vec![&self.secondary];
+                Box::new(v.into_iter())
+            }
+        }
+
+        #[test]
+        #[cfg(feature = "iterator")]
+        fn composite_key_failure() {
+            let indexes = Indexes {
+                secondary: MultiIndex::new(
+                    |secondary, k| (U64Key::new(*secondary), k),
+                    "test_map",
+                    "test_map__secondary",
+                ),
+            };
+            let map = IndexedMap::<&str, u64, Indexes>::new("test_map", indexes);
+            let mut store = MockStorage::new();
+
+            map.save(&mut store, "one", &1).unwrap();
+
+            let items: Vec<_> = map
+                .idx
+                .secondary
+                .range(
+                    &store,
+                    None,
+                    Some(Bound::inclusive((U64Key::new(1), vec![]).joined_key())),
+                    Order::Ascending,
+                )
+                .collect::<Result<_, _>>()
+                .unwrap();
+
+            // Strip the index from values (for simpler comparison)
+            let items: Vec<_> = items.into_iter().map(|(_, v)| v).collect();
+
+            assert_eq!(items, vec![1]);
+        }
+    }
 }


maurolacy:
No, this is correct, because you're setting  the **max** value of the full key.

`(1, "one")` comes lexicographically **after** `(1, vec![])`.

To make it work, you need to set a value greater than the full key, either inclusively or exclusively, i.e. `(1, "z")`, `(2, "")` (Or `(1, b"\xff")`, but that doesn't work (yet), because of `joined_key` limitations).

packages/storage-plus/src/indexed_map.rs
```diff
@@ -697,4 +697,53 @@ mod test {
...
+
+            map.save(&mut store, "one", &1).unwrap();
+
+            let items: Vec<_> = map
+                .idx
+                .secondary
+                .range(
+                    &store,
+                    None,
+                    Some(Bound::inclusive((U64Key::new(1), vec![]).joined_key())),
```
maurolacy, line 737:
```suggestion
                    Some(Bound::inclusive((U64Key::new(1), vec![]).joined_key())),
                    None,
```

or

```suggestion
                    None,
                    Some(Bound::inclusive((U64Key::new(2), vec![]).joined_key())),
```

are both correct. And "intuitive".

hashedone:
I think you misunderstood me - I exactly understand why it works such way. The problem is, that I find it faulty, that there is no proper tool to set exclusive upper bound for the secondary key while iterating over range. In most cases if I need this kind of upper bound I have no idea about the upper bound of primary index, as if I know - I iterate over secondary index, because I know nothing about the primary one and I don't care about it. And I cannot fix the prefix of the index - I have no knowledge about primary key at all, I want to fully disregard it.

Basically my point is: I would really appreciate atom which behaves like `MultiIndex`, but doesn't include the primary key in it.

Also the second point is: there is no usecase for using `Bound::inclusive` for multidex upper bound, because of how it behaves (you need to know your primary key anyway to properly use it, or at least its upper range), which is very confusing for someone who didn't spend time on its internal implementation.

maurolacy:
> I think you misunderstood me - I exactly understand why it works such way. The problem is, that I find it faulty, that there is no proper tool to set exclusive upper bound for the secondary key while iterating over range. In most cases if I need this kind of upper bound I have no idea about the upper bound of primary index, as if I know - I iterate over secondary index, because I know nothing about the primary one and I don't care about it. And I cannot fix the prefix of the index - I have no knowledge about primary key at all, I want to fully disregard it.
> 
> Basically my point is: I would really appreciate atom which behaves like `MultiIndex`, but doesn't include the primary key in it.

I agree it's a limited / poor implementation. I've said it above. But it's a working one (as far as I can tell). One thing is for it to be poor / limited, and the other is to be broken.

That said, I would very much like to get rid of the pk as part of the key. I tried (and failed) to do it in the past. See #211.

The pk must be used internally. But that requires using some generics magic (something like nested generics / types) that I don't know how to do, or even, if it's possible to do.

> 
> Also the second point is: there is no usecase for using `Bound::inclusive` for multidex upper bound, because of how it behaves (you need to know your primary key anyway to properly use it, or at least its upper range), which is very confusing for someone who didn't spend time on its internal implementation.

Perhaps a helper like `max_value()` / `max()` for these types would improve on that.

maurolacy:
Now that we'll have deserialized keys (see #432 ) we can probably use them in the `MultiIndex` too.

We wouldn"t handle them fully internally, but at least there will be a way to avoid dealing with their internal representation.

ethanfrey:
After reflecting on this a bit more, I agree this is an error calling the wrong function. The big hint is you using `joined_key` to generate bounds, which means you should be responsible for all the key mangling.

`range` Bounds work properly when you give actual keys you want to iterate on. This also makes sense for secondary indexes during pagination - where you provide the secondary index and the last Addr/Vec<u8> returned for that, and it can continue with the same secondary index and any following Addr/Vec<u8>.

However, this PR demonstrates that we have a missing method in the API: `prefix_range`.

If we have `(A, B)` and want to iterate over a range of `A`, then we would need to call such a function. Happy if you want to implement this. (Note the underlying Storage always considers the start as inclusive and end as exclusive - but please double-check that with how we implement the Bound -> Vec<u8> for the range queries).

My reasoning for bounds, when translating them to &[u8] for the underlying storage are: 

* None -> None
* start, Bound::inclusive(x) -> x
* start, Bound::exclusive(x) -> x, last byte incremented by 1
* end, Bound::inclusive(x) -> x, last byte incremented by 1
* end, Bound:: exclusive(x) -> x

actually, that is only for Order::Ascending. Maybe for Order::Descending these must change??

Happy if you want to try implementing such a function, so we could just:

```rust
let items: Vec<_> = map.idx.secondary.range(&store, None, Some(Bound::inclusive(U64Key::new(1))), Order::Ascending).collect();
```

ethanfrey:
> Now that we'll have deserialized keys (see #432 ) we can probably use them in the MultiIndex too.

I agree this will make things nicer for the caller and would be a good issue. But is a bit orthogonal issue to the issue with bounds.

ethanfrey:
This is resolved with #446

PR #431, opened 2021-09-17
Added implementation of timestamp key

### LINKED ISSUE #419: Implement `PrimaryKey` for `Timestamp`
It makes sense to use `Timestamp` as a key in maps in some contexts (in particular as multi-index).

hashedone:
Closes #419

### DIFF
--- a/packages/storage-plus/src/keys.rs
+++ b/packages/storage-plus/src/keys.rs
@@ -1,4 +1,4 @@
-use cosmwasm_std::Addr;
+use cosmwasm_std::{Addr, Timestamp};
 use std::marker::PhantomData;
 
 use crate::helpers::namespaces_with_key;
@@ -250,7 +250,8 @@ impl<T: Endian> From<T> for IntKey<T> {
 
 impl<T: Endian> From<Vec<u8>> for IntKey<T> {
     fn from(wrap: Vec<u8>) -> Self {
-        // TODO: assert proper length
+        assert_eq!(wrap.len(), std::mem::size_of::<T>());
+
         IntKey {
             wrapped: wrap,
             data: PhantomData,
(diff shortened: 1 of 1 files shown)


ueco-jb (approved):

packages/storage-plus/src/keys.rs
```diff
@@ -250,7 +250,8 @@ impl<T: Endian> From<T> for IntKey<T> {
 
 impl<T: Endian> From<Vec<u8>> for IntKey<T> {
     fn from(wrap: Vec<u8>) -> Self {
-        // TODO: assert proper length
+        assert_eq!(wrap.len(), std::mem::size_of::<T>());
```
maurolacy, line 253: Pretty nice. Still, `assert!`, `unwrap()` and friends are forbidden / discouraged in wasm code, as they fail silently, or obscurely.

This will have to return an `Err`, i.e. implementing `TryFrom` instead.

And continue ahead here... prhaps implementing `From<[u8; N]>` ? But; `N` will have to be dynamic, depending on `T`. I don't think that's possible... perhaps with const generics?

maurolacy (changes requested):

packages/storage-plus/src/keys.rs
ethanfrey: Agree. This is a nice and useful funcion. But let's return errors

packages/storage-plus/src/keys.rs
```diff
@@ -264,6 +265,42 @@ impl<T: Endian> From<IntKey<T>> for Vec<u8> {
...
+    }
+}
+
+impl From<Vec<u8>> for TimestampKey {
+    fn from(val: Vec<u8>) -> Self {
+        Self(val.into())
+    }
+}
+
+impl From<Timestamp> for TimestampKey {
```
ethanfrey, line 298: nice

packages/storage-plus/src/keys.rs
```diff
@@ -264,6 +265,42 @@ impl<T: Endian> From<IntKey<T>> for Vec<u8> {
...
+    }
+}
+
+impl<'a> Prefixer<'a> for TimestampKey {
+    fn prefix(&self) -> Vec<&[u8]> {
+        self.0.key()
+    }
+}
+
+impl From<Vec<u8>> for TimestampKey {
```
ethanfrey, line 292: This has the same issue as Intel from above.

I would make this a separate PR (these two functions), rest is fine to merge

ethanfrey (commented):
looks good minus the assert

packages/storage-plus/src/keys.rs
hashedone: I did the change mostly because I hate hanging "TODO" comments.

But there is an issue - it is impossible to return error in this trait impl (I mean you always may do it for `Result<IntKey<_>>`. Obviously it is a solution to check the length only in `TryFrom` implementation, but it seems very incorrect to me, as if the incomming vec is of invalid length, it is clear that something is wrong, and conversion should probably not happen at all. For me this kind of assert is exactly the same as bound checks on indexing operator (which is there, and asserts on out of bounds) - and I don't think we are avoiding using direct indexing. I understand reasoning behind not panicking, but the question is - is that better to panic, or have incorrect behavior which is completely undetectable?

@maurolacy it is possible, and even simple, as set of `IntKey` is very limited - just to implement the conversion separately for all supported keys. However I would not go this way, as it is not generic at all (I mean if you have anything which is not an array, you need to first convert it to array). I would go completely other way - implement `TryFrom<&[u8]>` for all `IntKey`s and return arror (and if pure `From` is really needed - make an assertion, as there is nothing else to do and it is still better than having incorrect behavior).

Anyway - I will remove it form this PR as it is not critical for the issue.

packages/storage-plus/src/keys.rs
```diff
@@ -250,7 +250,8 @@ impl<T: Endian> From<T> for IntKey<T> {
 
 impl<T: Endian> From<Vec<u8>> for IntKey<T> {
     fn from(wrap: Vec<u8>) -> Self {
-        // TODO: assert proper length
+        // TODO: Consider properly handling case, when `wrap` has length not conforming for the
```
ethanfrey, line 253: Fair enough. Let's do look at TryFrom at some point...

maurolacy: Solution here is in fact not to implement `From`, and implement `TryFrom` instead. That's what Rust std does.

PR #432, opened 2021-09-17
Full deserialization for `range`

### LINKED ISSUE #198: Better return values from range/prefix

maurolacy:
Closes #198.

First (working) attempt at deserializing range result keys.

TODO:

- Naming
  - [x] Proper naming for the new `range` and `Prefix` functionality. Perhaps `range_de`, `PrefixDe`?
  - [x] Proper naming for `Pair2`. Perhaps consolidating `Pair2` and `Pair` into one (with default args for both key and value). (See https://github.com/CosmWasm/cosmwasm/pull/1101)
  - [x] Better internal naming of deserialization functions, associated types, etc. Perhaps renaming `deserialize_kv` - > `deserialize_v`, and `deserialize2_kv` -> `deserialize_kv`.
  - [x] Better naming for `Deserializable`. Perhaps `DeserializeOwned` (although confusing with `serde`).
  - [x] Put `Prefix2` into its own module? Consolidate `Prefix` and `Prefix2` (default `K = Vec<u8>`) (Follow-up PR #439)?
- [x] `keys()` deserialization (`keys_de`).
- [ ] Implement `range_de` for the other iterables (`IndexedMap`, `IndexedSnapshotMap`, and the indexes). (#461)
- [x] Make `prefix_de` deserialization (i.e. suffixes deserialization) work. (Follow-up PR #434).
- [ ] Lifetimes / references. Currently only owned approach (Follow / take a look at `serde`'s approach here). (#463)
- Implement `Deserializable` for other types.
  - [x] String types.
  - [x] Integer types.
  - [x] Tuples.
  - [x] ...
- [x] Unit testing. (wip)


maurolacy:
This is now "rebased", and properly supporting the no-iterator feature.

packages/storage-plus/src/prefix.rs
```diff
@@ -6,8 +6,9 @@ use std::marker::PhantomData;
 use cosmwasm_std::{Order, Pair, StdResult, Storage};
 use std::ops::Deref;
 
+use crate::de::Deserializable;
```
ethanfrey, line 9: It is confusing to mix `crate::de::Deserializable` with `serde::de::DeserializeOwned` and `serde::Serialize`.

How about another name for `crate::de::Deserializable` like `crate::de::KeyDeserializable` (or a better name)

packages/storage-plus/src/prefix.rs
```diff
@@ -124,6 +129,96 @@ where
...
+    pub fn keys_de<'a>(
+        &self,
+        store: &'a dyn Storage,
+        min: Option<Bound>,
+        max: Option<Bound>,
+        order: Order,
+    ) -> Box<dyn Iterator<Item = StdResult<K::Output>> + 'a>
+    where
+        T: 'a,
+        K::Output: 'a,
```
ethanfrey, line 211: I wonder about this guard. Can't we just define `K::Output: 'static`, since we only deserialise into owned objects?

I mean this is true, `'static: 'a`, but it would be more clear. I guess with both K and T. If this is non-trivial to change, we can look at that in a different PR.

ethanfrey (approved):
Seems like a good implementation.

I would prefer improved names and some other details, but I think you have a few PRs coming on top that will fix that. Approving for merge, with the condition these comments are addressed in a future PR.

packages/storage-plus/src/map.rs
```diff
@@ -2,13 +2,15 @@ use serde::de::DeserializeOwned;
...
+#[cfg(feature = "iterator")]
+use crate::de::Deserializable;
 use crate::helpers::query_raw;
 #[cfg(feature = "iterator")]
 use crate::keys::Prefixer;
 use crate::keys::PrimaryKey;
 use crate::path::Path;
 #[cfg(feature = "iterator")]
-use crate::prefix::{Bound, Prefix};
+use crate::prefix::{Bound, Pair2, Prefix, Prefix2};
```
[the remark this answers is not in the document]
maurolacy: Yes. `Prefix2` is integrated into `Prefix` in #439

packages/storage-plus/src/prefix.rs
maurolacy: Good idea. I was just following the compiler's guidelines here.

packages/storage-plus/src/map.rs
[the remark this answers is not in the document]
ethanfrey: > Pair2 can be integrated into Pair, with some changes in cosmwasm_std and around.

Can you make an issue on cosmwasm and link it here? It won't block merge, but would be good to have

ethanfrey:
Please take a look at #446 
This adds type-safe ranges.

I think using `Bound<Vec<u8>>` with no connection to the Map keys. And then using `joined_keys` and such manually is just a recipe for disaster. You may be able to clean up that PR, but I think the idea of creating Bounds with some `K: PrimaryKey` or `K: Prefixer` will catch a whole lot of issues when writing code.

I know the crux of your PR is to add more deserialisation (of K), but I think while doing some of you work, and removing EmptyPrefix, you went more in the direction of people just manually appending prefixes.

packages/storage-plus/src/map.rs
```diff
@@ -108,6 +110,25 @@ where
     }
 }
 
+#[cfg(feature = "iterator")]
+impl<'a, K, T> Map<'a, K, T>
+where
+    T: Serialize + DeserializeOwned,
+    K: PrimaryKey<'a> + Deserializable,
```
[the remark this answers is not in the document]
maurolacy: It is like that already (`K::Output` is `Deserializable`). I tried to avoid requiring `K` to be `Deserializable`, but hit some compiler errors when not using `K` as output for the full key. Will revisit, to see if I can improve on it.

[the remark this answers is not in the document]
maurolacy: This requires `K` to be `Deserializable` just to be able to infer the `Output` type.

Let's revisit your idea of a `K::Owned` trait attribute in another iteration.

maurolacy:
> Please take a look at #446 This adds type-safe ranges.
> 
> I think using `Bound<Vec<u8>>` with no connection to the Map keys. And then using `joined_keys` and such manually is just a recipe for disaster. You may be able to clean up that PR, but I think the idea of creating Bounds with some `K: PrimaryKey` or `K: Prefixer` will catch a whole lot of issues when writing code.
> 
> I know the crux of your PR is to add more deserialisation (of K), but I think while doing some of you work, and removing EmptyPrefix, you went more in the direction of people just manually appending prefixes.

That's for another PR / issue: "Make `Bound`s type safe or so".

maurolacy:
Addressed remaining comments.

Not sure if you would prefer for me to integrate the other PR (#439) into this one, and then do a final thoroough review before merging.

packages/storage-plus/src/de.rs
```diff
@@ -0,0 +1,119 @@
+use crate::keys::{IntKey, TimestampKey};
+use cosmwasm_std::{Addr, StdError, StdResult};
+use std::array::TryFromSliceError;
+use std::convert::TryInto;
+
+pub trait KeyDeserialize {
```
hashedone, line 6: I wonder if it should return an error (eg. for cases where slice is of wrong length, so there is obviously something wrong). Obviously panicking is wrong, but I am not sure about silently accepting illformed data.

maurolacy: It will error, through `try_from`.

packages/storage-plus/src/de.rs
```diff
@@ -0,0 +1,119 @@
...
+
+bytes_de!(for Vec<u8>, &Vec<u8>, [u8], &[u8]);
+
+macro_rules! string_de {
+    (for $($t:ty),+) => {
+        $(impl KeyDeserialize for $t {
+            type Output = String;
+
+            fn from_slice(value: &[u8]) -> StdResult<Self::Output> {
+                // FIXME?: Use `from_utf8_unchecked` for String, &str
```
hashedone, line 40: Why unchecked? And what if for some strange reasons underlying think is not exactly utf (malformed data, some kind of attack, whatever). Iteration through couple character versus cost of loading data from storage is irrelevant.

packages/storage-plus/src/de.rs
```diff
@@ -0,0 +1,119 @@
...
+
+impl KeyDeserialize for () {
+    type Output = ();
+
+    fn from_slice(_value: &[u8]) -> StdResult<Self::Output> {
+        Ok(())
+    }
+}
+
+macro_rules! bytes_de {
```
hashedone, line 20: I am really not happy with this kind of macro. It saves you couple keystrokes, but at the end makes code way less readable (including strange error messages in some cases). Why not to implement it just for `&[u8]` and if you really need implementation for another 3 types forward it (so in case of changes, there is only one place to change)? Same for `string_de`.

packages/storage-plus/src/de.rs
```diff
@@ -0,0 +1,119 @@
...
+    type Output = (T::Output, U::Output);
+
+    fn from_slice(value: &[u8]) -> StdResult<Self::Output> {
+        let t_len = u16::from_be_bytes(
+            value[..2]
+                .try_into()
+                // FIXME: Add and use StdError try-from error From helper
+                .map_err(|err: TryFromSliceError| StdError::generic_err(err.to_string()))?,
+        ) as usize;
+        let t = T::from_slice(&value[2..2 + t_len])?;
```
hashedone, line 90: I find `&value[2..][..t_len]` way more readable, and less error prone, but it is minor thing.

packages/storage-plus/src/de.rs
maurolacy: OK. Let's stick to `from_utf8`. Will remove this comment.

packages/storage-plus/src/de.rs
```diff
@@ -0,0 +1,119 @@
+use crate::keys::{IntKey, TimestampKey};
+use cosmwasm_std::{Addr, StdError, StdResult};
+use std::array::TryFromSliceError;
+use std::convert::TryInto;
```
ueco-jb, line 4: Nit: I got the impression that we separate imports, so some newlines here would be welcome.

packages/storage-plus/src/de.rs
maurolacy: Good idea.

packages/storage-plus/src/de.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    fn from_slice(value: &[u8]) -> StdResult<Self::Output> {
+        let t_len = u16::from_be_bytes(
+            value[..2]
+                .try_into()
+                // FIXME: Add and use StdError try-from error From helper
+                .map_err(|err: TryFromSliceError| StdError::generic_err(err.to_string()))?,
+        ) as usize;
+        let t = T::from_slice(&value[2..2 + t_len])?;
+        let u = U::from_slice(&value[2 + t_len..])?;
```
hashedone, line 90: This `2` here (and in previous subslice) is technically correct. It is not so rare, that when extending serialization, bytes are shifted and it is easy to forget, while there is perfect function for this: `split_at`:
```rust
let (len, data) = value.split_at(2);
let t_len = u16::from_be_bytes(len.try_into().map_err(...)?) as usize;
let (val, tail) = data.split_at(t_len);
Ok((T::from_slice(val)?, tail))
```

packages/storage-plus/src/de.rs
```diff
@@ -0,0 +1,119 @@
...
+        ) as usize;
+        let t = T::from_slice(&value[2..2 + t_len])?;
+        let u_len = u16::from_be_bytes(
+            value[2 + t_len..4 + t_len]
+                .try_into()
+                // FIXME: Add and use StdError try-from error From helper
+                .map_err(|err: TryFromSliceError| StdError::generic_err(err.to_string()))?,
+        ) as usize;
+        let u = U::from_slice(&value[4 + t_len..4 + t_len + u_len])?;
+        let v = V::from_slice(&value[4 + t_len + u_len..])?;
```
hashedone, line 114: Basically same as previous comment - all those offsets are just asking for problems on maintenance of this.

hashedone (dismissed):
I am happy about the change. I have some comments related to maintinability - many offsets, and things which from my experience hit back when logic have to be adjusted. Basically I prefer more descriptive style (so my code tells "what it does"), than imperative ("how it does"). Rust is pretty neat in optimizing many of those things ;)

packages/storage-plus/src/de.rs
hashedone: Sorry, I removed my comment, as I reallized it before I've seen your response.

packages/storage-plus/src/keys.rs
```diff
@@ -51,9 +61,13 @@ impl<'a> PrimaryKey<'a> for &'a str {
...
 
 // use generics for combining there - so we can use &[u8], Vec<u8>, or IntKey
-impl<'a, T: PrimaryKey<'a> + Prefixer<'a>, U: PrimaryKey<'a>> PrimaryKey<'a> for (T, U) {
+impl<'a, T: PrimaryKey<'a> + Prefixer<'a> + KeyDeserialize, U: PrimaryKey<'a> + KeyDeserialize>
+    PrimaryKey<'a> for (T, U)
+{
     type Prefix = T;
     type SubPrefix = ();
+    type Suffix = U;
+    type SuperSuffix = Self;
```
ueco-jb, line 70: Could you explain to me how does this works?
This is a `impl PrimaryKey` for tuple `(T, U)`, so for `(Prefix, Suffix)` - but `SuperSuffix` type is `Self`, which means `(Prefix, Suffix)`?
So the `SuperSuffix` is actually whole key?

packages/storage-plus/src/keys.rs
```diff
@@ -9,6 +10,9 @@ pub trait PrimaryKey<'a>: Clone {
     type Prefix: Prefixer<'a>;
     type SubPrefix: Prefixer<'a>;
 
+    type Suffix: KeyDeserialize;
+    type SuperSuffix: KeyDeserialize;
```
hashedone, line 14: What are `Suffix` and `SuperSuffix`? It requires really going deep to the code and it could not if there would be solid doc coment here - we need to start using doc coments as they makes live far easier - it works with `cargo doc`, it works with ides, and it helps understanding read code.

packages/storage-plus/src/prefix.rs
```diff
@@ -153,6 +156,96 @@ where
     }
 }
 
+#[derive(Clone)]
+pub struct Prefix2<K, T>
```
hashedone, line 160: I would look fore some more descriptive name for that....

ueco-jb, line 160: Could this name be changed to something more descriptive?

packages/storage-plus/src/keys.rs
maurolacy: Good point. I agree. I know your questions is kind of rethoric, but `SuperSuffix` is the complement of `SubPrefix`, in the same way `Suffix` is the complement of `Prefix`.

ueco-jb (commented):
Implementation details are still kind of mystery to me, but I left some comments.

packages/storage-plus/src/prefix.rs
maurolacy: This is addressed in #439, where this functionality is integrated into `Prefix` directly.

I think it would be a good idea to merge that here, so that you can review everything at once before merging.

packages/storage-plus/src/keys.rs
hashedone: No, they were not rethoric at all. At the point of writing this comment I was really not sure about that, and it got mi confused. I understood it way later. I am not ironical, at least not on review - I like have as clear conversation as it is possible.

packages/storage-plus/src/prefix.rs
hashedone: Will look at this when I would finally finish my tests I try to do since yesterday ;) Anyway - if it is addressed there and it should be merged before this one, please rebase.

packages/storage-plus/src/keys.rs
maurolacy: Yes, in this case `SuperSuffix` is the whole key. To summarize, for `(T, U)`:

- `T`: Prefix. Prefix is eager. That is, it's always "one less" than the full key.
- `U`: Suffix: Suffix is the complement of prefix.
- `()`: Sub-prefix. "One less" than Prefix. This is tantamount of calling `range` over the full key in this case.
- `(T, U)`: Super-suffix: One more than suffix. Super-suffix is the complement of sub-prefix.

packages/storage-plus/src/prefix.rs
maurolacy: See #439. Will rebase / merge these recent changes into it, for ease of reviewing.

packages/storage-plus/src/keys.rs
hashedone: Good explanation - I would appreciate having it as a doc-comment.

packages/storage-plus/src/prefix.rs
maurolacy: Done.

packages/storage-plus/src/keys.rs
maurolacy: Will do.

ueco-jb: To be honest, I can follow this logic (`prefix` is "one-less", suffix is that "one-less"), but it seems rather arbitrary than intuitive for me.
When I read that first time I though that in case `ABCD` - `A` would be a prefix, `D` would be a suffix (instead of `ABC` as prefix).

maurolacy: This is for historical reasons. We can change it (`Prefix` becomes `SubPrefix`; then we'll have `SuperPrefix` and `SubSuffix` ). But still, it would be equaly arbitrary.

Anyway, this is mostly internal stuff. The only things you see as a user are the `prefix()` and `sub_prefix()` methods.

We don't store anything except for the full key, but build the partial keys using these types as references.

maurolacy: Plus, we need the `*Suffix` stuff for annotating key deserialization (sub) types.

maurolacy: https://github.com/CosmWasm/cw-plus/blob/af447a055977605fd21d5893a2194e93d1b0760e/packages/storage-plus/src/map.rs#L115-L128

See how a sub-prefix is returning a super-suffix, and a prefix is returning a suffix? That means: I fixed this part, give me the rest (deserialized).

ethanfrey:
Is this ready for a deep review?
When all this stuff is updated, ping me, and I can do one pass over this. But I don't want to while it is still a WIP

maurolacy:
> Is this ready for a deep review?

I think so, yes.

packages/storage-plus/src/de.rs
```diff
@@ -0,0 +1,171 @@
...
+
+impl KeyDeserialize for &Vec<u8> {
+    type Output = Vec<u8>;
+
+    fn from_slice(value: &[u8]) -> StdResult<Self::Output> {
+        <Vec<u8>>::from_slice(value)
+    }
+}
+
+impl KeyDeserialize for [u8] {
```
ethanfrey, line 38: Huh? Why this type and not just `&[u8]`

packages/storage-plus/src/de.rs
```diff
@@ -0,0 +1,171 @@
...
+impl KeyDeserialize for &str {
+    type Output = String;
+
+    fn from_slice(value: &[u8]) -> StdResult<Self::Output> {
+        String::from_slice(value)
+    }
+}
+
+impl KeyDeserialize for Addr {
+    type Output = String;
```
ethanfrey, line 89: Interesting choice the Output isn't Addr.

You handle the utf8 parsing for the String, but since we are reading something from storage, can't we just use `Addr::unchecked`?

packages/storage-plus/src/de.rs
```diff
@@ -0,0 +1,171 @@
+use std::array::TryFromSliceError;
+use std::convert::TryInto;
+
+use cosmwasm_std::{Addr, StdError, StdResult};
+
+use crate::keys::{IntKey, TimestampKey};
+
+pub trait KeyDeserialize {
```
ethanfrey, line 8: I like the trait, and use of `::Output`

packages/storage-plus/src/de.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+}
+
+macro_rules! integer_de {
+    (for $($t:ty),+) => {
+        $(impl KeyDeserialize for IntKey<$t> {
+            type Output = $t;
+
+            fn from_slice(value: &[u8]) -> StdResult<Self::Output> {
+                Ok(<$t>::from_be_bytes(value.try_into()
```
ethanfrey, line 110: This will fail if it is too short, right?
But not if it is too long?

Since you have a space to return errors in parsing (good), how about an explicit length check before this line?

packages/storage-plus/src/de.rs
```diff
@@ -0,0 +1,171 @@
...
+        ))
+    }
+}
+
+impl<T: KeyDeserialize, U: KeyDeserialize> KeyDeserialize for (T, U) {
+    type Output = (T::Output, U::Output);
+
+    fn from_slice(value: &[u8]) -> StdResult<Self::Output> {
+        let (len, data) = value.split_at(2);
+        let t_len = u16::from_be_bytes(
```
ethanfrey, line 138: This method `{u16,u64,...}::from_be_bytes(....)` is called 5 places in this code. It would be nice to pull it out into a helper method.

packages/storage-plus/src/de.rs
```diff
@@ -0,0 +1,171 @@
...
+            len.try_into()
+                // FIXME: Add and use StdError try-from error From helper
+                .map_err(|err: TryFromSliceError| StdError::generic_err(err.to_string()))?,
+        ) as usize;
+        let (t, u) = data.split_at(t_len);
```
ethanfrey, line 143: Nice job with split_at

packages/storage-plus/src/de.rs
```diff
@@ -0,0 +1,171 @@
...
+
+        Ok((T::from_slice(t)?, U::from_slice(u)?))
+    }
+}
+
+impl<T: KeyDeserialize, U: KeyDeserialize, V: KeyDeserialize> KeyDeserialize for (T, U, V) {
```
ethanfrey, line 149: nice

packages/storage-plus/src/indexed_map.rs
```diff
@@ -129,17 +129,17 @@ where
     }
 
     // use prefix to scan -> range
-    pub fn prefix(&self, p: K::Prefix) -> Prefix<T> {
+    pub fn prefix(&self, p: K::Prefix) -> Prefix<Vec<u8>, T> {
```
ethanfrey, line 132: Okay, this Vec<u8> is the "legacy" approach?

packages/storage-plus/src/indexed_snapshot.rs
```diff
@@ -59,7 +60,7 @@ impl<'a, K, T, I> IndexedSnapshotMap<'a, K, T, I> {
 impl<'a, K, T, I> IndexedSnapshotMap<'a, K, T, I>
 where
     T: Serialize + DeserializeOwned + Clone,
-    K: PrimaryKey<'a> + Prefixer<'a>,
+    K: PrimaryKey<'a> + Prefixer<'a> + KeyDeserialize,
```
ethanfrey, line 63: Do you need to add this for everything?

I would keep the bounds as they are, and then add a block with this extra bound for the functions that actually deserialise the key. From the diff, you aren't using `K: KeyDeserialize` yet

I had a number of longer discussions with Bart especially about only declaring bounds we really need, and not including them on the struct, but rather the impl (and using multiple impl blocks if we only need a given bound for a few methods.

packages/storage-plus/src/keys.rs
```diff
@@ -1,14 +1,36 @@
 use cosmwasm_std::{Addr, Timestamp};
 use std::marker::PhantomData;
 
+use crate::de::KeyDeserialize;
 use crate::helpers::namespaces_with_key;
 use crate::Endian;
 
-// pub trait PrimaryKey<'a>: Copy {
+/// `PrimaryKey` needs to be implemented for types that want to be a `Map` (or `Map`-like) key,
```
ethanfrey, line 8: Nice doc

packages/storage-plus/src/keys.rs
ethanfrey: I like the (new) doc comments above and would consider this resolved.
Maybe those trait types could use more docstrings?

```rust
type Suffix: KeyDeserialize;
type SuperSuffix: KeyDeserialize;
```

They make sense to me, but I am also guilty for inventing half the names.

packages/storage-plus/src/map.rs
```diff
@@ -112,6 +112,21 @@ where
...
+impl<'a, K, T> Map<'a, K, T>
+where
+    T: Serialize + DeserializeOwned,
+    K: PrimaryKey<'a>,
+{
+    pub fn sub_prefix_de(&self, p: K::SubPrefix) -> Prefix<K::SuperSuffix, T> {
+        Prefix::new(self.namespace, &p.prefix())
+    }
+
+    pub fn prefix_de(&self, p: K::Prefix) -> Prefix<K::Suffix, T> {
```
ethanfrey, line 125: And this is the good stuff I was waiting for 😄

packages/storage-plus/src/indexed_snapshot.rs
ethanfrey: Ah, I realize later that this called `deserialize_kv` and although you didn't change this code, the implementation of `deserialize_kv` has changed to require this bound. 

The above comment is most likely obsolete.

packages/storage-plus/src/map.rs
```diff
@@ -170,9 +185,41 @@ where
         'a: 'c,
     {
         let mapped =
-            namespaced_prefix_range(store, self.namespace, min, max, order).map(deserialize_kv);
+            namespaced_prefix_range(store, self.namespace, min, max, order).map(deserialize_v);
```
ethanfrey, line 188: Let's make this `prefix_range_de`. Output should be `(K::Output, T)` as below

packages/storage-plus/src/map.rs
```diff
@@ -389,6 +436,123 @@ mod test {
...
+        };
+        PEOPLE_ID.save(&mut store, U32Key::new(56), &data2).unwrap();
+
+        // let's try to iterate!
+        let all: StdResult<Vec<_>> = PEOPLE_ID
+            .range_de(&store, None, None, Order::Ascending)
+            .collect();
+        let all = all.unwrap();
+        assert_eq!(2, all.len());
+        assert_eq!(all, vec![(56, data2.clone()), (1234, data.clone())]);
```
ethanfrey, line 527: Ah, U32Key deserializes into u32. very nice for understanding.

packages/storage-plus/src/map.rs
```diff
@@ -485,22 +708,80 @@ mod test {
...
+
+        // let's try to iterate!
+        let all: StdResult<Vec<_>> = TRIPLE
+            .range_de(&store, None, None, Order::Ascending)
+            .collect();
+        let all = all.unwrap();
+        assert_eq!(4, all.len());
+        assert_eq!(
+            all,
+            vec![
```
ethanfrey, line 749: :+1:

packages/storage-plus/src/prefix.rs
```diff
@@ -65,31 +66,43 @@ impl<'a, K: Prefixer<'a>> PrefixBound<'a, K> {
...
+    _: &dyn Storage,
+    _: &[u8],
+    raw: Pair,
+) -> StdResult<(K::Output, T)> {
+    deserialize_kv::<K, T>(raw)
 }
 
 #[derive(Clone)]
-pub struct Prefix<T>
+pub struct Prefix<K = Vec<u8>, T = Vec<u8>>
```
ethanfrey, line 90: I don't see any need for the defaults. We didn't have one before/
T will never be default and if we specify T, we cannot use the K default.

Just make it <K, T>

packages/storage-plus/src/prefix.rs
```diff
@@ -99,21 +112,26 @@ where
...
+    K: KeyDeserialize,
     T: Serialize + DeserializeOwned,
 {
     pub fn new(top_name: &[u8], sub_names: &[&[u8]]) -> Self {
-        Prefix::with_deserialization_function(top_name, sub_names, &[], default_deserializer)
+        Prefix::with_deserialization_function(
+            top_name,
+            sub_names,
+            &[],
+            default_deserializer_kv::<K, T>,
```
ethanfrey, line 125: Does this actually solve the custom deserialiser issue for multi-index?

packages/storage-plus/src/prefix.rs
```diff
@@ -151,6 +170,43 @@ where
             range_with_prefix(store, &self.storage_prefix, min, max, order).map(|(k, _)| k);
         Box::new(mapped)
     }
+
+    pub fn range_de<'a>(
```
ethanfrey, line 174: There is no need for `range_de` as far as I can tell.

It is the same as `range`, right?
You just set `K = Vec<u8>` and this will work like the "legacy version"

ethanfrey (commented):
This is a big one!

First, I like the type changes - both KeyDeserializer (nicer naming) and the additions to PrimaryKey. 

I also like the addition of key information to Prefix.

Added some cleanup comments, and maybe some redundancy can be removed, but I think just one pass with some polish and this can be merged.

I think there is more than can be done to enhance it (in other PRs).
* properly handling `prefix_range_de`
* deprecating `range` and promoting `range_de` as the standard case (I would almost be in favor of just renaming them now and break everything once... `range` -> `range_raw` and `range_de` -> `range`.
* replace `KeyDeserializer::from_slice(&[u8])` with `KeyDeserializer::from_vec(Vec<u8>)`. When we call it, we can consume the Vec and save a clone. See [deserialize_kv](https://github.com/CosmWasm/cw-plus/pull/432/files#diff-b74fe4efaced9c75072d3e90271006a9952c20fb129a418deb54561d5b67ed96R22)

But no rush on those. Tackle what you can and make an issue for the items you will put into other PRs.

packages/storage-plus/src/de.rs
maurolacy: Just for completeness. I can remove it for conciseness. Along with `for str`.

packages/storage-plus/src/de.rs
maurolacy: > Interesting choice the Output isn't Addr.

This is because when we're iterating over these keys, it's usually to return these values from queries. We can change it to `Addr`. In fact, it's the logical choice.

> You handle the utf8 parsing for the String, but since we are reading something from storage, can't we just use `Addr::unchecked`?

Notice this includes an extra step: we first need to convert to `String` using `from_utf8`, and then build an `Addr` using `Addr::unchecked`. This second step is cheap, so, I'm OK with doing this change, for logical consistency.

packages/storage-plus/src/de.rs
maurolacy: Interestingly, `try_into` will err if the slice is too long too. Which is something I just checked, and didn't knew when writing this.

packages/storage-plus/src/de.rs
maurolacy: Let's thank @hashedone for that.

packages/storage-plus/src/indexed_map.rs
maurolacy: Yes. It's an internal thing, so totally warranted IMO.

packages/storage-plus/src/indexed_snapshot.rs
maurolacy: Exactly. This is because of the default deserialization we're now using. We can improve on that, by creating a series of extra constructors in `Prefix`, like `with_raw_key`. And then we can then segregate the required trait bounds. Will try to do it.

packages/storage-plus/src/prefix.rs
maurolacy: Good point. Will remove.

packages/storage-plus/src/prefix.rs
maurolacy: I would address that in a follow-up PR, yes.

packages/storage-plus/src/prefix.rs
maurolacy: Exactly.

maurolacy:
> I think there is more than can be done to enhance it (in other PRs).
> 
> * properly handling `prefix_range_de`

Yes. Will do in a follow-up.

> * deprecating `range` and promoting `range_de` as the standard case (I would almost be in favor of just renaming them now and break everything once... `range` -> `range_raw` and `range_de` -> `range`.

I would love to do that renaming. Let's do it in a follow-up, just to better handle the breaking change.

> * replace `KeyDeserializer::from_slice(&[u8])` with `KeyDeserializer::from_vec(Vec<u8>)`. When we call it, we can consume the Vec and save a clone. See [deserialize_kv](https://github.com/CosmWasm/cw-plus/pull/432/files#diff-b74fe4efaced9c75072d3e90271006a9952c20fb129a418deb54561d5b67ed96R22)

Will do in a follow-up.

> * Add `range_de` types to "Map-like" structs (Indexed, Snapshot)

Another follow-up.

> * Simplify usage in many contracts where we use `range` and custom parsing (but could now use `range_de`)

For sure. I would do this as part of the breaking change from `range` -> `range_raw`.

> 
> No need to handle that in this PR. Tackle what you can and make an issue for the items you will put into other PRs.

Good review, thanks. Will create the follow-up issues and link them here.

packages/storage-plus/src/de.rs
ethanfrey: In fact, I think Addr::unchecked should be free for String with a good optimizer

packages/storage-plus/src/de.rs
maurolacy: Hmm, not possible as is because the integer types impl of `from_be_bytes` is not associated to a trait. So, I need to be explicit when calling these types. Macros do it, but fns don't, as I need to bound the types.

A suggestion in SO is to use the `Num` trait from `num` crate, but I'm hesitant to introduce a new dep just for a trait.

I could also implement my own marker trait, and add a `from_be_bytes` in there like in `Endian`, but this also looks overkill to me. I would just leave these calls as they are. When we implement the error helper in cosmwasm-std, they will look much nicer.

packages/storage-plus/src/map.rs
maurolacy: Yes. In fact by using some macros magic we could in principle get rid of the `IntKey` trait entirely.

packages/storage-plus/src/map.rs
maurolacy: Just created the corresponding fn, for completeness.

maurolacy:
Will add unit tests for key deserialization, and I think then we can merge this.

ethanfrey:
Sounds good.

If you can get a 🟢 from the CI, I'm happy to merge it. Let's do this and 

> Replace KeyDeserializer::from_slice(&[u8]) with KeyDeserializer::from_vec(Vec<u8>). When we call it, we can consume the Vec and save a clone. See deserialize_kv. (From #464)

And update to final cosmwasm release. Then we can cut a v0.10.0

packages/storage-plus/src/de.rs
```diff
@@ -153,3 +153,143 @@ impl<T: KeyDeserialize, U: KeyDeserialize, V: KeyDeserialize> KeyDeserialize for
...
+
+    #[test]
+    fn deserialize_broken_integer_errs() {
+        // One byte less fails
+        assert!(matches!(
+            <IntKey<u16>>::from_slice(&[1]).err(),
+            Some(StdError::GenericErr { .. })
+        ));
+
+        // More bytes fails too
```
ethanfrey, line 253: nice

ethanfrey (approved):
looks good.
let's merge

ethanfrey (approved):
Another approval in case the last one was marked as stale

PR #433, opened 2021-09-18
Range with no prefix support

maurolacy:
~~Adds a `NoPrefix` associated type, to use for `range()` short-cuts in `Map` along with the `EmptyPrefix` marker.~~ Just force the empty prefix in `Map::range`.

~~This type was missing from the impl. Adding it basically allows the added test case, where we iterate over a triple key without a prefix.~~ Removed it altogether for simplicity.

Can be reviewed / merged independently of #432. Already integrated in #432 branch.

### DIFF
--- a/packages/storage-plus/src/indexed_map.rs
+++ b/packages/storage-plus/src/indexed_map.rs
@@ -6,7 +6,7 @@ use serde::de::DeserializeOwned;
 use serde::Serialize;
 
 use crate::indexes::Index;
-use crate::keys::{EmptyPrefix, Prefixer, PrimaryKey};
+use crate::keys::{Prefixer, PrimaryKey};
 use crate::map::Map;
 use crate::prefix::{Bound, Prefix};
 use crate::Path;
@@ -138,8 +138,8 @@ where
     }
 
     // use no_prefix to scan -> range
-    fn no_prefix(&self, p: K::NoPrefix) -> Prefix<T> {
-        Prefix::new(self.pk_namespace, &p.prefix())
+    fn no_prefix(&self) -> Prefix<T> {
+        Prefix::new(self.pk_namespace, &[])
     }
 }
 
@@ -149,7 +149,6 @@ where
     K: PrimaryKey<'a>,
     T: Serialize + DeserializeOwned + Clone,
     I: IndexList<T>,
-    K::NoPrefix: EmptyPrefix,
 {
     // I would prefer not to copy code from Prefix, but no other way
     // with lifetimes (create Prefix inside function and return ref = no no)
@@ -163,8 +162,7 @@ where
     where
         T: 'c,
     {
-        self.no_prefix(K::NoPrefix::new())
-            .range(store, min, max, order)
+        self.no_prefix().range(store, min, max, order)
     }
 }
 
--- a/packages/storage-plus/src/indexed_map.rs
+++ b/packages/storage-plus/src/indexed_map.rs
@@ -136,6 +136,11 @@ where
     pub fn sub_prefix(&self, p: K::SubPrefix) -> Prefix<T> {
         Prefix::new(self.pk_namespace, &p.prefix())
     }
+
+    // use no_prefix to scan -> range
+    fn no_prefix(&self, p: K::NoPrefix) -> Prefix<T> {
+        Prefix::new(self.pk_namespace, &p.prefix())
+    }
 }
 
 // short-cut for simple keys, rather than .prefix(()).range(...)
@@ -144,7 +149,7 @@ where
     K: PrimaryKey<'a>,
     T: Serialize + DeserializeOwned + Clone,
     I: IndexList<T>,
-    K::SubPrefix: EmptyPrefix,
+    K::NoPrefix: EmptyPrefix,
 {
     // I would prefer not to copy code from Prefix, but no other way
     // with lifetimes (create Prefix inside function and return ref = no no)
@@ -158,7 +163,7 @@ where
     where
         T: 'c,
     {
-        self.sub_prefix(K::SubPrefix::new())
+        self.no_prefix(K::NoPrefix::new())
             .range(store, min, max, order)
     }
 }
--- a/packages/storage-plus/src/indexes.rs
+++ b/packages/storage-plus/src/indexes.rs
@@ -138,6 +138,15 @@ where
         )
     }
 
+    fn no_prefix(&self, p: K::NoPrefix) -> Prefix<T> {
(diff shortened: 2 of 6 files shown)


packages/storage-plus/src/indexed_map.rs
```diff
@@ -136,6 +136,11 @@ where
     pub fn sub_prefix(&self, p: K::SubPrefix) -> Prefix<T> {
         Prefix::new(self.pk_namespace, &p.prefix())
     }
+
+    // use no_prefix to scan -> range
+    fn no_prefix(&self, p: K::NoPrefix) -> Prefix<T> {
+        Prefix::new(self.pk_namespace, &p.prefix())
+    }
```
ueco-jb, line 143: I noticed this repeated implementation and started to wonder why it's not solved with traits.
I proposed some design here: https://github.com/CosmWasm/cw-plus/pull/436
It's neither finished nor production ready, but I wanted to make a point about functionality.
Please let me know what you think.

packages/storage-plus/src/keys.rs
```diff
@@ -196,6 +209,7 @@ impl<'a> Prefixer<'a> for Addr {
 impl<'a, T: Endian + Clone> PrimaryKey<'a> for IntKey<T> {
     type Prefix = ();
     type SubPrefix = ();
+    type NoPrefix = ();
```
ethanfrey, line 212: This is always `()`. Why do we need such a type?

Couldn't you just use `()` in the code, or remove the argument?

packages/storage-plus/src/map.rs
```diff
@@ -48,6 +48,11 @@ where
         Prefix::new(self.namespace, &p.prefix())
     }
 
+    #[cfg(feature = "iterator")]
+    fn no_prefix(&self, p: K::NoPrefix) -> Prefix<T> {
+        Prefix::new(self.namespace, &p.prefix())
```
ethanfrey, line 53: `p` will always be `()`. `p.prefix()` will always be `vec![]`

Why not return `Prefix::new(self.namespace, &[])` and remove the argument?

packages/storage-plus/src/map.rs
```diff
@@ -401,6 +406,32 @@ mod test {
             .unwrap();
 
         // let's try to iterate!
+        let all: StdResult<Vec<_>> = TRIPLE.range(&store, None, None, Order::Ascending).collect();
```
ethanfrey, line 409: test case makes sense

packages/storage-plus/src/indexes.rs
```diff
@@ -138,6 +138,15 @@ where
         )
     }
 
+    fn no_prefix(&self, p: K::NoPrefix) -> Prefix<T> {
```
ethanfrey, line 141: function makes sense, but let's remove the arg and replace `&p.prefix()` with `&[]`

packages/storage-plus/src/indexes.rs
```diff
@@ -279,6 +288,12 @@ where
         })
     }
 
+    fn no_prefix(&self, p: K::NoPrefix) -> Prefix<T> {
```
ethanfrey, line 291: same as above, no argument needed.

ethanfrey (commented):
I see three useful no_prefix functions.

But since no prefix is always `()` / `vec![]`, let's remove the argument, the extra trait type and just hard code it those 3 places.

packages/storage-plus/src/keys.rs
maurolacy: Sure. I was following the existing impl, but yes, I don't think so much boilerplate is needed. Will simplify.

maurolacy: Simplified `no_prefix`.

I thought `NoPrefix` was needed for traits matching, but it turns out it's superfluous.

This means we can also remove the associated `EmptyTrait`.

packages/storage-plus/src/indexed_map.rs
maurolacy: As discussed below, it's not needed. I just got confused by the existing impl / usage of `EmptyPrefix`.

packages/storage-plus/src/indexed_map.rs
```diff
@@ -158,8 +159,7 @@ where
     where
         T: 'c,
     {
-        self.sub_prefix(K::SubPrefix::new())
```
ethanfrey, line 161: Good call, this is us forcing () onto it. We can just say "ignore all prefixes"

packages/storage-plus/src/iter_helpers.rs
```diff
@@ -1,5 +1,3 @@
-#![cfg(feature = "iterator")]
```
ethanfrey, line 1: I think this was here for "no dead code". But if it compiles, nice to remove it

packages/storage-plus/src/indexed_map.rs
```diff
@@ -1,12 +1,9 @@
-// this module requires iterator to be useful at all
-#![cfg(feature = "iterator")]
```
ethanfrey, line 2: Why remove iterator here?

Do we run tests without iterator feature? We should...

ethanfrey: Secret doesn't allow iterator, and we are working that their 1.0 contracts are compatible with other networks.

ethanfrey (commented):
Looks quite nice. Except for removing all the `#[cfg(feature = "iterator")]` everywhere.

Not sure how we can disable iterator (default) in cosmwasm-storage unless it is explicitly set. We should try to at least build the code without iterator enabled and ensure this works.

ethanfrey:
Please rebase on #438 (maybe after merge?) and fix any CI failures (due to iterator) before merge.

packages/storage-plus/src/iter_helpers.rs
maurolacy: Oh, this was a mistake. Will see the impact.

packages/storage-plus/src/indexed_map.rs
maurolacy: OK, sorry about that. I thought iterator was gone as a feature. Will revert, which will cost me a little rebasing. :-)

maurolacy: Reverted the iterator changes. Let's tackle any failing tests in a different PR.

maurolacy:
> Please rebase on #438 (maybe after merge?) and fix any CI failures (due to iterator) before merge.

Done.

ethanfrey (approved):
lgtm

PR #436, opened 2021-09-19
Use traits to remove repeating implementations

ueco-jb:
Implementation is still not finished, but proposition stands.
I'm fighting with some weird error during test compilation which I don't fully understand:
```
error[E0277]: the trait bound `&[u8; 5]: keys::Prefixer<'_>` is not satisfied
   --> packages/storage-plus/src/map.rs:374:21
    |
374 |             .prefix(b"owner")
    |                     ^^^^^^^^ the trait `keys::Prefixer<'_>` is not implemented for `&[u8; 5]`
    |
    = help: the following implementations were found:
              <&'a [u8] as keys::Prefixer<'a>>
```
(because it IS [implemented](https://github.com/CosmWasm/cw-plus/blob/main/packages/storage-plus/src/keys.rs#L92) - I think I messed something with lifetimes)
and since it's Sunday I don't have much patience to tackle it, so I'll pick it up tommorow.

### DIFF
--- a/packages/storage-plus/src/indexed_map.rs
+++ b/packages/storage-plus/src/indexed_map.rs
@@ -136,6 +136,11 @@ where
     pub fn sub_prefix(&self, p: K::SubPrefix) -> Prefix<T> {
         Prefix::new(self.pk_namespace, &p.prefix())
     }
+
+    // use no_prefix to scan -> range
+    fn no_prefix(&self, p: K::NoPrefix) -> Prefix<T> {
+        Prefix::new(self.pk_namespace, &p.prefix())
+    }
 }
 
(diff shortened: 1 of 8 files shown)


ethanfrey:
This is based on main, right, not on #432?

Can you please rebase it / resolve all conflicts so I have a minimal diff. When it is ready and passes CI, let me know and I can review it.

But first check to see if this overlaps with #439 and maybe combine them somehow.

There are two many PRs with unclear diffs here, making it hard to review or even figure out the order to review in. If this is something that doesn't depends on/conflict with Mauro's work, happy to get it in this week to avoid merge issues

maurolacy:
Yes, it can.

PR #438, opened 2021-09-19
Test storage-plus with iterator disabled

ethanfrey:
Especially for storage-plus.

It seems some code fails already. It would be good to also make iterator optional for any other package that doesn't require it (I think that is most of the packages except multitest, but none of the contracts)

### DIFF
--- a/packages/storage-plus/src/snapshot/mod.rs
+++ b/packages/storage-plus/src/snapshot/mod.rs
@@ -1,8 +1,8 @@
+#![cfg(feature = "iterator")]
 mod item;
 mod map;
 
 pub use item::SnapshotItem;
-#[cfg(feature = "iterator")]
 pub use map::SnapshotMap;
 
 use crate::{Bound, Map, Prefixer, PrimaryKey, U64Key};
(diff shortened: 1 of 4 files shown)


packages/storage-plus/src/snapshot/mod.rs
```diff
@@ -1,8 +1,8 @@
+#![cfg(feature = "iterator")]
 mod item;
 mod map;
 
 pub use item::SnapshotItem;
```
maurolacy, line 5: `SnapshotItem` doesn't require iterator.

maurolacy (commented):
lgtm.

maurolacy (approved):

maurolacy:
Damn, I rebased all of my MRs on top of the version without iterator flags. :-/

Will have to manually introduce these changes again, in each of the branches.

maurolacy:
> Damn, I rebased all of my MRs on top of the version without iterator flags. :-/
> 
> Will have to manually introduce these changes again, in each of the branches.

Perhaps it will be better to delay merging this right now, and do it only after merging the deserialization stuff. Anyway, just an idea. We'll figure it out anyway.

PR #434, opened 2021-09-18
Working `prefix_de`

maurolacy:
#432 follow-up. This separate PR implements / fixes `range_de` for a `prefix_de`.

PR #439, opened 2021-09-20
Prefix consolidation

maurolacy:
Builds on top of #434.

Uses one `Prefix` impl for both `range` and `range_de`. At the expense of a "dummy" deserialization to `Vec<u8>`.

### DIFF
--- a/packages/storage-plus/src/map.rs
+++ b/packages/storage-plus/src/map.rs
@@ -42,15 +42,18 @@ where
         Path::new(self.namespace, &k.key())
     }
 
-    pub fn prefix(&self, p: K::Prefix) -> Prefix<T> {
+    #[cfg(feature = "iterator")]
+    pub fn prefix(&self, p: K::Prefix) -> Prefix<Vec<u8>, T> {
         Prefix::new(self.namespace, &p.prefix())
     }
 
-    pub fn sub_prefix(&self, p: K::SubPrefix) -> Prefix<T> {
+    #[cfg(feature = "iterator")]
+    pub fn sub_prefix(&self, p: K::SubPrefix) -> Prefix<Vec<u8>, T> {
         Prefix::new(self.namespace, &p.prefix())
     }
 
-    pub(crate) fn no_prefix(&self) -> Prefix<T> {
+    #[cfg(feature = "iterator")]
+    pub(crate) fn no_prefix(&self) -> Prefix<Vec<u8>, T> {
         Prefix::new(self.namespace, &[])
     }
 
@@ -115,11 +118,11 @@ where
     T: Serialize + DeserializeOwned,
     K: PrimaryKey<'a>,
 {
-    pub fn sub_prefix_de(&self, p: K::SubPrefix) -> Prefix<T, K::SuperSuffix> {
+    pub fn sub_prefix_de(&self, p: K::SubPrefix) -> Prefix<K::SuperSuffix, T> {
         Prefix::new(self.namespace, &p.prefix())
     }
 
-    pub fn prefix_de(&self, p: K::Prefix) -> Prefix<T, K::Suffix> {
+    pub fn prefix_de(&self, p: K::Prefix) -> Prefix<K::Suffix, T> {
         Prefix::new(self.namespace, &p.prefix())
     }
 }
@@ -213,7 +216,7 @@ where
         self.no_prefix_de().keys_de(store, min, max, order)
     }
 
-    fn no_prefix_de(&self) -> Prefix<T, K> {
+    fn no_prefix_de(&self) -> Prefix<K, T> {
         Prefix::new(self.namespace, &[])
     }
 }
--- a/packages/storage-plus/src/map.rs
+++ b/packages/storage-plus/src/map.rs
@@ -12,7 +12,7 @@ use crate::keys::Prefixer;
 use crate::keys::PrimaryKey;
 use crate::path::Path;
 #[cfg(feature = "iterator")]
-use crate::prefix::{namespaced_prefix_range, Bound, Pair2, Prefix, Prefix2, PrefixBound};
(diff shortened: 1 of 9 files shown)


packages/storage-plus/src/de.rs
```diff
@@ -0,0 +1,157 @@
+use crate::keys::IntKey;
+use cosmwasm_std::{Addr, StdError, StdResult};
+use std::array::TryFromSliceError;
+use std::convert::TryInto;
+
+pub trait Deserializable {
+    type Output: Sized;
+
+    fn from_slice(value: &[u8]) -> StdResult<Self::Output>;
+    fn from_vec(value: &Vec<u8>) -> StdResult<Self::Output>;
```
ethanfrey, line 10: `fn from_vec(value: &Vec<u8>)` is a very odd signature.
if you made it `fn from_vec(value: Vec<u8>)` I would understand this better (to avoid clone when you can grant ownership).

If this is not the purpose, then I think we need better documentation for the difference

packages/storage-plus/src/de.rs
```diff
@@ -0,0 +1,157 @@
...
+            type Output = $t;
+
+            fn from_slice(value: &[u8]) -> StdResult<Self::Output> {
+                Ok(<$t>::from_be_bytes(value.try_into()
+                    // FIXME: Add and use StdError try-from error From helper
+                    .map_err(|err: TryFromSliceError| StdError::generic_err(err.to_string()))?))
+            }
+
+            fn from_vec(value: &Vec<u8>) -> StdResult<Self::Output> {
+                Ok(<$t>::from_be_bytes(value.to_vec().try_into().unwrap()))
```
ethanfrey, line 77: And why does this unwrap rather than map_err like above?

packages/storage-plus/src/de.rs
```diff
@@ -0,0 +1,157 @@
...
+                // FIXME: Add and use StdError try-from error From helper
+                .map_err(|err: TryFromSliceError| StdError::generic_err(err.to_string()))?,
+        ) as usize;
+        let u = U::from_slice(&value[4 + t_len..4 + t_len + u_len])?;
+        let v = V::from_slice(&value[4 + t_len + u_len..])?;
+
+        Ok((t, u, v))
+    }
+
+    fn from_vec(value: &Vec<u8>) -> StdResult<Self::Output> {
```
ethanfrey, line 138: If these are the same implementation, and we know we can Deref Vec to &[u8], can't you just implement as

```rust
fn from_vec(value: &Vec<u8>) -> StdResult<Self::Output> {
  Self::from_slice(&value)
}
```

packages/storage-plus/src/map.rs
```diff
@@ -108,13 +110,29 @@ where
...
 }
 
+#[cfg(feature = "iterator")]
+impl<'a, K, T> Map<'a, K, T>
+where
+    T: Serialize + DeserializeOwned,
+    K: PrimaryKey<'a>,
+{
+    #[cfg(feature = "iterator")]
+    pub fn sub_prefix_de(&self, p: K::SubPrefix) -> Prefix<T, K::SuperSuffix> {
```
ethanfrey, line 120: This prefix and suffix (sub prefix and subsuffix) stuff looks proper

ethanfrey (commented):
A first pass on reviewing.

It seems like you are adding proper functionality. The implementation doesn't look terribly clean.

Maybe you can just improve it after vacation. The ideas are here and just need some time to get polished.

packages/storage-plus/src/de.rs
maurolacy: This is wip at the moment. My idea here was to pass a Vec to try to do a zero-copy / no-op impl in the case the Output is a Vec.

Clippy says this is more expensive than passing a reference to a slice, so, seems `from_slice` is the way to go. Will revert this.

The only way I can think of to try to improve on this, would be to use a macro instead of a function to do the conversion, and (somehow) do nothing in the case of Vec output.

packages/storage-plus/src/de.rs
maurolacy: Good call, I forgot about it. This is just because of wip.

packages/storage-plus/src/de.rs
maurolacy: Good. I think that's what I was trying to achieve. But then this is probably no better than the original `from_slice` impl.

packages/storage-plus/src/map.rs
maurolacy: This is coolest thing I did in all this I guess.

packages/storage-plus/src/de.rs
```diff
@@ -0,0 +1,157 @@
...
+        let t_len = u16::from_be_bytes(
+            value[..2]
+                .try_into()
+                // FIXME: Add and use StdError try-from error From helper
+                .map_err(|err: TryFromSliceError| StdError::generic_err(err.to_string()))?,
+        ) as usize;
+        let t = T::from_slice(&value[2..2 + t_len])?;
```
ueco-jb, line 108: Is range here required?
`&value[2..2 `

ueco-jb (commented):
There are plenty of seemingly magic numbers in `Deserializable` methods which I would rather see exaplained either in some comment(s) or in named variables.

packages/storage-plus/src/de.rs
maurolacy: This is just length prefixed deserialization. See `to_length_prefixed` and friends to get a better grasp at it. A way to improve on this would be to add a `from_length_prefixed` to encapsulate this, passing the number of slices.

Can be done in a follow-up.

maurolacy:
Will merge this into #432 after CI checks, if no objections are raised.

ueco-jb (approved):
No objections, I'd rather review full changeset in #432

PR #442, opened 2021-09-20
Small storage-plus docs update

### LINKED ISSUE #435: small updates on storage-plus docs
- The [example for Map](https://github.com/CosmWasm/cw-plus/tree/main/packages/storage-plus#map) says `Map::new(b"people")` but it should just be `Map::new("people")`

- Says "This requires the iterator feature to be enabled in cw-storage-plus"... while technically true, the iterator is enabled by default, so it seems a little unnecessary to mention it (I went and enabled the feature before realizing)

hashedone:
Closes #435

### DIFF
--- a/packages/storage-plus/README.md
+++ b/packages/storage-plus/README.md
@@ -51,7 +51,7 @@ struct Config {
 }
 
 // note const constructor rather than 2 functions with Singleton
-const CONFIG: Item<Config> = Item::new(b"config");
+const CONFIG: Item<Config> = Item::new("config");
 
 fn demo() -> StdResult<()> {
     let mut store = MockStorage::new();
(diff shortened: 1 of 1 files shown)


packages/storage-plus/README.md
```diff
@@ -113,7 +113,8 @@ iteration. That's right, you can list all items in a `Map`, or only
 part of them. We can efficiently allow pagination over these items as
 well, starting at the point the last query ended, with low gas costs.
 This requires the `iterator` feature to be enabled in `cw-storage-plus`
-(which automatically enables it in `cosmwasm-std` as well).
+(which automatically enables it in `cosmwasm-std` as well, and which is
+enabled by default).
```
ueco-jb, line 117: IMO this sentence sounds bit confusing.

> We can efficiently allow pagination (...). This requires the `iterator` feature to be enabled in `cw-storage-plus` (which automatically enables it in `cosmwasm-std` as well, and which is enabled by default).

Maybe change to something like this?

> This requires the `iterator` feature to be enabled in `cw-storage-plus` (it is enabled by default; it also automatically enables it in `cosmwasm-std`).

ueco-jb (approved):
I would reorder one sentence, beside that 

LGMT

packages/storage-plus/README.md
```diff
@@ -327,9 +328,9 @@ over all items with `range(store, min, max, order)`. It supports `Order::Ascendi
 
 ```rust
 #[derive(Copy, Clone, Debug)]
-pub enum Bound<'a> {
-    Inclusive(&'a [u8]),
-    Exclusive(&'a [u8]),
+pub enum Bound {
+    Inclusive(Vec<u8>),
+    Exclusive(Vec<u8>),
```
ethanfrey, line 333: We don't have None anymore, right? Just `Option<Bound>`

ethanfrey (approved):
Thanks for this fix

Issue #407, opened 2021-09-09
storage-plus: Improve in-code documentation of map primitives, in particular `MultiIndex`


hashedone:
I found some storage-plus primitives not documented or poorly documented, and I think it should be improved. I know some of them are used around and those usages can be taken as an example (also including tests which are in pretty good shape there), but it takes way more time than reading doc comments (which IDEs are just bringing on hover), not mentioning generating doc with cargo.

Just as an example `MultiIndex::new` - there is no single word what is the second argument of the `idx_fn` (also no single word about first one, but it is kind of easier to deduce). Also there is really not good explanation how to use it, and why to include the second `idx_fn` argument as part of key.

maurolacy:
There's an explanation that addresses some of those points in the docs:

https://docs.cosmwasm.com/tutorials/storage/indexes/

Feel free to suggest updates / clarifications,  and / or to extend on it.

maurolacy:
Perhaps an issue must be opened in the [docs](https://github.com/CosmWasm/docs), so that those changes / suggestions can be easily tracked.

PR #443, opened 2021-09-20
Improve in-code documentation of map primitives

ueco-jb:
closes https://github.com/CosmWasm/cw-plus/issues/407

It's not a world-breaking change, but can improve work a bit - when IDE shows suggestion about structures, it often shows Rust's doc comments as well.

### DIFF
--- a/packages/storage-plus/src/snapshot/mod.rs
+++ b/packages/storage-plus/src/snapshot/mod.rs
@@ -11,7 +11,8 @@ use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 
 /// Structure holding a map of checkpoints composited from
-/// height as U64Key and checkpoint id.
+/// height (as U64Key) and counter of how many times it has
+/// been checkpointed (as u32).
 /// Stores all changes in changelog.
 #[derive(Debug, Clone)]
 pub(crate) struct Snapshot<'a, K, T> {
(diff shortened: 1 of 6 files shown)


packages/storage-plus/src/snapshot/mod.rs
```diff
@@ -10,6 +10,9 @@ use cosmwasm_std::{Order, StdError, StdResult, Storage};
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 
+/// Structure holding a map of checkpoints composited from
+/// height as U64Key and checkpoint id.
```
ueco-jb, line 14: Is this correct? I have hard time understanding purpose of this `u32`.

ethanfrey: u32 is the data stored there.

It is a counter of how many times it has been checkpointed (maybe 2 users checkpointed, 1 released, but this stays checkpointed).

Good to add that as a comment

packages/storage-plus/src/snapshot/mod.rs
```diff
@@ -23,6 +26,11 @@ pub(crate) struct Snapshot<'a, K, T> {
 }
 
 impl<'a, K, T> Snapshot<'a, K, T> {
+    /// Example:
```
ethanfrey, line 29: This is only used crate internal, so these docs are for other internal devs

ethanfrey (commented):
Nice start. Some feedback

packages/storage-plus/src/snapshot/mod.rs
ueco-jb: I've added that info as comment.

packages/storage-plus/src/snapshot/mod.rs
ueco-jb: I removed this whole comment.

ueco-jb:
I had plenty of issues while compiling those examples I've added, but finally I've managed to keep it small and precise.

ethanfrey (approved):
Looks good and brings knowledge from test code closer to the docs.

Issue #391, opened 2021-08-25
Proposal for improving contract interfaces


hashedone:
For now contract interfaces are defined by structures defining messages (queries/executes/migrations/instantiations) for contracts. On the one hand this seems simple, but writing contracts I found problems with this approach. First of all re usability and extend ability of contracts. If there is cw20-base contract defining some basic implementation of cw20 token, and then there is another contract which is based on it, but just extending it further, then the new contract is typically copying messages from base contract, adding on top of them, and in the implementation basically forward-calling implementation of cw20-base. I find this messy, and difficult to maintain - in case of update of cw20-base, all contracts basing on it should be probably updated, but in reality we technically cannot do that - because someone may just have another contract based on it, and he would not be notified anyhow about changes (unless he reads all changelogs).

My proposal is, to add possibility of more natural way of defining interfaces - as traits. For cw20 contract trait could look like:
```rust
#[cw_derive::interface]
trait Cw20Base {
  #[exec]
  fn transfer(&self, ctx: CtxMut, recipient: Addr, amount: u128) -> Result<Response, Error>;
  
  #[exec]
  fn burn(&self, ctx: CtxMut, amount: String) -> Result<Response, Error>;
  
  // Another exec messages
  // ...
  
  #[query]
  fn balance(&self, ctx: Ctx, address: Addr) -> Result<BalanceResponse, Error>;
  
  #[query]
  fn token_info(&self, ctx: Ctx) -> Result<TokenInfoResponse, Error>;
  
  // Another queries
  // ...
}
```

The whole idea is to provide `interface` attribute which generates structures for parsing from jsons, and generates dispatching function which takes parsed input and calls proper functions. Also this handles intermediate wrappers like `Uint128` which as far as I understand are just for proper parsing, but using them in Rust code itself seems inconvenient. Also taking `Addr` in interfaces is proper here - dispatcher can know to convert it to validate it with `Addr::validate(...)`.  The additional `Ctx` type is helper providing access to things such as `Deps`/`DepsMut`, `Env`, `MessageInfo`.

The second part of this would be actually generating implementation structures, like:

```rust
#[cw_derive::contract(Cw20Base)]
struct Cw20BaseContract;

#[cw_derive::contract_impl]
impl Cw20BaseContract {
  #[instantiate]
  fn instantiate(&mut self, ctx: CtxMut, name: String, symbol: String, ...)
    -> Result<Response, Error> { todo!() }
  
  #[migrate]
[...]

PR #445, opened 2021-09-21
Make NFT Customisable

### LINKED ISSUE #440: Allow NFTs to include custom data
A first step on #391

A few people have tried to build on cw721-base but want to add extra metadata to the contract and cannot extend it. Basically, we would need:

```rust
pub struct TokenInfo<T> {
    /// The owner of the newly minted NFT
    pub owner: Addr,
    /// Approvals are stored here, as we clear them all upon transfer and cannot accumulate much
    pub approvals: Vec<Approval>,

    /// Identifies the asset to which this NFT represents
    pub name: String,
    /// Describes the asset to which this NFT represents
    pub description: String,
    /// A URI pointing to an image representing the asset
    pub image: Option<String>,

    pub extension: Option<T>,
}
```
([source](https://github.com/CosmWasm/cw-plus/blob/main/contracts/cw721-base/src/state.rs#L9-L21))

The default implementation can simply leave `extension: None`. (Or use `extension: Empty`, removing the Option).

This can be done and then we make all query and execute methods accept this T. However, when thinking about it, it would be a great first example of using a struct to represent a contract, as described in #391

ethanfrey:
Closes #440 

- [x] turn cw721 contract into a struct
- [x] make it generic over all Response types
- [x] tests pass
- [x] add generic extension to the state
- [x] expose generic extensions in query

This is also a step to make a more reusable contract structure, looking more like controllers and the ideas in #391 
We can iterate on this (what belongs in lib.rs, how to customise the generics, etc), but I think this is a big step to re-usability. Maybe best to get feedback from people trying to extend this.


contracts/cw721-base/src/contract.rs
```diff
@@ -14,442 +12,609 @@ use cw721::{
...
+pub struct Cw721Contract<'a, T, C>
+where
+    T: Serialize + DeserializeOwned + Clone,
+{
+    pub contract_info: Item<'a, ContractInfoResponse>,
+    pub minter: Item<'a, Addr>,
+    pub token_count: Item<'a, u64>,
+    /// Stored as (granter, operator) giving operator full control over granter's account
+    pub operators: Map<'a, (&'a Addr, &'a Addr), Expiration>,
+    pub tokens: IndexedMap<'a, &'a str, TokenInfo<T>, TokenIndexes<'a, T>>,
```
hashedone, line 38: I like the idea of having state as part of implementation structure as a part of contract type itself - I didn't put it in the original proposal as I wanted to make it minimal, but if it is achievable then I think it is a good idea.

contracts/cw721-base/src/contract.rs
```diff
@@ -14,442 +12,609 @@ use cw721::{
...
+
+    _custom_response: PhantomData<C>,
```
hashedone, line 40: There is no reason to make this contract generic over `C`. If contract interface would be a trait, it should be an associated type (as for any type `T` there is only one well defined type `C`).

contracts/cw721-base/src/contract.rs
```diff
@@ -14,442 +12,609 @@ use cw721::{
...
-
-    // Send message
-    Ok(Response::new()
-        .add_message(send.into_cosmos_msg(contract.clone())?)
-        .add_attribute("action", "send_nft")
-        .add_attribute("sender", info.sender)
-        .add_attribute("recipient", contract)
-        .add_attribute("token_id", token_id))
-}
+    pub fn execute(
```
hashedone, line 119: I would definitely move this out of this struct implementation, probably to the structure message, and make it dispatching over any contract which implements a trait representing the interface of this contract. In this particular case, the trait would be a `Cw721`, and the `execute` would be:

```rust
impl<T> ExecMsg<T> {
  /// Dispatches this message to contract
  fn execute<Contract: Cw721>(self, contract: &Contract) -> Result<Response<Contract::C>, Contract::Err> {
    match self {
      ExecuteMsg::Mint(msg) => self.mint(deps, env, info, msg),
      // All other branches
    }
  }
}
```

`Err` and `C` (Which should get some meaningfull name - there is no restriction on generic arguments identifier lengths ;) ) are dependent on actual contract implementation (as it may and possibly would differ for different contracts).

contracts/cw721-base/src/contract.rs
```diff
@@ -14,442 +12,609 @@ use cw721::{
...
-    let mut token = tokens().load(deps.storage, &token_id)?;
-    // ensure we have permissions
-    check_can_send(deps.as_ref(), env, info, &token)?;
-    // set owner and remove existing approvals
-    token.owner = deps.api.addr_validate(recipient)?;
-    token.approvals = vec![];
-    tokens().save(deps.storage, &token_id, &token)?;
-    Ok(token)
-}
+    pub fn execute_mint(
```
hashedone, line 154: Now there is the clue of #391 - the idea is to move all of those functions to the trait like:

```rust
trait Cw721<T> where T: Serialize + DeserializeOwned + Clone {
  type Err: ToString;
  type C: Serialize + DeserializeOwned + Clone;

  fn mint(&self, deps: DepsMut, _env: Env, MessageInfo, msg: MingMsg<T>) -> Result<Response<Self::C>, Self::Err>;
}
```

Then the trait is implemented by contract itself, but may also be implemented by any other contract which also react to this contract msg.

Also I find this naming very verbose - basically `execute` sounds like `do action` so it feels like `mint` is just expressive enough. I would personally avoid prefixing functions if the name is not ambiguous, and leave them matching the message name - but this is side note.

contracts/cw721-base/src/contract.rs
```diff
@@ -14,442 +12,609 @@ use cw721::{
...
-        )?),
-        QueryMsg::NumTokens {} => to_binary(&query_num_tokens(deps)?),
-        QueryMsg::Tokens {
-            owner,
-            start_after,
-            limit,
-        } => to_binary(&query_tokens(deps, owner, start_after, limit)?),
-        QueryMsg::AllTokens { start_after, limit } => {
-            to_binary(&query_all_tokens(deps, start_after, limit)?)
+    pub fn query(&self, deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
```
hashedone, line 438: Same as for `execute`

hashedone (commented):
Basically like it, but I find very important point of #391 - extract interfacing of contract to a trait instead of purely message (and leaving message only to be the [de]serialization and dispatching point). Not approving nor requesting changes at this point, as it is more like to be discussed stuff.

contracts/cw721-base/src/contract.rs
```diff
@@ -14,442 +12,609 @@ use cw721::{
...
-        .add_attribute("sender", info.sender)
-        .add_attribute("recipient", recipient)
-        .add_attribute("token_id", token_id))
-}
+impl<'a, T, C> Cw721Contract<'a, T, C>
+where
+    T: Serialize + DeserializeOwned + Clone,
+    C: Clone + std::fmt::Debug + PartialEq + JsonSchema,
+{
+    pub fn instantiate(
```
hashedone, line 100: In contradiction to `execute`/`query` this should stay in contract itself (as it is very much per contract stuff).

contracts/cw721-base/src/contract.rs
```diff
@@ -14,442 +12,609 @@ use cw721::{
...
-        symbol: msg.symbol,
-    };
-    CONTRACT_INFO.save(deps.storage, &info)?;
-    let minter = deps.api.addr_validate(&msg.minter)?;
-    MINTER.save(deps.storage, &minter)?;
-    Ok(Response::default())
+const DEFAULT_LIMIT: u32 = 10;
+const MAX_LIMIT: u32 = 30;
+
+pub struct Cw721Contract<'a, T, C>
```
hashedone, line 29: I would consider moving this to `state.rs` as it would be more consistent with way we stored it before, and leave only the methods implementation here - but I am not sure.

contracts/cw721-base/src/lib.rs
```diff
@@ -4,3 +4,43 @@ pub mod msg;
...
+
+    #[cfg_attr(not(feature = "library"), entry_point)]
+    pub fn execute(
+        deps: DepsMut,
+        env: Env,
+        info: MessageInfo,
+        msg: msg::ExecuteMsg<Extension>,
+    ) -> Result<Response, ContractError> {
+        let tract = contract::Cw721Contract::<Extension, Empty>::default();
+        tract.execute(deps, env, info, msg)
```
hashedone, line 38: Relating to other comment about moving this to message itself it would obviously go into form of `msg.execute(contract, deps, env, info)`, and similar for `query`.

contracts/cw721-base/src/contract.rs
ethanfrey: It's not just achievable, it is essential in this use-case to allow the state itself to be generic over T and ensure all state access uses the same T

contracts/cw721-base/src/contract.rs
ethanfrey: Good point. Each method (`execute_*`) should then just take `C` as a generic parameter for this method, right?

I didn't think of it, but also part of it comes from laziness.

`C: Clone + std::fmt::Debug + PartialEq + JsonSchema` is a mouthful to put everywhere.

I would like to do something like `type CustomMsg = Clone + std::fmt::Debug + PartialEq + JsonSchema` to make this simpler, but last time I tried something similar, I got compiler errors. I could make a trait:

```rust
pub trait CustomMsg: Clone + std::fmt::Debug + PartialEq + JsonSchema {}
```

But that would require me to also implement the trait specifically for all messages (which are very few, so many that is actually a good idea)

contracts/cw721-base/src/contract.rs
ethanfrey: This makes sense.

The problem is this is also not reusable if they add even one field to the ExecMsg in some extension. They would need something like `impl TryFrom<CustomExecMsg> for ExecMsg`, which would be quite elegant if we could use [`#[serde(flatten)]`](https://serde.rs/attr-flatten.html) but we [cannot do that yet](https://github.com/CosmWasm/serde-json-wasm/issues/20).

I could do this inversion of control (no problem), but it is only really useful if all derived contracts have the exact same message structure or we can use flatten to extend them.

contracts/cw721-base/src/contract.rs
ethanfrey: Yes, I was doing the minimal subset here. I will pull out traits, sounds good.

contracts/cw721-base/src/contract.rs
ethanfrey: I was thinking of that, but afaik, I need to implement all methods on a struct in the same module as the struct. Can I implement them in other modules in the same package?

ethanfrey:
Thanks for the feedback. Some stuff I will implement right away, other we can discuss more.
Let me update this for the places we agree then we discuss the remainders

contracts/cw721-base/src/contract.rs
ethanfrey: @hashedone I addressed this in b32b21e6

I agree it is more proper. I also made the CustomMsg trait as a placeholder (but we need to move it somewhere else to really use it).

One thing I noticed is it made tests a bit more verbose (adding `::<Empty>` in lots of places), but this is the cost of generics I guess.

contracts/cw721-base/src/contract.rs
ethanfrey: > type C: Serialize + DeserializeOwned + Clone

Note this contradicts what I just did in https://github.com/CosmWasm/cw-plus/pull/445/commits/b32b21e6919cb15d49875ec5b6df0b5930562a8a which makes C generic on the method (you fix it for the contract when implementing the trait).

If we remove Self::C and keep Self::Error I can see this coming along

ethanfrey: I started a bit of this in 19803f21 to get first feedback.

Note that Mint is not in the generic cw721 standard, so we have the case already of the contract extending the default. This is more when you try to move execute into the message type.

contracts/cw721-base/src/contract.rs
hashedone: It still can be done bypassing serde (implementing `Serialize` just by forwarding calls to the enums), but it looks like much to be done (except if it may be generated in which case it is free).

contracts/cw721-base/src/contract.rs
hashedone: No, you need to do it on the same crate (and you need to have visibility for fields so they have to be `pub` or at least `pub(in: ...)`.

ethanfrey: Interesting.
I will try to pull this out later once we sort out the trait issues

contracts/cw721-base/src/lib.rs
```diff
@@ -4,3 +4,43 @@ pub mod msg;
...
+        tract.instantiate(deps, env, info, msg)
+    }
+
+    #[cfg_attr(not(feature = "library"), entry_point)]
+    pub fn execute(
+        deps: DepsMut,
+        env: Env,
+        info: MessageInfo,
+        msg: msg::ExecuteMsg<Extension>,
+    ) -> Result<Response, ContractError> {
```
hashedone, line 36: I was thinking about the problem of handling multiple types of messages (like base message and extension message), and I got a revelation - this is a place. As it is just a boilerplate it makes sense to instead of take `msg` as `msg::ExecuteMsg<Extension>`, take it here as something generic, before it is parsed (like `Binary`), and then just make deserialization here, like (this is more pseudocode):
```rust
let contract = contract::Cw721Contract::default();
if let Ok(msg) = deserialize::<msg::ExecuteMsg<Extension>>() {
  Cw721::execute(&contract, deps, env, info, msg)
} else if let Ok(msg) = deserialize::<extension::ExecuteMsg>() {
  Extension::execute(&contract, deps, env, info, msg)
} else {
  return Err(StdError::InvalidMessage.into());
}
```

The only problem I have here is if there is any type which is cheap/free to deserialize into by `entry_point` (so deserialization is not performed twice). In future, when this is generated it can even generate raw entry point.

ethanfrey: I would auto-generate some ser/deserialization code that does that.

eg.

```rust
enum ExtendedExecMsg {
  Cw721(cw721::ExecMsg),
  My(crate::ExecMsg),
}
```

And auto-generate ser/deser such that it flattens them, basically doing what you do above - try to deter into cw721::ExecMsg, if that fails, deser into My.

This is a subset of the serde "flatten" option, and maybe you want to make a macro to implement Serialize on this? Or maybe we try it first with a manually implemented Serialize/Deserialize and see if it makes sense.

contracts/cw721-base/src/state.rs
```diff
@@ -18,6 +19,9 @@ pub struct TokenInfo {
     pub description: String,
     /// A URI pointing to an image representing the asset
     pub image: Option<String>,
+
+    /// You can add any custom metadata here when you extend cw721-base
+    pub extension: T,
```
JakeHartnell, line 24: This is great! Not sure if description / image are needed anymore if there is an extension?

Much more usable like this though. :pray:

contracts/cw721-base/src/contract.rs
ethanfrey: The issue is that I want my `Cw721BaseContract` struct to implement` Cw721<C, T>` for any `C` and `T`.
There is no correlation between the two. C is usually Empty, but if you need to integrate with Custom blockchain messages it is possible. `T` may be many values based on who is extending it.

If I do something like:

```rust
trait Cw721<T> where T: Serialize + DeserializeOwned + Clone {
  type Err: ToString;
  type C: Serialize + DeserializeOwned + Clone;
}
```

like you do below, it hardcodes the Response<C> and only allows 1 C for a given contract. This does now allow us to extend it for CustomMsg.

I agree fully with `type Err`, as that is fixed in the implementation (but not by the trait). But I would move C up to a trait generic rather than a trait attribute (not sure the proper terms here... please correct me, but I think you got the idea)

contracts/cw721-base/src/contract.rs
ethanfrey: > I actually don't see any sense in making C generic in methods - its semantic is to be always the same for given T. it makes no sense for having it generic to be decided on call moment, and more important - there is no real way to elide its type there!

T is the extra data added to TokenInfo when storing it.
C is custom message type from Response<C>.
These have nothing to do with each other.

But I do agree to putting this back into a trait/struct bound not a method bound

ethanfrey:
@hashedone I think I have done all I will do in this PR.

Outstanding questions:
- Do you agree with my use of C as a trait generic?
- What do you think of the two smaller interfaces (Cw721Query, Cw721Execute) which can be implemented separately?
  - And how to deal with associated types like Err?
- Should we implement and export a Cw721Base trait for the extra calls we have (mint and minter) that extends Cw721?
- Separating query.rs and execute.rs? Good idea or bad? (The 1300+ line contract.rs was already giving me headaches scrolling up and down rather than switching files, so I prefer to separate the tests out at least)

The issue with dispatching in execute and query is a good one. I think putting that in the Msg types makes sense as soon as we can easily merge multiple enums (even manually implementing Serialize/Deserialize). Happy for a follow-up PR if you want to do that, then pull most of the dispatching into cw721::ExecuteMsg / cw721::QueryMsg. That is too big to pile into an already enormous PR.

contracts/cw721-base/src/contract.rs
hashedone: I am looking at this and I don't get the usecase, but it would probably be more clear when I would see it. I just cannot imagine a case when you want more than one implementation of a trait per contract. Now I am even thinking that even T is probably the associated type, as at the end for your entry points you have to make proper type for deserialization, and at the end your contract is basically setting it in stone. On the other hand it may be, that the contract itself is operating on very specific `T`/`C`, but maybe somewhere in future when you extend it you may pass-forward calls and then it is useable. But I don't see generic usage in C in the implementation so I am not sure.

Anyway - let leave it for now, it is huge step forward anyway, and I think if we will observe how it evolves, we would have better POV on this.

contracts/cw721-base/src/contract.rs
hashedone: Yup, I figured it out bit later. It should definitely be on trait bound, probably not necessary for a struct itself (you can implement generic struct with generics which your struct is not generic over), but it is a detail (I elaborated in different comment).

hashedone (approved):
Nice refactor, I like it at this point.

PR #446, opened 2021-09-21
Add proper prefix_range helper when you want to iterate over the prefix space

### LINKED ISSUE #430: Unintuitive behavior of range on multi-index
Failing test case can on #429.

Technically it is not a bug, but the behavior seems to make `Bound::inclusive` not very useful for upper bound of multi index. This whole issue is basically consequence of relying on lexicographical order for comparing keys, and raw `&[u8]` keys instead of proper key semantics. Somehow related to #423 and #198.

ethanfrey:
Closes #430 

Fixes issue demonstrated in #429 (takes that commit as a starter)

### DIFF
--- a/packages/storage-plus/src/indexed_map.rs
+++ b/packages/storage-plus/src/indexed_map.rs
@@ -747,7 +747,7 @@ mod test {
 
         #[test]
         #[cfg(feature = "iterator")]
-        fn composite_key_failure() {
+        fn composite_key_query() {
             let indexes = Indexes {
                 secondary: MultiIndex::new(
                     |secondary, k| (U64Key::new(*secondary), k),
@@ -759,6 +759,7 @@ mod test {
             let mut store = MockStorage::new();
 
             map.save(&mut store, "one", &1).unwrap();
+            map.save(&mut store, "two", &2).unwrap();
 
             let items: Vec<_> = map
                 .idx
--- a/packages/storage-plus/src/indexed_map.rs
+++ b/packages/storage-plus/src/indexed_map.rs
@@ -700,4 +700,53 @@ mod test {
         assert_eq!(datas[0], marias[0].1);
         assert_eq!(datas[1], marias[1].1);
     }
+
+    mod inclusive_bound {
+        use super::*;
+        use crate::U64Key;
+
+        struct Indexes<'a> {
+            secondary: MultiIndex<'a, (U64Key, Vec<u8>), u64>,
+        }
+
+        impl<'a> IndexList<u64> for Indexes<'a> {
+            fn get_indexes(&'_ self) -> Box<dyn Iterator<Item = &'_ dyn Index<u64>> + '_> {
+                let v: Vec<&dyn Index<u64>> = vec![&self.secondary];
+                Box::new(v.into_iter())
+            }
+        }
+
+        #[test]
+        #[cfg(feature = "iterator")]
+        fn composite_key_failure() {
+            let indexes = Indexes {
+                secondary: MultiIndex::new(
+                    |secondary, k| (U64Key::new(*secondary), k),
+                    "test_map",
+                    "test_map__secondary",
+                ),
+            };
+            let map = IndexedMap::<&str, u64, Indexes>::new("test_map", indexes);
+            let mut store = MockStorage::new();
+
+            map.save(&mut store, "one", &1).unwrap();
+
+            let items: Vec<_> = map
+                .idx
+                .secondary
+                .range(
+                    &store,
+                    None,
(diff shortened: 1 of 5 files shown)


packages/storage-plus/src/indexed_map.rs
```diff
@@ -700,4 +729,53 @@ mod test {
...
+        impl<'a> IndexList<u64> for Indexes<'a> {
+            fn get_indexes(&'_ self) -> Box<dyn Iterator<Item = &'_ dyn Index<u64>> + '_> {
+                let v: Vec<&dyn Index<u64>> = vec![&self.secondary];
+                Box::new(v.into_iter())
+            }
+        }
+
+        #[test]
+        #[cfg(feature = "iterator")]
+        fn composite_key_failure() {
```
hashedone, line 750: I guess this test should be renamed (`composite_key_range`?) as it is never more showing the failure.

ethanfrey: yes, good point.

is that the only change request needed for approval?

hashedone: Sorry, I was commenting while still reading code :D But I guess yes.

hashedone (approved):
Thanks for a change. Very much like how it is working now.

packages/storage-plus/src/indexed_map.rs
```diff
@@ -700,4 +729,54 @@ mod test {
...
+
+            let items: Vec<_> = map
+                .idx
+                .secondary
+                .prefix_range(
+                    &store,
+                    None,
+                    Some(PrefixBound::inclusive(1)),
+                    Order::Ascending,
+                )
```
maurolacy, line 772: Sorry for the late comment, but how is this different from
```
                .prefix(U64Key::new(1))
                .range(
                    &store,
                    None,
                    None,
                    Order::Ascending,
                )
```
?

maurolacy: Also, I fail to see the use case where the current `range` / `prefix` limitations with specifying bounds would be an issue.

hashedone: It would include every items with prefixes up to `1`, not only `1` - in case of unsigned it would return all items with prefix `0` and `1` basically, without guessing the whole key upper bound.

ethanfrey: This avoids the need of the joined_key and manually incrementing.
If you look at the different implementations of bounds, one "extends with a 0 byte" the other "increments last byte by 1". This makes it different than the basic range.

That and I added a bit of type-safety to the queries.

If you look at the other query tests in Map, they show how to easily search for prefix=2 to prefix=5 inclusive or exclusive.

In this case 2 exclusive starts with `3, vec![]` and doesn't include `2, vec![0xff]`

maurolacy: I see, this would be similar to a full range, but without the need of guessing / knowledge on internals; and avoiding the need of `joined_key`.

I still think it's confusing having three different range impls (`range`, `prefix` plus `range`, and now `prefix_range`). And, with subtle differences between them.

maurolacy: This would then be similar to:
```rust
                .range(
                    &store,
                    Some(Bound::inclusive((U64Key::new(0), "").joined_key())),
                    Some(Bound::exclusive((U64Key::new(2), "").joined_key())),
                    Order::Ascending,
                )
```

Which, I agree, is a little bit of a mouthful. And difficult to get right / reason about.

Nevertheless, if we eliminate the need of specifying the pk, and of using `joined_key`, this `range` impl would be acceptable / good enough in my opinion.

ethanfrey: I think we need to overhaul range and the prefix() + range() one.
We do have two ways:
 - iterate over a range specified by entire keys
 - iterate over a range specified by some prefix/subprefix of a compound key (very common with secondary indexes)

The range implementation should be improved as well (aside for the key deserialisation which is interesting work and makes all these methods better)

maurolacy: Just for completeness, the code above is also equivalent to:

```rust
                .range(
                    &store,
                    Some(Bound::inclusive(
                        map.idx.secondary.index_key((U64Key::new(0), vec![])),
                    )),
                    Some(Bound::exclusive(
                        map.idx.secondary.index_key((U64Key::new(2), vec![])),
                    )),
                    Order::Ascending,
                )
```

where the `index_key` helper method is used, instead of the "free" `joined_key`. This doesn't look very user friendly either, I must say.

See #462 for a possible way forward from here.

PR #448, opened 2021-09-22
Release 0.10.0-soon

ethanfrey:
Prepare a release, so we can start integrating the latest and greatest into the rest of our stack.

Still some work with the range_de stuff, but mostly there.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -241,7 +241,7 @@ dependencies = [
 
 [[package]]
 name = "cw-controllers"
-version = "0.9.0"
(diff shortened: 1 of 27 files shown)


ueco-jb (approved):
lgmt

ueco-jb:
Regenerate changelog now or in separate review?

ethanfrey:
> Regenerate changelog now or in separate review?

Will do that when we hit 0.10.0... no need for a pre-release.
But thanks for reminding me, I would have forgotten

PR #450, opened 2021-09-23
Demo metadata extension

ethanfrey:
Follow up of #440 and #445

It was requested to have something out of the box that works like Ethereum with a "metadata uri".

I added a new contract that just extends `cw721-base` to demonstrate how to easily add that (or any other metadata). It also let me do a bit of cleanup on the interfaces and how we export objects.

Look at the [README](https://github.com/CosmWasm/cw-plus/blob/440-demo-metadata-extension/contracts/cw721-metadata-uri/README.md) for a better description

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,6 +17,7 @@ workflows:
       - contract_cw20_staking
       - contract_cw20_merkle_airdrop
       - contract_cw721_base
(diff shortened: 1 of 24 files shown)


hashedone (approved):
Very trivial extending case, good starting point showing that this approach makes extension easier without going into some complicated cases.

PR #453, opened 2021-09-24
Expose sudo powers on Router we give to Modules

ethanfrey:
I came across this need when trying to implement a custom router that could perform privileged operations on contracts.

It actually seems like it can be generalised for all "governance proposals" and such as done in the sdk, which is a privileged interface to modules.

I implemented this for Wasm::Sudo and Bank::Mint

I also added a test for a CustomModule that makes these permissions calls to mint tokens (also needed by staking module) and came across a lot of usability issues with the UI, which I was able to sort out.

I am getting a much higher confidence we can start plugging in external implementations.

### DIFF
--- a/packages/multi-test/src/app.rs
+++ b/packages/multi-test/src/app.rs
@@ -5,7 +5,7 @@ use anyhow::Result as AnyResult;
 use cosmwasm_std::testing::{mock_env, MockApi, MockStorage};
 use cosmwasm_std::{
     from_slice, to_vec, Addr, Api, Binary, BlockInfo, ContractResult, CustomQuery, Empty, Querier,
-    QuerierResult, QuerierWrapper, QueryRequest, Storage, SystemError, SystemResult,
+    QuerierResult, QuerierWrapper, QueryRequest, StdResult, Storage, SystemError, SystemResult,
 };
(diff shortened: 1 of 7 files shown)


packages/multi-test/src/app.rs
```diff
@@ -665,6 +665,33 @@ where
     }
 }
 
+// TODO: extend this later to give privileged access to various modules
```
ueco-jb, line 668: Just leaving a comment to create later issue for this one.

ueco-jb (dismissed):
lgtm

packages/multi-test/src/wasm.rs
```diff
@@ -862,13 +877,16 @@ mod test {
...
-        FailingModule<Empty, Empty>,
-        WasmKeeper<Empty, Empty>,
+        FailingModule<ExecC, QueryC, Empty>,
+        WasmKeeper<ExecC, QueryC>,
         FailingStaking,
         FailingDistribution,
-    > {
+    >;
+
+    fn mock_router() -> BasicRouter {
```
ueco-jb, line 889: I like this!

ueco-jb (approved):
lgtm

PR #459, opened 2021-10-01
Fix improper assert_matches usage

ueco-jb:
By mistake I used incorrect pattern while adding `assert_matches` there.


hashedone (approved):
lgtm

Issue #464, opened 2021-10-01
Various `range_de` / `prefix_de` improvements


maurolacy:
A number of slight improvements to the current impl:

- [x] ~~Replace `KeyDeserializer::from_slice(&[u8])` with `KeyDeserializer::from_vec(Vec<u8>)`. When we call it, we can consume the Vec and save a clone. See [deserialize_kv](https://github.com/CosmWasm/cw-plus/pull/432/files#diff-b74fe4efaced9c75072d3e90271006a9952c20fb129a418deb54561d5b67ed96R22).~~ (done in #467)
- [x] ~~Make a `K::Owned: Deserializable` trait attribute. `String`, `Vec<u8>` would have `type Deserializable = Self`. `&str` would have `type Deserializable = String`, for example. The idea behind this is to try to avoid having to declare `K` as `Deserializable`, that is, moving / enforcing the `Deserializable` bound to `K::Output` instead.~~ (won't do)
- [x] ~~Add a `from_length_prefixed` helper, and use it / call it where needed~~. (not needed with key deserialization already working) 

#432 follow-up.

ethanfrey:
> Replace KeyDeserializer::from_slice(&[u8]) with KeyDeserializer::from_vec(Vec<u8>). When we call it, we can consume the Vec and save a clone.

I would be happy to see this as a small PR first, and fit it in 0.10.0. The rest can follow later.

maurolacy:
> > Replace KeyDeserializer::from_slice(&[u8]) with KeyDeserializer::from_vec(Vec). When we call it, we can consume the Vec and save a clone.
> 
> I would be happy to see this as a small PR first, and fit it in 0.10.0. The rest can follow later.

Working on this, but my take is that a method `from_vec(Vec<u8>)` is similar / equivalent to from_slice(&[u8])`, in terms of memory management / number of clones.

In the first case, we pass an owned value, that we can use / consume. But, this value is being copied as argument, when calling the function.
In the second case, we pass a reference, but then we need to clone it when we want to consume it / return it.

A third option would be to have a signature like `from_vec(&Vec<u8>)`; but this is similar to passing a reference to a slice. In fact, a slice reference is simpler, because `try_from` works on it directly, when needed.

I can publish the PR if you want, but my opinion is we leave this just as it is. In the case of a "raw" deserialization to `Vec<u8>` we will have a useless call to the function, but still, no extra copy will be done.

Perhaps adding `#inline` can be used, to avoid the extra call.

[...]

PR #467, opened 2021-10-04
Key deserializer improvements

maurolacy:
Closes first part of #464.

### DIFF
--- a/packages/storage-plus/src/de.rs
+++ b/packages/storage-plus/src/de.rs
@@ -14,6 +14,7 @@ pub trait KeyDeserialize {
 impl KeyDeserialize for () {
     type Output = ();
 
+    #[inline]
     fn from_slice(_value: &[u8]) -> StdResult<Self::Output> {
         Ok(())
     }
@@ -22,6 +23,7 @@ impl KeyDeserialize for () {
 impl KeyDeserialize for Vec<u8> {
     type Output = Vec<u8>;
 
+    #[inline]
     fn from_slice(value: &[u8]) -> StdResult<Self::Output> {
         Ok(value.to_vec())
     }
@@ -30,6 +32,7 @@ impl KeyDeserialize for Vec<u8> {
 impl KeyDeserialize for &Vec<u8> {
     type Output = Vec<u8>;
 
+    #[inline]
     fn from_slice(value: &[u8]) -> StdResult<Self::Output> {
         <Vec<u8>>::from_slice(value)
     }
@@ -38,6 +41,7 @@ impl KeyDeserialize for &Vec<u8> {
 impl KeyDeserialize for &[u8] {
     type Output = Vec<u8>;
 
+    #[inline]
     fn from_slice(value: &[u8]) -> StdResult<Self::Output> {
         <Vec<u8>>::from_slice(value)
     }
@@ -46,6 +50,7 @@ impl KeyDeserialize for &[u8] {
 impl KeyDeserialize for String {
     type Output = String;
 
+    #[inline]
     fn from_slice(value: &[u8]) -> StdResult<Self::Output> {
         String::from_utf8(value.to_vec())
             // FIXME: Add and use StdError utf-8 error From helper
@@ -56,6 +61,7 @@ impl KeyDeserialize for String {
 impl KeyDeserialize for &String {
     type Output = String;
 
+    #[inline]
     fn from_slice(value: &[u8]) -> StdResult<Self::Output> {
         String::from_slice(value)
     }
@@ -64,6 +70,7 @@ impl KeyDeserialize for &String {
 impl KeyDeserialize for &str {
     type Output = String;
 
+    #[inline]
     fn from_slice(value: &[u8]) -> StdResult<Self::Output> {
         String::from_slice(value)
     }
@@ -72,6 +79,7 @@ impl KeyDeserialize for &str {
 impl KeyDeserialize for Addr {
     type Output = Addr;
 
+    #[inline]
     fn from_slice(value: &[u8]) -> StdResult<Self::Output> {
         Ok(Addr::unchecked(String::from_slice(value)?))
     }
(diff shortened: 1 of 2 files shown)


packages/storage-plus/src/de.rs
```diff
@@ -9,92 +9,153 @@ pub trait KeyDeserialize {
...
     fn from_slice(value: &[u8]) -> StdResult<Self::Output> {
         String::from_utf8(value.to_vec())
             // FIXME: Add and use StdError utf-8 error From helper
             .map_err(|err| StdError::generic_err(err.to_string()))
     }
+
+    #[inline(always)]
+    fn from_vec(value: Vec<u8>) -> StdResult<Self::Output> {
+        String::from_slice(value.as_slice())
+    }
```
maurolacy, line 84: Hmm, better to invert these impls.

packages/storage-plus/src/de.rs
```diff
@@ -56,6 +77,11 @@ impl KeyDeserialize for String {
             // FIXME: Add and use StdError utf-8 error From helper
             .map_err(|err| StdError::generic_err(err.to_string()))
     }
+
+    #[inline(always)]
+    fn from_vec(value: Vec<u8>) -> StdResult<Self::Output> {
```
ethanfrey, line 82: All these cases which just dispatch to the same implementation could be handled by a default trait implementation. This would also make it clearer where we are optimising.

Note that this case is a missed chance. It calls `String::from_utf8(value.to_vec())` right after `value.as_slice()`.

`value.as_slice().to_vec()` is a hidden and unneeded clone.

packages/storage-plus/src/de.rs
```diff
@@ -65,6 +91,11 @@ impl KeyDeserialize for &String {
     fn from_slice(value: &[u8]) -> StdResult<Self::Output> {
         String::from_slice(value)
     }
+
+    #[inline(always)]
+    fn from_vec(value: Vec<u8>) -> StdResult<Self::Output> {
+        String::from_slice(value.as_slice())
```
ethanfrey, line 97: I would delegate to `String::from_vec(value)`

packages/storage-plus/src/de.rs
```diff
@@ -74,6 +105,11 @@ impl KeyDeserialize for &str {
...
```
ethanfrey, line 111: ditto

packages/storage-plus/src/de.rs
```diff
@@ -83,6 +119,11 @@ impl KeyDeserialize for Addr {
     fn from_slice(value: &[u8]) -> StdResult<Self::Output> {
         Ok(Addr::unchecked(String::from_slice(value)?))
     }
+
+    #[inline(always)]
+    fn from_vec(value: Vec<u8>) -> StdResult<Self::Output> {
+        Addr::from_slice(value.as_slice())
```
ethanfrey, line 125: `String::from_utf8(value)` ?

ethanfrey (commented):
Lots of missed chances for optimisations.

Please just give a default trait implementation, and it is clear which ones you leave un-optimized.

I highlighted a number that could be optimised. (All String/Addr related stuff)

maurolacy:
Now it's better.

packages/storage-plus/src/de.rs
maurolacy: `Addr::from_vec`, just to keep the base impl unique.

packages/storage-plus/src/de.rs
maurolacy: Done.

packages/storage-plus/src/de.rs
```diff
@@ -8,18 +8,16 @@ use crate::keys::{IntKey, TimestampKey};
 pub trait KeyDeserialize {
     type Output: Sized;
 
-    fn from_slice(value: &[u8]) -> StdResult<Self::Output>;
     fn from_vec(value: Vec<u8>) -> StdResult<Self::Output>;
+
+    fn from_slice(value: &[u8]) -> StdResult<Self::Output> {
```
ethanfrey, line 13: nice to use this default.

packages/storage-plus/src/de.rs
```diff
@@ -191,18 +139,13 @@ impl<T: KeyDeserialize, U: KeyDeserialize> KeyDeserialize for (T, U) {
 
         Ok((T::from_slice(t)?, U::from_slice(u)?))
```
ethanfrey: hmmm.. could you use `from_vec` here as well?

I guess you would need [value.split_off](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.split_off) rather than `split_at`. But should be more consistent/efficient

packages/storage-plus/src/de.rs
```diff
@@ -221,11 +164,6 @@ impl<T: KeyDeserialize, U: KeyDeserialize, V: KeyDeserialize> KeyDeserialize for
 
         Ok((T::from_slice(t)?, U::from_slice(u)?, V::from_slice(v)?))
```
ethanfrey: Same as the (T, U) tuples (from_vec)

ethanfrey (commented):
Looks good.

Two last places that could be a bit more efficient, then happy to merge

packages/storage-plus/src/de.rs
maurolacy: Did it. It's a little unclear with the var names. And, not sure it's more efficient.

If we want a self-contained version, that is, one that doesn't depend on `from_slice`, we can always use `from_vec(t.to_vec())`

maurolacy: In fact, for triples, the version with `split_off()` has four vec allocations instead of three; because of the small vec for the length of `V. Plus, it requieres a conversion to `to_slice()` for calling `try_into()`.

I really like the previous version better.

ethanfrey (approved):
great!

maurolacy:
So, I guess you prefer the version with `split_off` then?

ethanfrey:
> The version with split_at is better IMO.

Why is it better? Easier to read code?

maurolacy:
OK then.

PR #470, opened 2021-10-04
Improve error handling / remove FIXMEs

maurolacy:
Improve utf-8 error handling.

Remove FIXMEs for `SliceError`s.

### DIFF
--- a/packages/storage-plus/src/de.rs
+++ b/packages/storage-plus/src/de.rs
@@ -56,9 +56,7 @@ impl KeyDeserialize for String {
 
     #[inline(always)]
     fn from_vec(value: Vec<u8>) -> StdResult<Self::Output> {
-        String::from_utf8(value)
-            // FIXME: Add and use StdError utf-8 error From helper
-            .map_err(|err| StdError::generic_err(err.to_string()))
+        String::from_utf8(value).map_err(StdError::invalid_utf8)
     }
 }
 
@@ -106,7 +104,6 @@ macro_rules! integer_de {
             #[inline(always)]
             fn from_vec(value: Vec<u8>) -> StdResult<Self::Output> {
                 Ok(<$t>::from_be_bytes(value.as_slice().try_into()
-                    // FIXME: Add and use StdError try-from error From helper
                     .map_err(|err: TryFromSliceError| StdError::generic_err(err.to_string()))?))
             }
         })*
@@ -134,7 +131,6 @@ impl<T: KeyDeserialize, U: KeyDeserialize> KeyDeserialize for (T, U) {
             value
                 .as_slice()
                 .try_into()
-                // FIXME: Add and use StdError try-from error From helper
                 .map_err(|err: TryFromSliceError| StdError::generic_err(err.to_string()))?,
         ) as usize;
         let u = tu.split_off(t_len);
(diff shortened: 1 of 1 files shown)


packages/storage-plus/src/de.rs
```diff
@@ -106,7 +104,6 @@ macro_rules! integer_de {
             #[inline(always)]
             fn from_vec(value: Vec<u8>) -> StdResult<Self::Output> {
                 Ok(<$t>::from_be_bytes(value.as_slice().try_into()
-                    // FIXME: Add and use StdError try-from error From helper
                     .map_err(|err: TryFromSliceError| StdError::generic_err(err.to_string()))?))
```
webmaster128, line 107: Arg. Here you can't do the same

```
.map_err(StdError::generic_err)?
```

because the function takes `Into<String>` instead of `ToString`?

packages/storage-plus/src/de.rs
```diff
@@ -56,9 +56,7 @@ impl KeyDeserialize for String {
 
     #[inline(always)]
     fn from_vec(value: Vec<u8>) -> StdResult<Self::Output> {
-        String::from_utf8(value)
-            // FIXME: Add and use StdError utf-8 error From helper
-            .map_err(|err| StdError::generic_err(err.to_string()))
+        String::from_utf8(value).map_err(StdError::invalid_utf8)
```
ethanfrey, line 59: nice

packages/storage-plus/src/de.rs
```diff
@@ -134,7 +131,6 @@ impl<T: KeyDeserialize, U: KeyDeserialize> KeyDeserialize for (T, U) {
             value
                 .as_slice()
                 .try_into()
-                // FIXME: Add and use StdError try-from error From helper
```
ethanfrey, line 137: All these changes are the same lines:

```rust
u16::from_be_bytes(
             value
                 .as_slice()
                 .try_into()
                 .map_err(|err: TryFromSliceError| StdError::generic_err(err.to_string()))?,
         ) as usize;
```

Maybe they can be their own helper function as well, so we just call

`let t_len = parse_length(&value)?;`

ethanfrey (approved):
Good fix, but can you also remove those duplicates?

packages/storage-plus/src/de.rs
webmaster128: And once you do that, you can ignore the content of `err` and make it use case specific, like

```
.map_err(|_err| StdError::generic_err("Could not read 2 byte length"))?,
```

maurolacy: That's not easy to do, because `parse_length` would have to be generic over the integers, and I don't have an associated trait for it.

maurolacy: Already mentioned it in #432: https://github.com/CosmWasm/cw-plus/pull/432#discussion_r720134129

packages/storage-plus/src/de.rs
maurolacy: ✔️

packages/storage-plus/src/de.rs
webmaster128: But in this current diff it is always `u16::from_be_bytes`

maurolacy: Ah, you're right! Will do.

maurolacy:
Approval was automatically dismissed or so.

ethanfrey (approved):
Awesome. 

BTW, at the backlog meeting during your vacation the other devs all asked for strict protected branch rules... no push on Master, no merge pr without approval and CI checks passing, dismiss stale reviews.

I applied on all rust repos I maintain. We can discuss these rules if you don't like them, happy to implement what the team requests

maurolacy:
I agree with those rules, except perhaps dismiss stale reviews. An author can always request a new review over an already approved PR, if he deems it necessary.

PR #491, opened 2021-10-14
Fix cw721 query_all_tokens

giansalex:
`token_id` is not a valid address, `maybe_addr` throw an error

### DIFF
--- a/contracts/cw721-base/src/query.rs
+++ b/contracts/cw721-base/src/query.rs
@@ -113,8 +113,7 @@ where
         limit: Option<u32>,
     ) -> StdResult<TokensResponse> {
         let limit = limit.unwrap_or(DEFAULT_LIMIT).min(MAX_LIMIT) as usize;
-        let start_addr = maybe_addr(deps.api, start_after)?;
-        let start = start_addr.map(|addr| Bound::exclusive(addr.as_ref()));
+        let start = start_after.map(Bound::exclusive);
 
         let tokens: StdResult<Vec<String>> = self
             .tokens


ethanfrey (approved):
good catch

evandekieft:
Hi @ethanfrey @giansalex i know this is an old PR but I have just encountered this buggy behavior in some older NFTs deployed on Terra (including Galactic Punks and LunaBulls). You can’t paginate them. The all_tokens call always fails since it’s expecting an address.

Is there any way to work around this from the query client side? Doesn’t this make it impossible to get the full set of token ids using on chain data?

kasunindikaliyanage:
> Hi @ethanfrey @giansalex i know this is an old PR but I have just encountered this buggy behavior in some older NFTs deployed on Terra (including Galactic Punks and LunaBulls). You can’t paginate them. The all_tokens call always fails since it’s expecting an address.
> 
> Is there any way to work around this from the query client side? Doesn’t this make it impossible to get the full set of token ids using on chain data?

I am also facing similar issue for HellCats, DystopAI, rekt wolf collections as well. 

Following is the code segment I have used to retrieve the tokens for a given collection.

`  const response = await this.lcdClient.wasm.contractQuery(contractAddr,
{
        "all_tokens": {
          "limit": limit, 
          "start_after" : lastNftFromQuery
        },
})`

This code segment worked so far only for NIP Terra collection (terra1wue25pe63g22uxy6ssjrkhnjscshdv8ztjwz5y)
Any suggestions would be greatly helpful

ethanfrey:
not sure.

I just made this as a demo assuming people would get involved before using it in production.

there are more maintained contracts, worked on by actual nft projects, in cw-nfts repo

maurolacy:
To fix this for already instantiated contracts, those contracts would need to be upgraded to a new version with the fix.

maurolacy:
Another option could be, to try and add a suffix to the `start_after` param, so that the address validation check pass. Not sure that hack would work for all cases, or if it would work at all.

giansalex:
@evandekieft I'm not sure about Terra, but in cosmwasm v1, you can query raw state, and get this info:
```json
[
   {
      "key":[
         "tokens",
         "0A2C58673DB644FE" // token_id
      ],
      "value":{
         "owner":"juno194v8uwee2fvs2s8fa5k7j03ktwc87h5yyvnaj4",
         "approvals":[
            
         ],
         "token_uri":"ipfs://QmVGWXcvevARusYfCf7FXrchEor6FswDcgetjVQyjnfuuE",
         "extension":null
      }
   }
]
```
https://gist.github.com/giansalex/50ddf9b9cad130d6d72b0a75b065b3f8

Issue #461, opened 2021-10-01
Add `range_de` to `Map`-like structs


maurolacy:
Let's extend range deserialization to map-like `IndexedMap`, `SnapshotMap` and `IndexedSnapshotMap`.

#432 follow-up.

maurolacy:
This is partially addressed by #498 and #500, but there are still a number of missing impls:

`Map` requires:

- `sub_prefix_range_de`, for completeness.

`SnapshotMap` requires:

- `prefix_de`
- `sub_prefix_de`
- `keys_de`
- `prefix_range_de`
- `sub_prefix_range_de`

`IndexedMap` requires:

- `prefix_de`
- `sub_prefix_de`
- `sub_prefix_range_de`

`IndexedSnapshotMap` requires:

- `range_de`
- `prefix_de`
- `prefix_range_de`
- `sub_prefix_de`
- `sub_prefix_range_de`
- `keys_de`

PR #498, opened 2021-10-18
range_de for IndexMap

uint:
Part of #461


packages/storage-plus/src/indexed_map.rs
```diff
@@ -730,6 +793,225 @@ mod test {
         assert_eq!(datas[1], marias[1].1);
     }
 
+    #[test]
+    #[cfg(feature = "iterator")]
+    fn range_de_simple_string_key() {
+        let mut store = MockStorage::new();
+        let map = build_map();
```
maurolacy, line 800: Here it would be better to test it against a map with a pk that does not deserialize to `Vec<u8>`. Either `String`, `&str`, or even better, one Uint key (`Uint8` or `Uint32`), would work.

You would have to either adapt the non-deserializing tests, or create a different map here (i.e. a `build_map_de()` helper). I think the first option is slightly more work but it's better, as we'll be able to clearly show the difference between `range` and `range_de` result types.

maurolacy (dismissed):
Base `Map` impl.

Good to merge. It would be good to improve / clarify tests first.

maurolacy (approved):
Nice.

PR #499, opened 2021-10-19
cw1-subkeys-ng: Basic implementation

### LINKED ISSUE #494: Reimplement cw1-whitelist contract in terms of semantic structures
Part of #493 
* [x] Define trait with separate function for each exec and query variant handling contract execution
* [x] Define dispatch function dispatching given exec message to the defined trait
* [x] Define dispatch function dispatching given query message to defined trait
* [x] Encapsulate contract state into structure
* [x] Implement defined trait on contract structure
* [x] Make entry points construct the contract structure (or make structure global variable on contract if it is cheaper and possible), then dispatch on it using defined functions
* [x] Provide helpers for query
* [x] Provide helpers for tests
* [x] Migrate tests from cw1-whitelist
* [x] Schema generation

hashedone:
Closes #494


ethanfrey:
Starting review now... I'm excited 😁

contracts/cw1-whitelist-ng/src/state.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+impl Cw1WhitelistContract {
+    pub const fn new() -> Self {
+        Self {
+            admin_list: Item::new("admin_list"),
+        }
+    }
+
+    pub const fn admin_list(&self) -> &Item<AdminList> {
```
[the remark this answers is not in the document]
hashedone: I removed it for now, as it made more problems that gave benefits, but basically any changes of state on contract should be done by its API. I described why in another comments. If "subcontract" want to customize the item itself it means it is not subcontract and should not share implementation.

contracts/cw1-whitelist-ng/src/interfaces.rs
```diff
@@ -0,0 +1,40 @@
...
+
+    fn update_admins(
+        &self,
+        deps: DepsMut,
+        env: Env,
+        info: MessageInfo,
+        admins: Vec<String>,
+    ) -> Result<Response<T>, Self::Error>;
+
+    fn admin_list(&self, deps: Deps, env: Env) -> Result<AdminListResponse, Self::Error>;
```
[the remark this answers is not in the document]
hashedone: I completely don't see a reason why `Whitelist` would have any relation to `Cw1`. I can have whitelist restricting any other contract, those are two separate ideas.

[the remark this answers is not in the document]
ethanfrey: Fair enough. No need to have the traits extend each other

Let's get it to work now as one trait.
Having it as two separate traits and showing how the messages work with that would be a great demo to show composition. This can be a future step, just see how you can demo that even in one contract.

contracts/cw1-whitelist-ng/src/contract.rs
```diff
@@ -0,0 +1,440 @@
...
+        &self,
+        deps: DepsMut,
+        env: Env,
+        info: MessageInfo,
+        msg: &[u8],
+    ) -> Result<Response<T>, ContractError>
+    where
+        T: DeserializeOwned,
+    {
+        let mut errs = vec![];
```
ethanfrey, line 72: Interesting....

So it tries to parse into each enum and returns all the error messages appended.
I would be curious to see a test that asserted the error string when some garbage like `{"foo": "bar"}` was passed in. (I know, asserting error strings is bad, maybe doc test?? something to show this doesn't look too terrible).

Minor point - can be a follow up PR.

contracts/cw1-whitelist-ng/src/contract.rs
```diff
@@ -0,0 +1,440 @@
...
+
+        match from_slice::<Cw1ExecMsg<T>>(msg) {
```
ethanfrey, line 74: This multiple attempts to deserialize is a nice approach. It looks a little ugly here, but I see the beauty of each message being able to implement dispatch and the contract being able to implement multiple traits.

If this is auto-generated (which it will be shortly), then I definitely see the bonus.

contracts/cw1-whitelist-ng/src/contract.rs
```diff
@@ -0,0 +1,440 @@
...
+                deps.as_mut(),
+                mock_env(),
+                mock_info(&alice, &[]),
+                msgs.clone(),
+            )
+            .unwrap();
+
+        assert_eq!(
+            res.messages,
+            msgs.into_iter().map(SubMsg::new).collect::<Vec<_>>()
```
ethanfrey, line 375: This chain could make a nice helper function (in cosmwasm-std or cw-plus).
Or not... just a thought

contracts/cw1-whitelist-ng/src/contract.rs
```diff
@@ -0,0 +1,440 @@
...
+            .unwrap();
+        assert_eq!(
+            res.messages,
+            msgs.into_iter().map(SubMsg::new).collect::<Vec<_>>()
+        );
+        assert_eq!(res.attributes, [("action", "execute")]);
+    }
+
+    #[test]
+    fn execute_custom_messages_works() {
```
ethanfrey, line 351: Thanks for adding this test to show T is properly supported

contracts/cw1-whitelist-ng/src/interfaces.rs
```diff
@@ -0,0 +1,44 @@
+use crate::msg::AdminListResponse;
+use cosmwasm_std::{CosmosMsg, Deps, DepsMut, Env, MessageInfo, Response};
+use cw1::query::CanExecuteResponse;
+
+pub trait Cw1<T> {
```
ethanfrey, line 5: Great to show support for multiple interfaces.

contracts/cw1-whitelist-ng/src/state.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+pub struct Cw1WhitelistContract<T> {
+    // I am pretty sure that just form this with some proper hint attributes it would be possible
+    // to provide helpers for raw queries, this might be fun idea
+    pub(crate) admin_list: Item<'static, AdminList>,
+    _msg: PhantomData<T>,
+}
+
+impl Cw1WhitelistContract<Empty> {
```
ethanfrey, line 34: Okay, this is to fix the T we use in this contract as opposed to the generic T we allow for anyone who imports as a library to extend.

I see the point. I wonder if this makes sense to be in lib.rs next to the entry points.
Some contract importing it would just need to copy lib.rs and edit it to allow custom messages.

contracts/cw1-whitelist-ng/src/query.rs
```diff
@@ -0,0 +1,57 @@
+// This whole thing is definitely to be easly generated from trait itself
+
+use cosmwasm_std::{Addr, CosmosMsg, CustomQuery, QuerierWrapper, StdResult};
+use cw1::CanExecuteResponse;
+use serde::Serialize;
+
+use crate::msg::{AdminListResponse, Cw1QueryMsg, WhitelistQueryMsg};
+
+#[must_use]
+pub struct Cw1Querier<'a, C>
```
ethanfrey, line 10: Nice implementation of Queriers. Look simpler and cleaner.

contracts/cw1-whitelist-ng/src/multitest/contract.rs
```diff
@@ -0,0 +1,239 @@
...
+            addr,
+            app,
+            sender,
+            send_funds,
+        }
+    }
+
+    pub fn execute<C>(self, msgs: Vec<CosmosMsg<C>>) -> AnyResult<AppResponse>
+    where
+        C: Clone + std::fmt::Debug + PartialEq + JsonSchema + Serialize + 'static,
```
ethanfrey, line 78: This should be `CustomMsg` type now (maybe we need to update in cw-plus now this has landed in cosmwasm-std)

Nice use of late binding of where clause.

contracts/cw1-whitelist-ng/src/multitest/contract.rs
```diff
@@ -0,0 +1,239 @@
...
+#[must_use]
+pub struct WhitelistExecutor<'a, App> {
+    addr: Addr,
+    app: &'a mut App,
+    sender: Addr,
+    send_funds: &'a [Coin],
+}
+
+impl<'a, App> WhitelistExecutor<'a, App> {
+    pub fn new(addr: Addr, app: &'a mut App, sender: Addr, send_funds: &'a [Coin]) -> Self {
```
ethanfrey, line 100: This is a nice pattern. Always has the same `XyzExecutor.new()` args, then contract-specific function call.

contracts/cw1-whitelist-ng/src/multitest/contract.rs
```diff
@@ -0,0 +1,239 @@
...
+        self.label = label.to_owned();
+        self
+    }
+
+    pub fn with_admin(mut self, admin: &str) -> Self {
+        self.admin = Some(admin.to_owned());
+        self
+    }
+
+    pub fn with<C>(self, admins: Vec<String>, mutable: bool) -> AnyResult<Cw1WhitelistProxy>
```
ethanfrey, line 169: I don't like this name. The other optional `with_{admin, label}` methods are great.

This should have a more different name. Maybe `do`? I'm sure you can come up with something better...

`let contract = Instantiator::new(....).with(...).unwrap();` looks a bit weird to me,.

contracts/cw1-whitelist-ng/src/multitest/contract.rs
```diff
@@ -0,0 +1,239 @@
...
+// Proxy for direct execution in multitest.
+#[derive(Clone, PartialEq, Debug)]
+pub struct Cw1WhitelistProxy(Addr);
+
+impl Cw1WhitelistProxy {
+    pub fn addr(&self) -> Addr {
+        self.0.clone()
+    }
+
+    pub fn instantiate<'a, App>(
```
ethanfrey, line 196: Hmmm... so I would really call it.

`let contract = Cw1WhitelistProxy::instantiate(...).with(...).unwrap();`

That makes more sense, but still... I could imagine a better name.
Not going to block merge with it, but happy for more naming feedback before this name gets ported to tons of auto-generated code.

contracts/cw1-whitelist-ng/src/multitest/contract.rs
hashedone: I missed its existence there, will change.

contracts/cw1-whitelist-ng/src/multitest/contract.rs
hashedone: I don't like `do` to much. I see your point, I somehow agree. Maybe `call`? `perform`? `with_args`? I am struggling. Maybe someone have an idea?

contracts/cw1-whitelist-ng/src/contract.rs
hashedone: I am struggling here. Before it was done by just dumping as an error the error message (not even `ParseErr` was constructed). I would actually prefer to return error looking like:
```json
{
  "parse_err": {
    "Cw1ExecMsg": "...",
    "WhitelistExecMsg": "...",
  }
}
```

This actually can be done by providing additional custom error type which is either `DispatchErr(HashMap<String, StdError>)` (this very case) or `ContractErr(Self::Error)` (for failing in contract after dispatching succeed. For now I think this would be the best approach, but didn't want to put it here yet as I am polishing this idea.

hashedone:
Or as I am thinking about it... there are no so bad things there, so I am merging to not asking for reaproval of this monster on every rebase. Tomorrow morning I would make a PR with minor changes (in particular introduce `CustomMsg`), and there would be good place to think about better naming without Rust.

PR #500, opened 2021-10-20
UniqueIndex range_de

uint:
Deals with #461


packages/storage-plus/src/indexes.rs
```diff
@@ -323,7 +386,15 @@ where
...
-    Ok((t.pk.into(), t.value))
+    Ok((t.pk.to_vec(), t.value))
+}
+
+fn deserialize_unique_kv_de<T: DeserializeOwned, K: KeyDeserialize>(
+    kv: Record,
+) -> StdResult<(K::Output, T)> {
+    let (_, v) = kv;
+    let t = from_slice::<UniqueRef<T>>(&v)?;
+    Ok((K::from_vec(t.pk.to_vec())?, t.value))
```
maurolacy, line 397: This code block is correct, but the test below is failing because `K` here is the **unique index** key type (`U32Key` in the test), not the primary key type (`&[u8]` in the test).

I think you will need to modify `UniqueIndex` to pass the primary key type (a `PK` type parameter) to it, and use that when instantiating `deserialize_unique_kv`.

packages/storage-plus/src/indexes.rs
```diff
@@ -394,3 +465,66 @@ where
         self.no_prefix().keys(store, min, max, order)
     }
 }
+
+#[cfg(feature = "iterator")]
+impl<'a, K, T> UniqueIndex<'a, K, T>
```
maurolacy, line 470:
```suggestion
impl<'a, PK, K, T> UniqueIndex<'a, PK, K, T>
```

packages/storage-plus/src/indexes.rs
```diff
@@ -394,3 +465,66 @@ where
...
+where
+    T: Serialize + DeserializeOwned + Clone,
+    K: PrimaryKey<'a> + KeyDeserialize,
```
maurolacy, line 473:
```suggestion
    K: PrimaryKey<'a>,
    PK: KeyDeserialize,
```

and so on.

packages/storage-plus/src/indexes.rs
```diff
@@ -394,3 +465,66 @@ where
...
+    where
+        T: 'c,
+        K::Output: 'static,
+    {
+        self.no_prefix_de().keys_de(store, min, max, order)
+    }
+
+    fn no_prefix_de(&self) -> Prefix<K, T> {
+        Prefix::with_deserialization_function(self.idx_namespace, &[], &[], |_, _, kv| {
+            deserialize_unique_kv_de::<_, K>(kv)
```
maurolacy, line 527:
```suggestion
    fn no_prefix_de(&self) -> Prefix<PK, T> {
        Prefix::with_deserialization_function(self.idx_namespace, &[], &[], |_, _, kv| {
            deserialize_unique_kv_de::<_, PK>(kv)
```

and so on.

packages/storage-plus/src/indexed_map.rs
```diff
@@ -793,6 +826,35 @@ mod test {
...
+        let map = build_map();
+
+        // save data
+        let (pks, datas) = save_data(&mut store, &map);
+
+        let res: StdResult<Vec<_>> = map
+            .idx
+            .name_lastname
+            .prefix(b"Maria".to_vec())
+            .range_de(&store, None, None, Order::Ascending)
```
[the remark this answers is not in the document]
maurolacy: `Prefix` must work with the key of the index in this case (that's what you are prefixing). In this impl, what must be changed is the return value of `Prefix`. Instead of being `Suffix`, it has to be `PK` (for `UniqueIndex`).

Not sure about this, by the way. Will have to take a detailed look to answer further, in a more detailed or precise way.

What we can do is to leave this on hold, documenting the issues, and take a second attempt at it in the future.

ethanfrey:
I agree this is mergeable.
Some follow up is clear.

Let's find time to discuss architecture for the larger follow-up issues before tackling them (but feel free to make issues in the meantime)

ethanfrey:
Oh, man, there are like 5 PRs built on top of this, right?

maurolacy:
Yes, I'll rebase everything once the int key changes are merged to main.

maurolacy:
> Can you merge main in sometime (no more rebase possible)

Just rebased it without issues.

ethanfrey:
Cool. Want to merge this first, then the other branches on top?

maurolacy:
Sure.

maurolacy:
OK, all the follow-up issues / comments are addressed, are changes are merged / integrated here.

This is also rebased from master.

Let's merge this.

packages/storage-plus/src/indexes/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+pub use unique::UniqueIndex;
+
+use serde::de::DeserializeOwned;
+use serde::Serialize;
+
+use cosmwasm_std::{StdResult, Storage};
+
+use crate::U32Key;
+
+pub fn index_string(data: &str) -> Vec<u8> {
```
uint, line 16: These are test helpers, right? This isn't really relevant to the PR, but shouldn't we compile these only in `test` builds?

packages/storage-plus/src/indexed_snapshot.rs
```diff
@@ -212,12 +213,90 @@ where
...
+}
+
+#[cfg(feature = "iterator")]
+impl<'a, K, T, I> IndexedSnapshotMap<'a, K, T, I>
+where
+    T: Serialize + DeserializeOwned + Clone,
+    K: PrimaryKey<'a> + KeyDeserialize,
+    I: IndexList<T>,
+{
+    /// while range_de assumes you set the prefix to one element and call range over the last one,
```
ueco-jb, line 239: I'd be inclined to rephrase this part, personally I'm not sure I follow:
> assumes you set the prefix to one element

meaning assumes that each element has prefix set?
and 
> call range over the last one

meaning call range over last one (element)? So since last element?

uint:
LGTM overall. Does what we need it to!

packages/storage-plus/src/indexes/mod.rs
maurolacy: Those are helpers that _could_ be used from contract code. They are in fact exported in `lib.rs`. They are not very general, but still.

ueco-jb (approved):

packages/storage-plus/src/indexed_snapshot.rs
maurolacy: The difference is somewhat subtle: `range` over a `prefix` only iterates over all the elements with **that** prefix. Whereas `prefix_range` allows you to set the range directly over the prefix. Meaning prefix can now vary (inclusively or exclusively, depending on `PrefixBound`) between the `min` and `max` `prefix_range` parameters, if specified.

maurolacy: I'll try to clarify this in the comment.

PR #502, opened 2021-10-21
Parse reply helpers

### LINKED ISSUE #480: Add helpers for parsing the protobuf MsgInstantiate and MsgExecute responses
When we call into another contract using `SubMsg::reply_on_success`, we get data fields with important info from the other contract. Especially the address on instantiate.

Let's provide some helper functions for this as it is a common task that needs a bit of knowledge of the internals.

Maybe multiple implementations - one to parse from the data field, the other from the events fields. 

Events [defined here](https://github.com/CosmWasm/wasmd/blob/master/EVENTS.md#standard-events-in-xwasm) See the `instantiate` and `execute` events.

You can alternately (also) extract from the data field. That is a protobuf encoded message:

[InstantiateResponse](https://github.com/CosmWasm/wasmd/blob/master/proto/cosmwasm/wasm/v1/tx.proto#L65-L71)

[ExecuteResponse](https://github.com/CosmWasm/wasmd/blob/master/proto/cosmwasm/wasm/v1/tx.proto#L88-L92)

It seems very heavy to import protobuf libs here. Best would be to manually decode this. (Write 2 functions to parse those particular protobuf formats and test them against a test suite using eg prost/protobuf as dev dependency)

There is working code for this in tgrade-valset (when it calls tg4-engagement). The point here is to make this general.

maurolacy:
Closes #480.

Events helpers not implemented, but they don't require protobuf, so, can be deserialized directly. In any case, we can add them in a follow-up.

### DIFF
--- a/packages/cw0/src/parse_reply.rs
+++ b/packages/cw0/src/parse_reply.rs
@@ -1,0 +1,83 @@
+use prost::Message;
+use thiserror::Error;
+
+use cosmwasm_std::Reply;
+
+#[derive(Clone, PartialEq, Message)]
+pub struct MsgInstantiateContractResponse {
+    #[prost(string, tag = "1")]
+    pub contract_address: ::prost::alloc::string::String,
+    #[prost(bytes, tag = "2")]
+    pub data: ::prost::alloc::vec::Vec<u8>,
+}
+
+pub fn parse_reply_instantiate_data(
+    msg: Reply,
+) -> Result<MsgInstantiateContractResponse, ParseReplyError> {
+    let id = msg.id;
+    let res: MsgInstantiateContractResponse = Message::decode(
+        msg.result
+            .into_result()
+            .map_err(ParseReplyError::SubMsgFailure)?
+            .data
+            .ok_or_else(|| ParseReplyError::ParseFailure {
+                id,
+                err: "Missing reply data".to_owned(),
+            })?
+            .as_slice(),
+    )
+    .map_err(|err| ParseReplyError::ParseFailure {
+        id,
+        err: err.to_string(),
+    })?;
+
+    Ok(res)
+}
+
+#[derive(Error, Debug, PartialEq)]
+pub enum ParseReplyError {
+    #[error("Failure response from sub-message: {0}")]
+    SubMsgFailure(String),
+
+    #[error("Invalid reply from sub-message {id}: {err}")]
+    ParseFailure { id: u64, err: String },
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use cosmwasm_std::{ContractResult, SubMsgExecutionResponse};
+
+    #[test]
+    fn parse_reply_instantiate_data_works() {
+        let instantiate_reply_data: &str = "Contract #1";
+        let instantiate_reply = MsgInstantiateContractResponse {
+            contract_address: instantiate_reply_data.to_string(),
+            data: vec![],
+        };
+        let mut encoded_instantiate_reply =
+            Vec::<u8>::with_capacity(instantiate_reply.encoded_len());
+        // The data must encode successfully
+        instantiate_reply
+            .encode(&mut encoded_instantiate_reply)
+            .unwrap();
+
+        // Build reply message
+        let msg = Reply {
+            id: 1,
+            result: ContractResult::Ok(SubMsgExecutionResponse {
+                events: vec![],
+                data: Some(encoded_instantiate_reply.into()),
+            }),
+        };
+
+        let res = parse_reply_instantiate_data(msg).unwrap();
+        assert_eq!(
+            res,
+            MsgInstantiateContractResponse {
+                contract_address: instantiate_reply_data.into(),
+                data: vec![],
+            }
+        );
+    }
+}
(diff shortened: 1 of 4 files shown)


packages/cw0/src/parse_reply.rs
```diff
@@ -0,0 +1,214 @@
...
+
+/// Helper function to parse length-prefixed protobuf fields.
+/// The remaining of the data is kept in the data parameter.
+fn parse_protobuf_length_prefixed(data: &mut Vec<u8>) -> Result<Vec<u8>, ParseReplyError> {
+    if data.is_empty() {
+        return Ok(vec![]);
+    };
+    let mut rest_1 = data.split_off(1);
+    let wire_type = data[0] & 0b11;
+    let field = data[0] >> 3;
```
ethanfrey, line 27: When parsing, we know which field we expect.
It would be nice to pass this in as an argument and return an error if it is a different field.

Thus, 

```
let address = parse_protobuf_string(data, 1)?;
let data = parse_protobuf_bytes(data, 2)?;
```

Nice that this returns `Ok(vec![])` if the data was empty (no field 2 present)

packages/cw0/src/parse_reply.rs
```diff
@@ -0,0 +1,214 @@
...
+    let data = msg
+        .result
+        .into_result()
+        .map_err(ParseReplyError::SubMsgFailure)?
+        .data
+        .ok_or_else(|| ParseReplyError::ParseFailure("Missing reply data".to_owned()))?;
+    // Manual protobuf decoding
+    let mut data = data.0;
+    // Parse contract addr
+    let contract_addr = parse_protobuf_string(&mut data)?;
```
ethanfrey, line 86: Yup, asserting field numbers here would be a bit safer and just one check above (you already get the field)

packages/cw0/src/parse_reply.rs
```diff
@@ -0,0 +1,214 @@
...
+        )));
+    }
+    if rest_1.is_empty() {
+        return Err(ParseReplyError::ParseFailure(format!(
+            "failed to decode Protobuf message: field #{}: message too short",
+            field
+        )));
+    }
+    let mut rest_2 = rest_1.split_off(1);
+    let len = rest_1[0] as usize;
```
ethanfrey, line 42: This is valid for sizes <= 127 bytes.

You can either:

* add a caveat for that as rust comment and issue to follow up (but return error if this is multi-byte - if `rest1[0] & 0x80 != 0`)
* support this now as defined here: https://developers.google.com/protocol-buffers/docs/encoding#varints

packages/cw0/src/parse_reply.rs
```diff
@@ -0,0 +1,214 @@
...
+    pub struct MsgExecuteContractResponse {
+        #[prost(bytes, tag = "1")]
+        pub data: ::prost::alloc::vec::Vec<u8>,
+    }
+
+    #[test]
+    fn parse_reply_instantiate_data_works() {
+        let contract_addr: &str = "Contract #1";
+        for data in [vec![], vec![1u8, 2, 255, 7, 5]] {
+            let expected = if data.is_empty() {
```
ethanfrey, line 148: If you are doing a loop ("table-tests" in Go speak), I would define input and expected to in a `vec![struct testData]` and then just test it. Building the expected based on the responses is a bit odd to read.

packages/cw0/src/parse_reply.rs
```diff
@@ -0,0 +1,214 @@
...
+    FromUtf8(#[from] std::string::FromUtf8Error),
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use cosmwasm_std::{ContractResult, SubMsgExecutionResponse};
+    use prost::Message;
+
+    #[derive(Clone, PartialEq, Message)]
```
ethanfrey, line 130: Nice to see these super-lightweight proto defs and prost just in dev-dependencies

packages/cw0/src/parse_reply.rs
```diff
@@ -0,0 +1,214 @@
...
+    pub struct MsgInstantiateContractResponse {
+        #[prost(string, tag = "1")]
+        pub contract_address: ::prost::alloc::string::String,
+        #[prost(bytes, tag = "2")]
+        pub data: ::prost::alloc::vec::Vec<u8>,
+    }
+
+    #[derive(Clone, PartialEq, Message)]
+    pub struct MsgExecuteContractResponse {
```
ethanfrey, line 139: It would be nice to have a few lower-level unit tests with these structs as well... also failing on bad data parsing (if we encode some unknown struct the parser returns an error).

Just covering `parse_protobuf_string` and `parse_protobuf_bytes`

packages/cw0/src/parse_reply.rs
maurolacy: By the way, I thought this is was an inefficient / inconvenient format for length encoding, and this seems to confirm it: https://news.ycombinator.com/item?id=11656973

maurolacy: I suggest we cap it at two / three bytes, and error otherwise.

packages/cw0/src/parse_reply.rs
maurolacy: Will do.

PR #522, opened 2021-11-02
Add base-helpers.ts for contracts

findolor:
Resolves InterWasm/DAO#26

Based on the discussion on https://github.com/CosmWasm/cw-nfts/pull/8

Added ```base-helpers.ts``` for duplicated code in contact helpers. This new base file includes wallet-related stuff and configs.

### DIFF
--- a/contracts/base-helper.ts
+++ b/contracts/base-helper.ts
@@ -1,0 +1,120 @@
+import axios from  "axios";
+import fs from "fs";
+import { SigningCosmWasmClient } from "@cosmjs/cosmwasm-stargate";
+import { GasPrice, calculateFee, StdFee } from "@cosmjs/stargate";
+import { DirectSecp256k1HdWallet, makeCosmoshubPath } from "@cosmjs/proto-signing";
+import { Slip10RawIndex } from "@cosmjs/crypto";
+import path from "path";
+
+/*
+ * This base-helper is here to use with different helper files in contracts
+ *
+ * Usage: npx @cosmjs/cli@^0.26 --init https://raw.githubusercontent.com/CosmWasm/cosmwasm-plus/master/contracts/base-helpers.ts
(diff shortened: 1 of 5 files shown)


findolor:
Resolves https://github.com/InterWasm/DAO/issues/26

contracts/base-helper.ts
```diff
@@ -0,0 +1,120 @@
...
+  readonly feeToken: string
+  readonly bech32prefix: string
+  readonly hdPath: readonly Slip10RawIndex[]
+  readonly faucetUrl?: string
+  readonly defaultKeyFile: string,
+  readonly fees: {
+    upload: StdFee,
+    init: StdFee,
+    exec: StdFee
+  }
```
webmaster128, line 35: It would make sense to just store gasPrice once and then gas limits for each operation. This makes it much easier to go from network to network because only gas price needs to be updated.

`calculateFee` can calculate a `StdFee` at any time from gas price and gas limit.

contracts/base-helper.ts
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * If you want to use this code inside an app, you will need several imports from https://github.com/CosmWasm/cosmjs
+*/
+
+interface Options {
+  readonly httpUrl: string
+  readonly networkId: string
+  readonly feeToken: string
+  readonly bech32prefix: string
+  readonly hdPath: readonly Slip10RawIndex[]
```
webmaster128, line 28: You can use the alias

```suggestion
  readonly hdPath: HdPath
```

if the existing type looks too scary

webmaster128 (commented):
👍 I did not check all the contract logic, but have minor improvements ideas

findolor:
I'll close this PR and add/update these files in separate PRs

PR #523, opened 2021-11-02
cw1-whitelist-ng: Slight messages parsing improvement

hashedone:
Removed collecting errors to vector which might be significant cost on wasm, while it is completely not needed. This should also improve efficiency of generating error message, as its format is now well known (instead of iteration based) - but it is less important as it would happen only on err-path. And message itself might improve.


ethanfrey (approved):
Looks cheaper and cleaner.

PR #524, opened 2021-11-03
Release 0.10.2

maurolacy:
Let's release this with the multitest fixes, and keep the renaming of `cw0` to `utils` (along with index key deserialization stuff)  for 0.11.0.


ethanfrey (approved):
Nice. Good stuff.

I can publish today.

PR #525, opened 2021-11-03
cw1-subkeys: Migration example

### LINKED ISSUE #510: Semver parsing / comparison
Useful feature for migration support.

Pretty much every contract uses cw2 to store the [version as defined by cargo](https://github.com/CosmWasm/cw-plus/blob/main/contracts/cw1-subkeys/src/contract.rs#L33):

`const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");`

This is a string however, and not the easiest to use in migrate. to say, "allow migration where `version.major == 0 && [8..-=10].contains(version.minor)`".

As a first step, is a utility to parse a valid semver string into:

```rust
pub Semver struct {
  pub major: u32;
  pub minor: u32;
  pub patch: u32;
  // if you think this needs even more structure here, feel free to add
  pub tag: Option<String>;
}
```

Also please add some basic constructors and implement Ord, so we can do something like:

```rust
if version < Semver::new(0, 8, 0) {
  return Err("too low");
} else if version >= Semver::new(0, 10, 1) {
  return Err("too high");
} else {
  // do migration
}
```

The types above are meant to give inspiration and you can change them if you have a better idea that fulfills the same purpose.

### LINKED ISSUE #511: Migrate example
Builds on #510 

I think we no longer have valid migrate implementations in cw-plus, which makes it hard to explain. We should at least provide a stub to serve as documentation and a guide.

My idea was to allow cw-plus to upgrade from, say, v0.8.0 or newer to the current version (but must be <= current version). This migrate should be a no-op, but will allow someone to swap out the code ids to use newer functionality.

Make use of the `Semver` type from #510 to check for version requirements.

This is mainly meant for example/documentation purposes, go heavy on the comments.

hashedone:
Closes #510
Closes #511

### DIFF
--- a/contracts/cw1-subkeys/src/error.rs
+++ b/contracts/cw1-subkeys/src/error.rs
@@ -42,6 +42,9 @@ pub enum ContractError {
 
     #[error("Allowance already expired while setting: {0}")]
     SettingExpiredAllowance(Expiration),
+
+    #[error("Semver parsing error: {0}")]
+    SemVer(String),
 }
 
 impl From<cw1_whitelist::ContractError> for ContractError {
@@ -50,5 +53,11 @@ impl From<cw1_whitelist::ContractError> for ContractError {
             cw1_whitelist::ContractError::Std(error) => ContractError::Std(error),
             cw1_whitelist::ContractError::Unauthorized {} => ContractError::Unauthorized {},
         }
+    }
+}
+
+impl From<semver::Error> for ContractError {
+    fn from(err: semver::Error) -> Self {
+        Self::SemVer(err.to_string())
     }
 }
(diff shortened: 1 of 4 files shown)


contracts/cw1-subkeys/src/error.rs
```diff
@@ -42,6 +42,9 @@ pub enum ContractError {
 
     #[error("Allowance already expired while setting: {0}")]
     SettingExpiredAllowance(Expiration),
+
+    #[error("Semver parsing error: {0}")]
+    SemVer(String),
```
hashedone, line 47: I think this might be moved to `StdError` if we decide this is common one.

maurolacy (approved):
LGTM.

ethanfrey (approved):
Nice example. Thanks

Issue #505, opened 2021-10-22
Benchmark "ng" vs. "classic" frameworks


maurolacy:
Once the "ng" framework is more or less stable, let's create benchmarks to compare contract instantiation, execution, and querying on both versions. By example using the cw1-whitelist contracts as baseline / reference.

Results must be pretty similar, but better to measure and track gas usage over time.

JakeHartnell:
Awesome gas profiling tool by @de-husk: https://github.com/de-husk/cosm-orc

PR #534, opened 2021-11-04
cw1-whitelist-ng: Benchmarking

hashedone:
Addresses #505 

This is not finished proper benchmark, but rather POC for verifying gas usage regression with ng. Created tests uses the optimized binaries which has to be generated to use them. To make it work I needed to create "fork" of cosmwasm-std which doesn't perform message parsing (so it is delayed to contract). It is not changing behaviour of anything in stdlib, it just adds possibility to bypass message deserialization. In final ng contracts it would not be needed, as entry points would be generated for the contracts from scratch. 

Unfortunately tests went bad for ng. Gas usages for cw1-whitelist are:

```
Instantiation gas usage: 5480850233
Execution gas usage: 10734450010
```

For cw1-whitelist-ng results are:

```
Instantiation gas usage: 5939100233
Execution gas usage: 13134000010
```

Instantiation regression is not-so-bad (~8%), considering it is one-time thing, and probably can be improved with some tweaking, but the regression of execution worries me a lot (over 20% if I am not mistaken). However I have strong suspicions what causes that and how to improve that, so it might be way better. However this shows, that before any further work on ng is done, performance should be addressed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,11 +4,11 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.16.0"
+version = "0.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3e61f2b7f93d2c7d2b08263acaa4a363b3e276806c68af6134c44f523bf1aacd"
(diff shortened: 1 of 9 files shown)


ueco-jb:
NG works continues on different branches anyway.

PR #540, opened 2021-11-09
Refactor indexes into files

### LINKED ISSUE #530: Refactor `UniqueIndex` and `MultiIndex` into their own files

maurolacy:
Closes #530. Based on the #461 branch which is still wip.


packages/storage-plus/src/lib.rs
```diff
@@ -8,25 +8,29 @@ mod item;
...
-pub use indexes::{
-    index_string, index_string_tuple, index_triple, index_tuple, Index, MultiIndex, UniqueIndex,
-};
+pub use indexes::{index_string, index_string_tuple, index_triple, index_tuple, Index};
 pub use item::Item;
 pub use keys::{I128Key, I16Key, I32Key, I64Key, I8Key};
 pub use keys::{Prefixer, PrimaryKey, U128Key, U16Key, U32Key, U64Key, U8Key};
 pub use map::Map;
+#[cfg(feature = "iterator")]
+pub use multi_index::MultiIndex;
```
ethanfrey, line 29: If we pull them out (which I agree with), how about a module for them all.

eg. indexes/mod.rs, indexes/unique.rs, indexes/multi.rs

And then we can keep the top-level export them same even, just handle differently internally.
(There are so many files in storage-plus I am not in favour of sub-modules when possible)

ethanfrey (approved):
Looks like a clean split.

I would use a module though (which can be a follow up PR if you want, just moving these files around)

packages/storage-plus/src/lib.rs
maurolacy: Like with `snapshot`, I agree, let's do it in a follow-up

maurolacy: Done in #548.

PR #541, opened 2021-11-09
Add `prefix_de` to SnapshotMap

maurolacy:
#461 follow-up. Adds `prefix_de` and friends to `SnapshotMap` (Addresses second part of https://github.com/CosmWasm/cw-plus/issues/461#issuecomment-961153138).

### DIFF
--- a/packages/storage-plus/src/snapshot/map.rs
+++ b/packages/storage-plus/src/snapshot/map.rs
@@ -534,4 +534,30 @@ mod tests {
         assert_eq!(1, all.len());
         assert_eq!(all, vec![("A".into(), 22),]);
     }
+
+    #[test]
+    #[cfg(feature = "iterator")]
+    fn sub_prefix_de_composite_key() {
+        use cosmwasm_std::Order;
+
+        let mut store = MockStorage::new();
+        init_data_composite_key(&EVERY_COMPOSITE_KEY, &mut store);
+
+        // Let's sub-prefix and iterate.
+        // This is similar to calling range() directly, but added here for completeness /
+        // sub_prefix_de type checks
+        let all: StdResult<Vec<_>> = EVERY_COMPOSITE_KEY
+            .sub_prefix_de(())
+            .range_de(&store, None, None, Order::Ascending)
+            .collect();
+        let all = all.unwrap();
+        assert_eq!(2, all.len());
+        assert_eq!(
+            all,
+            vec![
+                (("B".into(), "B".into()), 13),
+                (("C".into(), "A".into()), 22)
+            ]
+        );
+    }
 }
--- a/packages/storage-plus/src/snapshot/map.rs
+++ b/packages/storage-plus/src/snapshot/map.rs
@@ -479,4 +479,22 @@ mod tests {
             ]
         );
     }
+
+    #[test]
+    #[cfg(feature = "iterator")]
+    fn prefix_de_composite_key() {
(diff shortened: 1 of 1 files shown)


packages/storage-plus/src/snapshot/map.rs
```diff
@@ -191,6 +192,28 @@ where
...
+        max: Option<PrefixBound<'a, K::Prefix>>,
+        order: cosmwasm_std::Order,
+    ) -> Box<dyn Iterator<Item = StdResult<(K::Output, T)>> + 'c>
+    where
+        T: 'c,
+        'a: 'c,
+        K: 'c,
+        K::Output: 'static,
+    {
+        let mapped = namespaced_prefix_range(store, self.primary.namespace(), min, max, order)
```
ethanfrey, line 212: This looks same as others

packages/storage-plus/src/snapshot/map.rs
```diff
@@ -205,6 +228,28 @@ where
...
+        K::Output: 'static,
+    {
+        self.no_prefix_de().keys_de(store, min, max, order)
+    }
+
+    pub fn prefix_de(&self, p: K::Prefix) -> Prefix<K::Suffix, T> {
+        Prefix::new(self.primary.namespace(), &p.prefix())
+    }
+
+    pub fn sub_prefix_de(&self, p: K::SubPrefix) -> Prefix<K::SuperSuffix, T> {
```
ethanfrey, line 249: It is still a bit funny for me that the types here (connection between `p: K::SubPrefix` and `Prefix<K::SuperSuffix` seems to only be when writing this code and not enforced in the compile, but it is fine. Here we hide this chance to error from any user, and we are quite careful with this library code.

Mainly a thought for a future type-system

packages/storage-plus/src/snapshot/map.rs
```diff
@@ -215,8 +260,8 @@ mod tests {
     use super::*;
     use cosmwasm_std::testing::MockStorage;
 
-    type TestMap = SnapshotMap<'static, &'static [u8], u64>;
-    type TestMapCompositeKey = SnapshotMap<'static, (&'static [u8], &'static [u8]), u64>;
+    type TestMap = SnapshotMap<'static, &'static str, u64>;
```
ethanfrey, line 263: :+1:

packages/storage-plus/src/snapshot/map.rs
```diff
@@ -492,9 +511,27 @@ mod tests {
...
-                ((b"C".to_vec(), b"A".to_vec()), 22)
+                (("B".into(), "B".into()), 13),
+                (("C".into(), "A".into()), 22)
             ]
         );
     }
+
+    #[test]
+    #[cfg(feature = "iterator")]
+    fn prefix_de_composite_key() {
```
ethanfrey, line 522: good test

ethanfrey (approved):
lgtm

packages/storage-plus/src/snapshot/map.rs
maurolacy: Yes, we can easily mess up here. A type system with something like composite types, whose parts or elements could be referred to independently would be nice, yes.

packages/storage-plus/src/snapshot/map.rs
maurolacy: Added a test / example anyway.

PR #542, opened 2021-11-09
Add `prefix_de` to `IndexedMap`

maurolacy:
#461 follow-up. Adds `sub_`/`prefix_de` to `IndexedMap` for completeness (Addresses third part of https://github.com/CosmWasm/cw-plus/issues/461#issuecomment-961153138).

### DIFF
--- a/packages/storage-plus/src/indexed_map.rs
+++ b/packages/storage-plus/src/indexed_map.rs
@@ -1041,6 +1041,213 @@ mod test {
 
     #[test]
     #[cfg(feature = "iterator")]
+    fn prefix_de_simple_string_key() {
+        let mut store = MockStorage::new();
+        let map = build_map();
+
+        // save data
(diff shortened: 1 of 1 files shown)


packages/storage-plus/src/indexed_map.rs
```diff
@@ -196,6 +196,22 @@ where
     }
 }
 
+#[cfg(feature = "iterator")]
+impl<'a, K, T, I> IndexedMap<'a, K, T, I>
```
ethanfrey, line 200: I don't see any tests for these.

ethanfrey (commented):
Code looks fine, but can we add a simple test case please?

packages/storage-plus/src/indexed_map.rs
maurolacy: There are no tests because they are the ranges for the base maps. But I will add them, it's more robust and maintainable.

PR #543, opened 2021-11-10
Add `range_de` to `IndexedSnapshotMap`

maurolacy:
#461 follow-up. Adds `keys_de ` / `range_de` to `IndexedSnapshotMap`. Also adds `sub_`/`prefix_de` for completeness.
(Addresses fourth part of https://github.com/CosmWasm/cw-plus/issues/461#issuecomment-961153138).


ethanfrey (commented):
As in #542 please add a simple test case as a sanity check.

Then happy to review both in more detail

PR #503, opened 2021-10-21
Deprecate IntKey

ueco-jb:
closes https://github.com/CosmWasm/cw-plus/issues/472

PR #547, opened 2021-11-15
Deprecate IntKey

### LINKED ISSUE #472: Deprecate `IntKey`
We can implement `PrimaryKey` over integer types easily using macros.

Like in https://github.com/CosmWasm/cw-plus/blob/d536358fb9bc45a43a9e055ed05f59f0998ba03a/packages/storage-plus/src/de.rs#L99-L111

and then https://github.com/CosmWasm/cw-plus/blob/d536358fb9bc45a43a9e055ed05f59f0998ba03a/packages/storage-plus/src/de.rs#L113

That means, we can remove the need for `IntKey`, and use the integer values directly. So, better usability, clarity, and user experience in general.

ueco-jb:
closes #472 
Successor of https://github.com/CosmWasm/cw-plus/pull/503.

Idea is to add very small wrapper, which takes either reference or small value (of limited number of bytes).
References should be used for any slice types, while refs for `int` types.
It's basically same approach as with `IntKey`, but instead of wrapping at beginning we do it at the end (in the result).

### DIFF
--- a/packages/storage-plus/src/de.rs
+++ b/packages/storage-plus/src/de.rs
@@ -98,7 +98,7 @@ impl KeyDeserialize for &Addr {
 
 macro_rules! integer_de {
     (for $($t:ty),+) => {
-        $(impl KeyDeserialize for IntKey<$t> {
+        $(impl KeyDeserialize for $t {
             type Output = $t;
 
             #[inline(always)]
@@ -111,6 +111,22 @@ macro_rules! integer_de {
 }
 
 integer_de!(for i8, u8, i16, u16, i32, u32, i64, u64, i128, u128);
+
+macro_rules! intkey_de {
+    (for $($t:ty),+) => {
+        $(impl KeyDeserialize for IntKey<$t> {
+            type Output = $t;
+
+            #[inline(always)]
+            fn from_vec(value: Vec<u8>) -> StdResult<Self::Output> {
+                Ok(<$t>::from_be_bytes(value.as_slice().try_into()
+                    .map_err(|err: TryFromSliceError| StdError::generic_err(err.to_string()))?))
+            }
+        })*
+    }
+}
+
+intkey_de!(for i8, u8, i16, u16, i32, u32, i64, u64, i128, u128);
 
 impl KeyDeserialize for TimestampKey {
     type Output = u64;
--- a/packages/storage-plus/src/helpers.rs
+++ b/packages/storage-plus/src/helpers.rs
@@ -7,6 +7,8 @@
 use serde::de::DeserializeOwned;
 use std::any::type_name;
 
+use crate::keys::Key;
+
 use cosmwasm_std::{
     from_slice, to_vec, Addr, Binary, ContractResult, Empty, QuerierWrapper, QueryRequest,
     StdError, StdResult, SystemResult, WasmQuery,
@@ -54,15 +56,15 @@ pub(crate) fn namespaces_with_key(namespaces: &[&[u8]], key: &[u8]) -> Vec<u8> {
 /// there are multiple sets we do not want to combine just to call this
 pub(crate) fn nested_namespaces_with_key(
     top_names: &[&[u8]],
-    sub_names: &[&[u8]],
+    sub_names: &[Key],
     key: &[u8],
 ) -> Vec<u8> {
     let mut size = key.len();
     for &namespace in top_names {
         size += namespace.len() + 2;
     }
-    for &namespace in sub_names {
-        size += namespace.len() + 2;
+    for namespace in sub_names {
+        size += namespace.as_ref().len() + 2;
     }
 
     let mut out = Vec::with_capacity(size);
@@ -70,9 +72,9 @@ pub(crate) fn nested_namespaces_with_key(
         out.extend_from_slice(&encode_length(namespace));
         out.extend_from_slice(namespace);
     }
-    for &namespace in sub_names {
-        out.extend_from_slice(&encode_length(namespace));
-        out.extend_from_slice(namespace);
+    for namespace in sub_names {
+        out.extend_from_slice(&encode_length(namespace.as_ref()));
+        out.extend_from_slice(namespace.as_ref());
     }
     out.extend_from_slice(key);
     out
--- a/packages/storage-plus/src/helpers.rs
+++ b/packages/storage-plus/src/helpers.rs
@@ -22,7 +22,7 @@ pub(crate) fn may_deserialize<T: DeserializeOwned>(
     value: &Option<Vec<u8>>,
 ) -> StdResult<Option<T>> {
     match value {
-        Some(vec) => Ok(Some(from_slice(&vec)?)),
+        Some(vec) => Ok(Some(from_slice(vec)?)),
         None => Ok(None),
     }
 }
@@ -30,7 +30,7 @@ pub(crate) fn may_deserialize<T: DeserializeOwned>(
 /// must_deserialize parses json bytes from storage (Option), returning NotFound error if no data present
 pub(crate) fn must_deserialize<T: DeserializeOwned>(value: &Option<Vec<u8>>) -> StdResult<T> {
     match value {
-        Some(vec) => from_slice(&vec),
+        Some(vec) => from_slice(vec),
         None => Err(StdError::not_found(type_name::<T>())),
     }
 }
--- a/packages/storage-plus/src/keys.rs
+++ b/packages/storage-plus/src/keys.rs
@@ -202,7 +202,7 @@ impl<'a> PrimaryKey<'a> for Vec<u8> {
     type SuperSuffix = Self;
 
     fn key(&self) -> Vec<Key> {
-        vec![Key::Ref(&self)]
+        vec![Key::Ref(self)]
     }
 }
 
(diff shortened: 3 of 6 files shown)


packages/storage-plus/src/keys.rs
```diff
@@ -450,7 +528,7 @@ mod test {
     #[test]
     fn proper_prefixes() {
         let simple: &str = "hello";
-        assert_eq!(simple.prefix(), vec![b"hello"]);
+        assert_eq!(simple.prefix()[0].as_ref(), b"hello");
```
maurolacy, line 531: Let's add a `assert_eq!(simple.prefix().len(), 1)` here.

packages/storage-plus/src/map.rs
```diff
@@ -43,7 +43,14 @@ where
...
 
     pub fn key(&self, k: K) -> Path<T> {
-        Path::new(self.namespace, &k.key())
+        Path::new(
+            self.namespace,
+            &k.key()
+                .iter()
+                .map(|e| e.as_ref())
+                .collect::<Vec<_>>()
+                .as_slice(),
```
maurolacy, line 52: This can be removed I guess.

packages/storage-plus/src/prefix.rs
```diff
@@ -128,7 +129,7 @@ where
 
     pub fn with_deserialization_function(
         top_name: &[u8],
-        sub_names: &[&[u8]],
+        sub_names: &[Key],
```
maurolacy, line 132: Nice that (for now at least) this key hasn't leaked to `Map` and friends.

packages/storage-plus/src/map.rs
ueco-jb: It couldn't, but I used simpler `.map`:
```rust
&k.key().iter().map(Key::as_ref).collect::<Vec<_>>(),
```

packages/storage-plus/src/de.rs
```diff
@@ -112,6 +112,22 @@ macro_rules! integer_de {
 
 integer_de!(for i8, u8, i16, u16, i32, u32, i64, u64, i128, u128);
 
+macro_rules! intkey_de {
```
ethanfrey, line 115: Ah, cool.. just started commenting how we need both. Then see how we do.

Nice!

packages/storage-plus/src/helpers.rs
```diff
@@ -20,15 +22,15 @@ pub(crate) fn may_deserialize<T: DeserializeOwned>(
     value: &Option<Vec<u8>>,
 ) -> StdResult<Option<T>> {
     match value {
-        Some(vec) => Ok(Some(from_slice(&vec)?)),
+        Some(vec) => Ok(Some(from_slice(vec)?)),
```
ethanfrey, line 25: I think `&` used to be required by the compiler.
But now it is smarter?

Any idea when this changed?

packages/storage-plus/src/keys.rs
```diff
@@ -5,6 +5,33 @@ use crate::de::KeyDeserialize;
 use crate::helpers::namespaces_with_key;
 use crate::Endian;
 
+#[derive(Debug)]
+pub enum Key<'a> {
```
ethanfrey, line 9: Okay.... seems simpler than associated types.

Ah, `[u8; N]` is on the stack and easily passed rather than `Vec<u8>` which is on the heap... and we only have to support up to 8 bytes. Nice idea

packages/storage-plus/src/keys.rs
```diff
@@ -5,6 +5,33 @@ use crate::de::KeyDeserialize;
...
+    Ref(&'a [u8]),
+    Val8([u8; 1]),
+    Val16([u8; 2]),
+    Val32([u8; 4]),
+    Val64([u8; 8]),
```
ethanfrey, line 14: Actually, how about adding `Val128([u8; 16])` I am sure that sooner or later someone will want u128 as a key

packages/storage-plus/src/keys.rs
```diff
@@ -225,18 +255,47 @@ impl<'a> PrimaryKey<'a> for Addr {
...
+            type SuperSuffix = Self;
+
+            fn key(&self) -> Vec<Key> {
+                vec![Key::$v(self.to_be_bytes())]
+            }
+        })*
+    }
+}
+
+integer_key!(for i8, Val8, u8, Val8, i16, Val16, u16, Val16, i32, Val32, u32, Val32, i64, Val64, u64, Val64);
```
ethanfrey, line 285: Huh? I am a bit lost here with the macros, wouldn't it be:

`integer_key!(i8, Val8);`
`integer_key!(u8, Val8);`
`integer_key!(i16, Val16);`
`integer_key!(u16, Val16);`
...

Or that `for i8` does something magic?

packages/storage-plus/src/keys.rs
```diff
@@ -362,15 +421,67 @@ mod test {
         let k: U64Key = 134u64.into();
         let path = k.key();
         assert_eq!(1, path.len());
-        assert_eq!(134u64.to_be_bytes().to_vec(), path[0].to_vec());
+        assert_eq!(134u64.to_be_bytes(), path[0].as_ref());
```
ethanfrey, line 424: Huh?

Isn't `134u64.to_be_bytes()` type `[u8; 8]`
while `path[0].as_ref()` type `&[u8]`?

(pointing to the equal underlying bytes however)

packages/storage-plus/src/keys.rs
```diff
@@ -225,18 +255,47 @@ impl<'a> PrimaryKey<'a> for Addr {
...
+macro_rules! integer_key {
+    (for $($t:ty, $v:tt),+) => {
+        $(impl<'a> PrimaryKey<'a> for $t {
+            type Prefix = ();
+            type SubPrefix = ();
+            type Suffix = Self;
+            type SuperSuffix = Self;
+
+            fn key(&self) -> Vec<Key> {
+                vec![Key::$v(self.to_be_bytes())]
```
ethanfrey, line 279: makes sense

packages/storage-plus/src/keys.rs
```diff
@@ -362,15 +421,67 @@ mod test {
...
+        assert_eq!(4242u32.to_be_bytes(), path[0].as_ref());
+
+        let k: i32 = 4242i32;
+        let path = k.key();
+        assert_eq!(1, path.len());
+        assert_eq!(4242i32.to_be_bytes(), path[0].as_ref());
+    }
+
+    #[test]
+    fn naked_64key_works() {
```
ethanfrey, line 475: Good to cover all these cases...

I know if is just cut-and-paste, but maybe cover all the prefix cases as well?
(Maybe overkill as I can visually inspect they are the same, but test cases ensures no one else breaks this in 8 months in some refactor)

packages/storage-plus/src/keys.rs
```diff
@@ -380,7 +491,7 @@ mod test {
         let k: K = "hello";
         let path = k.key();
         assert_eq!(1, path.len());
-        assert_eq!("hello".as_bytes(), path[0]);
+        assert_eq!(b"hello", path[0].as_ref());
```
ethanfrey, line 494: Yes, the compiler has gotten smarter.

packages/storage-plus/src/keys.rs
```diff
@@ -474,4 +597,28 @@ mod test {
             vec![one.as_slice(), two.as_slice(), three.as_slice()]
         );
     }
+
+    #[test]
+    fn naked_proper_prefixes() {
+        let pair: (u32, &[u8]) = (12345, b"random");
```
ethanfrey, line 603: nice to see this behaves identically to U32Key (just by looking up)

packages/storage-plus/src/prefix.rs
ethanfrey: I agree very much. I would try hard to hide that from contract developers, even if we use it everywhere in internal APIs. (And they can see it if they try to implement their own PrimaryKey type, but then we self-select for experienced devs)

ethanfrey (approved):
Looks good to me.

Left a few comments, but most are my surprise with the changes of the compiler, to auto-reference on eg. `match a: &Option<X>` and handling equals between `[u8; 5]` and `&[u8]`.

Both nice and surprising ergonomic improvements. Happy for a link to the Rust upgrade that brought them.

packages/storage-plus/src/keys.rs
maurolacy: Didn't implement it for 128 bits on purpose. Enums are tagged unions. So, they have the size of their largest element (plus an extra for the tagging).

Adding 128 bits support would double the size of `Key`  to 16 bytes (plus the extra for tagging, which must end up in a 32-bits aligned boundary). So, 20 bytes instead of 12.

And, we are not using 128 int keys anywhere. We can easily add it, if you want. Or we can always add it later, when needed / requested.

packages/storage-plus/src/keys.rs
maurolacy: It's the `for` keyword yes. Sadly, with declarative macros we cannot take the type (`i8`), obtain the size (`8`), and build an identifier (`Val8`). That's why those are repeated. I've heard something like that can be done with procedural macros, though.

packages/storage-plus/src/keys.rs
ueco-jb: Added.

packages/storage-plus/src/helpers.rs
maurolacy: No idea. One way to check it could be compiling with different versions, and see in which one this started to work. The match is producing a ref, and that ref is being converted to slice without any help. Not sure this wasn't supported earlier...

PR #553, opened 2021-11-16
Align tests to mock_dependencies changes

hashedone:
The `mock_dependencies` function was changed in breaking way, but semver-compatible version was set after the change, which made most test not compile after removing `Cargo.lock` (or doing some change which would make update it). This aligns tests to work with those changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.16.0"
+version = "0.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3e61f2b7f93d2c7d2b08263acaa4a363b3e276806c68af6134c44f523bf1aacd"
+checksum = "b9ecd88a8c8378ca913a680cd98f0f13ac67383d35993f86c90a70e3f137816b"
 dependencies = [
  "gimli",
 ]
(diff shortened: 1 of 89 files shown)


ueco-jb (approved):
Not all heroes wear capes.

Also, you forgot to run schema.

hashedone:
Ah, forgot schema. Will come back to this little bit later and then merge with proper schemas.

webmaster128 (approved):
LGTM.

I recommend never removing Cargo.lock completely to avoid dealing with all kind of unrealated changes at once (here schemars upgrade). You can delete the blocks you care about and let cargo repait the file.

hashedone:
@webmaster128 I absolutely disagree. If semver is properly used there would never be this kind of a problem. Therefore I would say - you should always remove your `Cargo.lock` before your final commit to make sure, you didn't mess up the world around.

The official recommendation is to keep `Cargo.lock` committed for binaries to avoid unknown bugs in non-breaking-changes of your binary dependency assuming, that binary might have additional manual integration done besides CI testing, and nothing should update the dependencies (even non-breaking) without such integration. But for library crates the recommendation is to never commit the `Cargo.lock`, as library should build and work with any version of dependencies, and the final dependencies should be determined by binary using it.

In this very case I had to update the failing dependency anyway so it won't make a difference.

webmaster128:
Feel free to disagree. But you are missing a bunch of practical aspects here:
- In semver anything can break at any time in 0.x and pre-releases. So you must not use any of those dependencies.
- It's not even clear what breakage means. Is a different schemars output breaking? Well, here it clearly causes a diff worth thinking about.
- Deleting the lockfile makes the diff almost impossible to read. Nobody will look at the changes anymore.
- You create merge conflicts.
- Not checking in lockfiles for libraries created non-deterministic CI builds, which is very annoying if you try to hunt down a problem.

webmaster128:
If the delete & repair flow bothers you (which works perfectly well for both cargo any yarn), you can use the upgrade commands of cargo as well.

PR #554, opened 2021-11-16
Implement WasmQuery::ContractInfo

ethanfrey:
This one for you @ueco-jb 

I'd be happy if you add a unit test


ueco-jb (approved):

packages/multi-test/src/wasm.rs
```diff
@@ -1005,6 +1031,44 @@ mod test {
         );
     }
 
+    #[test]
```
ethanfrey, line 1034: Thanks for adding this

ethanfrey (commented):
thanks

Issue #545, opened 2021-11-11
Issue running wasm compilation out-of-the-box


jonbae:
when running the docker command to compile
```
docker run --rm -v "$(pwd)":/code \
  --mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target \
  --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
  cosmwasm/workspace-optimizer:0.12.3
```
I get a error of 
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 20, kind: NotADirectory, message: "Not a directory" }', src/main.rs:58:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I have done `npm install ` in the root directory on all the necessary packages for base-helpers.ts to run. 

I have also run the command `RUSTFLAGS="-C link-arg=-s" cargo build --release --target=wasm32-unknown-unknown --locked` as suggested to debug but all of the contracts compile fine. 

currently stuck and would appreciate help. 

Rust: rustc 1.56.1
OS: windows subsystem for linux

wojtczal:
I assume you're seeing a log such as `Building "contracts/base-helpers.ts" ...` right before what you pasted as an error. This happens because in the `contracts` directory which rust is trying to compile there is a typescript file. This is a recent change made by @findolor - maybe he can shed more light on it ;)

findolor:
Thanks for bringing this up! You are right this is expecting every item to be a directory in ```contracts```.
Opened a PR to fix this one.

webmaster128:
Thank you for the report. This is a missing _is directory_ check in https://github.com/CosmWasm/rust-optimizer/blob/v0.12.3/build_workspace/src/main.rs#L28-L32

@maurolacy could you look into this?

ethanfrey:
I believe this has since been fixed.
Both in a more robust rust-optimiser as well as in removing the base_helpers.ts file.

Can this be closed?

webmaster128:
True, fixed in 0.12.4 with this function: https://github.com/CosmWasm/rust-optimizer/blob/v0.12.4/build_workspace/src/main.rs#L22-L31.

I think it can be closed once all references to rust-optimizer in this project are updated.

PR #556, opened 2021-11-17
Change contract file prefix to be cw* in Cargo.toml

findolor:
Resolves https://github.com/CosmWasm/cw-plus/issues/545

When trying to compile all the contracts, we get an error because of the new ```base-helpers.ts```. Since ```Cargo.toml``` has ```contracts/*``` as a workspace, it tries to get inside the helpers file. This is not possible because it's a file. 

Added **cw** prefix to go through only the contract folders.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-members = ["packages/*", "contracts/*"]
+members = ["packages/*", "contracts/cw*"]
 
 [profile.release.package.cw1-subkeys]
 codegen-units = 1


webmaster128:
How do you get the error? What do you run and what do you get?

If it is related to the workspace-optimizer, we better add a directory check.

webmaster128:
Sorry, missed the ticket.

webmaster128:
Thank you for your contribution! This is very much appreciated. But we have to solve this differently in order to avoid having the same problem in diferent projects (see #545). Closing here because cargo itself does not have an issue with the file.

PR #561, opened 2021-11-19
Update cw2 readme - contract_info key

korzewski:
Hi everyone!

Based on my raw queries on terra contracts, all `Item` keys are not length prefixed.
Only `Map` keys add this prefix.

I'm using `cw2 = "0.9.1"`


maurolacy (approved):
LGTM.

Issue #531, opened 2021-11-04
Improve `MultiIndex` pk deserialization


maurolacy:
~~Signal the primary key deserialization type of `MultiIndex` using a `PK` trait or similar, like in `UniqueIndex`~~. (done in #568)

In passing, ~~rename `K` to `IK` (Index Key) for clarity in both indexes~~. (done in #569)

If possible, make those `PK` traits automatic / inherited from the `IndexedMap` specification. That means, better encapsulation / coupling between `IndexedMap` and the `*Index` impls.

maurolacy:
Breaking change, as it would change the internal representation of the index key. For 0.11, or pehaps, 0.12.

ethanfrey:
Is this done? @maurolacy 

If so, let's close. If not, can you improve the description so we can follow up on it?

maurolacy:
This is "closed" by the `index_trait` procedural macro.

Implementing the third point in the description would require a significant refactor of `IndexedMap` and associated indexes.

So, I would say, let's close it as solved through a workaround.

Issue #460, opened 2021-10-01
Deprecate `range` to `range_raw`


maurolacy:
After #432, we'll have full range deserialization working for `Map`.

Let's deprecate `range` in favour of `range_de`, keeping `range` as `range_raw`.

Regarding `prefix` vs. `prefix_de`, let's (try to) keep `prefix_de` as `prefix`, and remove the current `prefix`. That is, no `prefix_raw` impl, which will be confusing. That implies making `range_raw` work with / over `prefix_de`.

This also includes migrating the code in contracts to take advantage of the new key deserialization when useful.

Also, as part of this, let's improve traits segregation over `Map` and `Prefix` range-related methods, so that trait bounds are enforced only when strictly needed.

ethanfrey:
I support this.
I would target 0.11 for this, and get the new features in 0.10, so people can start opt-ing in.

PR #568, opened 2021-12-06
Index keys consistency

### LINKED ISSUE #532: `UniqueIndex` / `MultiIndex` key consistency
Unify the key returned by `UniqueIndex` and `MultiIndex`.

Currently, `UniqueIndex` returns just the pk, whereas `MultiIndex` returns the remaining of the key, including the pk as last element. Either make both return just the pk, ~~or make `UniqueIndex` behave like `MultiIndex` (which is better for usability / options)~~. (opted for the first option, for simplicity).

There are a number of FIXMEs in the code where this needs to be changed.

maurolacy:
Closes #532.

This simplifies the key returned by `MultiIndex`, so that it is consistent with the `UniqueIndex` key.

In principle the other approach is also possible: making `UniqueIndex` return the same than `MultiIndex`; see https://github.com/CosmWasm/cw-plus/tree/532-index-keys-consistency.

But this approach is simpler: we just return a tuple with the associated primary key, and the value. Given that the index key is a combination of the data fields present in the value, this makes sense, both to avoid returning redundant data and for simplicity.

This PR also introduces a generic type for signaling the type for primary key deserialization, like in `UniqueIndex`. So, it also closes (the first part of) #531.

Overall, the end result is that both indices are now behaving more similarly. Both in their returned values and also in their specification.

### DIFF
--- a/packages/storage-plus/src/indexed_map.rs
+++ b/packages/storage-plus/src/indexed_map.rs
@@ -722,7 +722,7 @@ mod test {
             last_name: "".to_string(),
             age: 42,
         };
-        let pk1: &str = "5627";
+        let pk1 = "5627";
         map.save(&mut store, pk1, &data1).unwrap();
 
         let data2 = Data {
@@ -730,7 +730,7 @@ mod test {
             last_name: "Perez".to_string(),
             age: 13,
         };
-        let pk2: &str = "5628";
+        let pk2 = "5628";
         map.save(&mut store, pk2, &data2).unwrap();
 
         let data3 = Data {
@@ -738,7 +738,7 @@ mod test {
             last_name: "Young".to_string(),
             age: 24,
         };
-        let pk3: &str = "5629";
+        let pk3 = "5629";
         map.save(&mut store, pk3, &data3).unwrap();
 
         let data4 = Data {
@@ -746,7 +746,7 @@ mod test {
             last_name: "Bemberg".to_string(),
             age: 43,
         };
-        let pk4: &str = "5630";
+        let pk4 = "5630";
         map.save(&mut store, pk4, &data4).unwrap();
 
         let marias: Vec<_> = map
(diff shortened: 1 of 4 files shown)


packages/storage-plus/src/indexed_map.rs
```diff
@@ -722,31 +722,31 @@ mod test {
             last_name: "".to_string(),
             age: 42,
         };
-        let pk1: &[u8] = b"5627";
+        let pk1: &str = "5627";
```
ueco-jb, line 725: I think those type annotations are unnecessary here, `&str` is default type for string slice.

packages/storage-plus/src/indexes/multi.rs
```diff
@@ -24,14 +25,16 @@ use crate::{Bound, Index, Prefix, Prefixer, PrimaryKey};
...
 /// is always a n-tuple (n >= 2) and its last element must be the pk.
 /// The index function must therefore put the pk as last element, when generating the index.
-pub struct MultiIndex<'a, K, T> {
+/// The optional PK type defines the type of Primary Key deserialization.
+pub struct MultiIndex<'a, K, T, PK = ()> {
     index: fn(&T, Vec<u8>) -> K,
     idx_namespace: &'a [u8],
     idx_map: Map<'a, K, u32>,
     pk_namespace: &'a [u8],
+    _phantom: PhantomData<PK>,
```
ueco-jb, line 34: It needs to be initialized in constructor anyway, and `PhantomData` shows that it's Phantom anway - I'd argue it's "rusty" to keep name without unused mark `_`:
```suggestion
    phantom: PhantomData<PK>,
```

ueco-jb (approved):
lgtm, I like those changes very much.
Indexing's getting simpler and simpler with every such change.

maurolacy:
Merging this. Please take a look at the follow-up (#569). After that and #460, I think this would be ready for cutting 0.11.

PR #569, opened 2021-12-06
`MultiIndex` primary key spec removal

### LINKED ISSUE #533: Remove the primary key from the `MultiIndex` key specification
In `MultiIndex`, remove the primary key from the index key **specification** (i.e. when building / defining the `MultiIndex). That is, handle index key multiplicity internally, using the pk or another method.

~~This can be done by signalling the pk type with its own generic type (see #531), and then internally building an index key of type `(IK, PK)`, that is, a tuple of Index key and Primary key. Of course, the index function must return `(IK, PK)` accordingly~~. (done by collapsing the ik, concatenated with the pk (#569)).

~~Alternatively, if the above is not possible make the index function type-aware on the pk element. That is, instead of using a bytes vector, use a proper type for the pk element. So that it's easier to handle internally~~. (not a good idea, as the pk is already in raw format at the indexes level)

maurolacy:
Closes #533.

Builds on top of #568. Makes it simpler to handle `MultiIndex` by removing the pk from the index key specification.

Solved it by collapsing the index key along with an extra element (the raw pk). This makes it possible to keep the old format in the store, as the new (internally generated) index key would be exactly the same as the one corresponding to the previous key spec.

Overall, I'm pretty happy with this implementation. It has some tricks, like shifting the definition of what is a prefix and a sub-prefix. So that prefixes can still be specified over the **full** index key (which doesn't include the primary key in its spec anymore). Despite (or because of) that, I think the end result is both clear and ergonomic.

This PR also renames `K` to `IK` for clarity. So, it also closes (the second part of) #531.

TODO:

- ~~Update related documentation / comments~~. (done)

### DIFF
--- a/packages/storage-plus/README.md
+++ b/packages/storage-plus/README.md
@@ -202,15 +202,14 @@ A `Map` key can be anything that implements the `PrimaryKey` trait. There are a 
  - `impl<'a, T: Endian + Clone> PrimaryKey<'a> for IntKey<T>`
 
 That means that byte and string slices, byte vectors, and strings, can be conveniently used as keys.
-Moreover, some other types can be used as well, like addresses and addresses references, pairs and triples, and
+Moreover, some other types can be used as well, like addresses and address references, pairs and triples, and
 integer types.
 
-If the key represents and address, we suggest using `&Addr` for keys in storage, instead of `String` or string slices. This implies doing address validation
-through `addr_validate` on any address passed in via a message, to ensure it's a legitimate address, and not random text
-which will fail later.
-
-Thus, `pub fn addr_validate(&self, &str) -> Addr` in `deps.api` can be used for address validation, and the returned
-`Addr` can be conveniently used as key in a `Map` or similar structure.
+If the key represents and address, we suggest using `&Addr` for keys in storage, instead of `String` or string slices.
+This implies doing address validation through `addr_validate` on any address passed in via a message, to ensure it's a
+legitimate address, and not random text which will fail later.
+`pub fn addr_validate(&self, &str) -> Addr` in `deps.api` can be used for address validation, and the returned `Addr`
+can then be conveniently used as key in a `Map` or similar structure.
 
 ### Composite Keys
 
@@ -265,7 +264,7 @@ reusing the calculated path to this key.
 For simple keys, this is just a bit less typing and a bit less gas if you
 use the same key for many calls. However, for composite keys, like
 `(b"owner", b"spender")` it is **much** less typing. And highly recommended anywhere
-you will use the a composite key even twice:
+you will use a composite key even twice:
 
 ```rust
 #[derive(Serialize, Deserialize, PartialEq, Debug, Clone)]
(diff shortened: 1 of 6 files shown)


packages/storage-plus/README.md
```diff
@@ -202,15 +202,14 @@ A `Map` key can be anything that implements the `PrimaryKey` trait. There are a
...
+Moreover, some other types can be used as well, like addresses and address references, pairs and triples, and
 integer types.
 
-If the key represents and address, we suggest using `&Addr` for keys in storage, instead of `String` or string slices. This implies doing address validation
-through `addr_validate` on any address passed in via a message, to ensure it's a legitimate address, and not random text
-which will fail later.
-
-Thus, `pub fn addr_validate(&self, &str) -> Addr` in `deps.api` can be used for address validation, and the returned
-`Addr` can be conveniently used as key in a `Map` or similar structure.
+If the key represents and address, we suggest using `&Addr` for keys in storage, instead of `String` or string slices.
```
ueco-jb, line 208: > If the key represents and address, we suggest ...
Hmm, this part confuses me.
Do you mean "represents and address [something, thing to address]"?

packages/storage-plus/src/indexed_map.rs
```diff
@@ -295,8 +295,8 @@ mod test {
     }
 
     struct DataIndexes<'a> {
-        // Second arg is for storing pk
-        pub name: MultiIndex<'a, (String, String), Data, String>,
+        // Last args are for signaling pk deserialization
```
ueco-jb, line 298: Nitpicking, but for clarity - you mean type parameters? (in more places below as well)

packages/storage-plus/src/indexes/multi.rs
```diff
@@ -22,19 +22,17 @@ use std::marker::PhantomData;
...
 ///
-/// The MultiIndex definition must include a field for the pk. That is, the MultiIndex K value
-/// is always a n-tuple (n >= 2) and its last element must be the pk.
-/// The index function must therefore put the pk as last element, when generating the index.
-/// The optional PK type defines the type of Primary Key deserialization.
-pub struct MultiIndex<'a, K, T, PK = ()> {
-    index: fn(&T, Vec<u8>) -> K,
+/// The (optional) PK type defines the type of Primary Key deserialization.
+pub struct MultiIndex<'a, IK, T, PK = ()> {
+    index: fn(&T) -> IK,
```
ueco-jb, line 27: This K -> IK change is quite sudden. I understand that `K` stands for `Key`, but is `IK` an `IndexKey`? 
I think it would be good to have some clear reference to it... I doubt it will be very clear to any people diving into this code from outside.

ueco-jb (approved):
lgtm, mostly nitpicking comments regarding comments.

ethanfrey:
I don't have the time to dig into this, which it deserves.

If you and Jake agree, feel free to merge.
I feel this is getting to larger changes and should be wrapped into a 0.11 release along with the other pending stuff.

Meaning, go ahead and merge, but we commit to no more 0.10.x releases now. And start all the 0.11 cleanup stuff. (When we get to them). Let's get that utils rename in (finally)

packages/storage-plus/README.md
maurolacy: It's "_an_ address". Will fix it.

packages/storage-plus/src/indexes/multi.rs
maurolacy: It's a way to differentiate it from `PK`, and a generic `K` that could stand for any of them.

PR #576, opened 2021-12-13
`range` to `range raw`

### LINKED ISSUE #460: Deprecate `range` to `range_raw`

maurolacy:
Closes #460.

Lots of changes, but basically renaming `range` to `range_raw`, and `range_de` to `range`.

- [x] Deprecates `range` in favour of `range_de`, keeping `range` as `range_raw`.
- [x] Keeps `prefix_de` as `prefix`, and remove the currents `prefix`. That is, no `prefix_raw` impl, which will be confusing. That implies making `range_raw` work with / over `prefix_de`.
- [x] Migrates the code in contracts to take advantage of the new key deserialization when useful.
- [x] ~~Improve traits segregation over `Map` and `Prefix` range-related methods, so that trait bounds are enforced only when strictly needed~~. (out of scope for this PR)

Could have chosen to provide two prefix functions, `prefix` and `prefix_raw`, but opted instead to provide one `prefix` with two ranges over it, `range` and `range_raw`, for simplicity of use. The only drawback is that if you want to prefix, you need to provide key deserialization impls for the involved keys, even if / when using only `range_raw`. This is reasonable IMO, as we are providing many of those impls already.

Most interesting stuff is in the contracts, where we leverage the new key deserialization functionality for clarity and convenience.

### DIFF
--- a/contracts/cw1-subkeys/src/contract.rs
+++ b/contracts/cw1-subkeys/src/contract.rs
@@ -409,8 +409,8 @@ pub fn query_all_allowances(
     // we use raw addresses here....
     let start = start_after.map(Bound::exclusive);
 
-    let res: StdResult<Vec<AllowanceInfo>> = ALLOWANCES
-        .range_raw(deps.storage, start, None, Order::Ascending)
+    let allowances = ALLOWANCES
+        .range(deps.storage, start, None, Order::Ascending)
         .filter(|item| {
             if let Ok((_, allow)) = item {
                 !allow.expires.is_expired(&env.block)
@@ -420,16 +420,14 @@ pub fn query_all_allowances(
         })
         .take(limit)
         .map(|item| {
-            item.and_then(|(k, allow)| {
-                Ok(AllowanceInfo {
-                    spender: String::from_utf8(k)?,
-                    balance: allow.balance,
-                    expires: allow.expires,
-                })
+            item.map(|(addr, allow)| AllowanceInfo {
+                spender: addr.into(),
+                balance: allow.balance,
+                expires: allow.expires,
             })
         })
-        .collect();
-    Ok(AllAllowancesResponse { allowances: res? })
+        .collect::<StdResult<Vec<_>>>()?;
+    Ok(AllAllowancesResponse { allowances })
 }
 
 // return a list of all permissions here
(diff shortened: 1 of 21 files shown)


contracts/cw1-subkeys/src/contract.rs
```diff
@@ -420,16 +420,14 @@ pub fn query_all_allowances(
         })
         .take(limit)
         .map(|item| {
-            item.and_then(|(k, allow)| {
-                Ok(AllowanceInfo {
-                    spender: String::from_utf8(k)?,
-                    balance: allow.balance,
-                    expires: allow.expires,
-                })
+            item.map(|(addr, allow)| AllowanceInfo {
```
ethanfrey, line 423: Nice, range uses deserialisation by default

packages/storage-plus/src/indexes/multi.rs
```diff
@@ -143,67 +143,59 @@ where
...
+// short-cut for simple keys, rather than .prefix(()).range_raw(...)
 impl<'a, IK, T, PK> MultiIndex<'a, IK, T, PK>
 where
     T: Serialize + DeserializeOwned + Clone,
     IK: PrimaryKey<'a> + Prefixer<'a>,
 {
     // I would prefer not to copy code from Prefix, but no other way
     // with lifetimes (create Prefix inside function and return ref = no no)
-    pub fn range<'c>(
+    pub fn range_raw<'c>(
```
ethanfrey, line 198: Yay!

packages/storage-plus/src/indexes/multi.rs
```diff
@@ -256,21 +248,23 @@ where
     T: Serialize + DeserializeOwned + Clone,
     IK: PrimaryKey<'a> + Prefixer<'a>,
 {
-    pub fn prefix_de(&self, p: IK) -> Prefix<PK, T> {
-        Prefix::with_deserialization_function(
+    pub fn prefix(&self, p: IK) -> Prefix<PK, T> {
```
ethanfrey, line 251: yay

ethanfrey (commented):
I like the idea.
Have not done a thorough review

Issue #570, opened 2021-12-07
Remove `IntKey` with surrounding implementation


ueco-jb:
After 0.11.0 release, it'd be a good time to remove deprecated `IntKey`, since we already aren't using it anywhere cw-plus.

Once `IntKey` is deprecated, let's use that name for `int_key::CwIntKey` (After a release or two, to avoid extra confusion).

PR #582, opened 2021-12-16
Signed int keys order

### LINKED ISSUE #489: Incorrect I32Key Index Ordering
When using an I32Key as the primary key for an index, the ordering is done incorrectly. Instead of sorting signed integers as you would expect them to be sorted, all negative numbers are considered to be "greater" than all positive numbers.

IntKey new implementation: 
```
impl<T: Endian> IntKey<T> {
    pub fn new(val: T) -> Self {
        IntKey {
            wrapped: val.to_be_bytes().into(),
            data: PhantomData,
        }
    }
}
```

The new implementation converts the i32 number to big endian bytes which doesn't preserve the sorting order.

My workaround was to restrict my range query with a max of the most negative number:
```
    let zero_winnings_key = leaderboard()
        .idx
        .winnings
        .index_key((I32Key::from(0), b"".to_vec()));
    let most_negative_key = leaderboard()
        .idx
        .winnings
        .index_key((I32Key::from(-2147483648), b"".to_vec()));

    let res = leaderboard()
        .idx
        .winnings
        .range(
            deps.storage,
            Some(Bound::inclusive(zero_winnings_key)),
            Some(Bound::exclusive(most_negative_key)),
            Order::Descending,
        )
        .take(limit)
        .collect::<StdResult<Vec<_>>>()?;
```
This fits my use case because I don't need the negative values when querying, but is quite counterintuitive and took me a long time to figure out what was going on.

maurolacy:
Closes #489.

There are a number of different ways to do the signed int key modification (a number of different impls in the commits). Not sure what is the best one; opted for doing it without mutable slices, and without branching.

I guess the only way to know for sure would be by benchmarking them. Will create an issue for it, and we can do  in the future.

### DIFF
--- a/packages/storage-plus/src/int_key.rs
+++ b/packages/storage-plus/src/int_key.rs
@@ -17,10 +17,12 @@ macro_rules! cw_uint_keys {
         $(impl CwIntKey for $t {
             type Buf = [u8; mem::size_of::<$t>()];
 
+            #[inline]
             fn to_cw_bytes(&self) -> Self::Buf {
                 self.to_be_bytes()
             }
 
+            #[inline]
             fn from_cw_bytes(bytes: Self::Buf) -> Self {
                 Self::from_be_bytes(bytes)
             }
@@ -35,10 +37,12 @@ macro_rules! cw_int_keys {
         $(impl CwIntKey for $t {
             type Buf = [u8; mem::size_of::<$t>()];
 
+            #[inline]
             fn to_cw_bytes(&self) -> Self::Buf {
                 (*self as $ut ^ <$t>::MIN as $ut).to_be_bytes()
             }
 
+            #[inline]
             fn from_cw_bytes(bytes: Self::Buf) -> Self {
                 (Self::from_be_bytes(bytes) as $ut ^ <$t>::MIN as $ut) as _
             }
--- a/packages/storage-plus/src/int_key.rs
+++ b/packages/storage-plus/src/int_key.rs
@@ -30,19 +30,19 @@ macro_rules! cw_uint_keys {
 cw_uint_keys!(for u8, u16, u32, u64, u128);
 
 macro_rules! cw_int_keys {
-    (for $($t:ty),+) => {
+    (for $($t:ty, $ut:ty),+) => {
         $(impl CwIntKey for $t {
             type Buf = [u8; mem::size_of::<$t>()];
 
             fn to_cw_bytes(&self) -> Self::Buf {
-                ((*self as u128 ^ <$t>::MIN as u128) as $t).to_be_bytes()
+                (*self as $ut ^ <$t>::MIN as $ut).to_be_bytes()
             }
 
             fn from_cw_bytes(bytes: Self::Buf) -> Self {
-                (Self::from_be_bytes(bytes) as u128 ^ <$t>::MIN as u128) as _
+                (Self::from_be_bytes(bytes) as $ut ^ <$t>::MIN as $ut) as _
             }
         })*
     }
 }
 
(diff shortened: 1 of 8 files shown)


packages/storage-plus/src/int_key.rs
```diff
@@ -0,0 +1,130 @@
+use std::mem;
+
+/// Our int keys are simply the big-endian representation bytes for unsigned ints,
+/// but "sign-flipped" (xored msb) big-endian bytes for signed ints.
+///
+/// So that the representation of signed integers is in the right lexicographical order.
+// TODO: Rename to `IntKey` when deprecating current `IntKey`
```
ueco-jb, line 7: Please mention that in https://github.com/CosmWasm/cw-plus/issues/570.

packages/storage-plus/src/keys.rs
```diff
@@ -615,7 +626,7 @@ mod test {
         assert_eq!(pair.prefix(), vec![one.as_slice(), two.as_slice()]);
 
         let pair: (i8, &[u8]) = (123, b"random");
-        let one: Vec<u8> = vec![123];
+        let one: Vec<u8> = vec![123 + 128];
```
ueco-jb, line 629: This might look reasonable now, but for some time (and for other dev) this "magic" addition might... not add up (see what I did there?).
Either assign some const to `128` or add some comments maybe.

ueco-jb (commented):
Good to see that this idea actually worked just fine.

packages/storage-plus/src/keys.rs
maurolacy: You are right. Perhaps doing `vec![123 ^ 0x80];` is clearer?

packages/storage-plus/src/keys.rs
```diff
@@ -326,16 +327,26 @@ impl<'a, T: Endian> Prefixer<'a> for IntKey<T> {
...
+#[deprecated(note = "It is suggested to use `u16` as key type instead of the `U16Key` wrapper")]
 pub type U16Key = IntKey<u16>;
+#[deprecated(note = "It is suggested to use `u32` as key type instead of the `U32Key` wrapper")]
 pub type U32Key = IntKey<u32>;
+#[deprecated(note = "It is suggested to use `u64` as key type instead of the `U64Key` wrapper")]
 pub type U64Key = IntKey<u64>;
+#[deprecated(note = "Consider using 64-bit keys instead of the `U128Key` wrapper")]
 pub type U128Key = IntKey<u128>;
 
+#[deprecated(note = "It is suggested to use `i8` as key type instead of the `I8Key` wrapper")]
```
ethanfrey, line 341: nice

packages/storage-plus/src/int_key.rs
```diff
@@ -0,0 +1,130 @@
...
+mod test {
+    use super::*;
+
+    #[test]
+    fn x8_int_key_works() {
+        let k: u8 = 42u8;
+        assert_eq!(k.to_cw_bytes(), k.to_be_bytes());
+
+        let k: i8 = 42i8;
+        assert_eq!(k.to_cw_bytes(), (k as u8 ^ 0x80).to_be_bytes());
```
ethanfrey, line 65: again, I'd like the expected results to be explicit

ethanfrey (commented):
looks good.
I would clean up the test cases to be more explicit rather than just contain the same math being tested

packages/storage-plus/src/int_key.rs
maurolacy: Will do.

ueco-jb (approved):
lgtm
Personally I'd replace couple big numbers with constants for better transparency, but overall it's fine.

Issue #495, opened 2021-10-18
Implement attribute macro for trait interface generating boilerplate specific for cw1-whitelist


hashedone:
Part of #493, follow up of #494 

* [x] Generate  `ExecMsg` from contract trait
* [x] Generate `QueryMsg` from contract trait
* [ ] Generate dispatch function for dispatching exec msg
* [ ] Generate dispatch function for dispatching query msg
* [ ] If entry points turns out being tricky - generate them, or provide tool for making them clear

egasimus:
Yo, these past few months we've been up to [more or less the same thing](https://github.com/hackbg/fadroma/tree/21.12/lib/derive-contract). (Contract traits too - though ours are in a private repo that we'll make public once it passes audits etc.) Good to see the design of the platform has been guiding everyone in the same direction :)

We've implemented ours on top of Secret Network's fork of CW 0.10, but word has it 0.16 is coming to SN soon. Wanna share the workload?

hashedone:
@egasimus First of all I am more than happy that someone besides me figured out the same idea. You are the second one who pointed me something similar after looking for my proposal (the other one was not CW-specific), which basically means there is a need for such solution.

I actually already started my implementation of contract generation. I have no problem with sharing workload, but we would need to decide what exactly shape are we modeling. What I lack in your approach is... documentation, so it is hard for me to what exact contract are you expecting to generate, and how do you want to handle next things. We started with creating "generated" version of contract created by hand (and actually even before it - I sketched a brief proposal about that: https://github.com/CosmWasm/cw-plus/issues/391). Possibly you have something similar, but you didn't link it and I cannot find it quickly - if there is something then please link so maybe we can influence this.

But if we are able to find some common sense of where are we going with that I think that sharing the workload would be very nice.

One think you should know about our solution is that for now it seems that it seems to be more gas costly than traditional contracts, at least for trivial contracts (it is hard to tell for now if its overhead is constant and would diminish for real life contracts or not). However we don't have a way to profile them correctly and I personally cannot figure out what is actually slowing down so much (see #567). If it is not a problem for you then it is ok, but if you have some idea how to improve the gas usage then it can be very helpful.

If you want we could probably setup some call or something to talk about this.

uint:
This is part of [Sylvia](https://github.com/CosmWasm/sylvia/tree/main/sylvia) now. Closing!

PR #584, opened 2021-12-16
Message types generation for cw-derive

hashedone:
Adresses #495 

I was really not sure how to handle working on this, so I decided that instead of merging it to main, I created the `cw-derive` branch out of main, and for now I would develop it here (at least until it would be remotely usable).

Here is first big part of code generation. It took me much time fighting attribute macro related issues, but I think now it generates ok structures.

In terms of `Ctx` - as most of you wanted rather types like separated `Deps`, `MessageInfo`, `Env` - I have it in mind, and it would be possible this way. This PR is only related to generating messages, and it is kind of important to make it easier that those variables are single one - to uniformly ignoring two first arguments of method (self, and context). In later steps I would make sure, that it is possible to put `(DepsMut, Env, MessageInfo)` (for execution - other messages might require another tuple) here conveniently.


packages/cw-derive/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+    let input = parse_macro_input!(item as ItemTrait);
+
+    let exec = build_msg(&attrs.exec, &input, parser::InterfaceMsgAttr::Exec);
+    let query = build_msg(&attrs.query, &input, parser::InterfaceMsgAttr::Query);
+
+    let input = StripInput.fold_item_trait(input);
+
+    let expanded = if let Some(module) = attrs.module {
+        quote! {
+            mod #module {
```
ethanfrey, line 93: `pub mod`

packages/cw-derive/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+                None
+            }
+        }
+        _ => None,
+    });
+
+    quote! {
+        #[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq, schemars::JsonSchema)]
+        #[serde(rename_all="snake_case")]
+        enum #name {
```
ethanfrey, line 137: `pub enum`

packages/cw-derive/examples/basic.rs
```diff
@@ -0,0 +1,30 @@
...
+use cosmwasm_std::Response;
+
+struct Ctx;
+struct Error;
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq, schemars::JsonSchema)]
+struct Member;
+
+#[cw_derive::interface(module=msg, exec=Execute, query=Query)]
+trait Cw4 {
```
ethanfrey, line 11: `pub trait`

packages/cw-derive/examples/basic.rs
```diff
@@ -0,0 +1,30 @@
+#![allow(dead_code)]
+use cosmwasm_std::Response;
+
+struct Ctx;
```
ethanfrey, line 4: In a real contract, Ctx would have a defined structure, and Error would be some custom `this_error` type, right?

ethanfrey (commented):
Nice start on the codegen macros.
A few comments on the code that is generated.
I cannot deeply review the workings of the macros, but what I see looks good.

packages/cw-derive/examples/basic.rs
hashedone: `Ctx` would be a tuple of `(DepsMut, Env, MessageInfo)` (or `Deps` for queries) or something which can be made from it with `.into()`. Here it is only for generation purposes.

Error would be any error type returned by contract, in particular it can be one generated with `this_error` (in our codebase it will be, but it is not enforced by codegen).

packages/cw-derive/examples/basic.rs
hashedone: True. Doesn't matter for example, but true.

packages/cw-derive/examples/basic.rs
```diff
@@ -0,0 +1,30 @@
...
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq, schemars::JsonSchema)]
+struct Member;
+
+#[cw_derive::interface(module=msg, exec=Execute, query=Query)]
+trait Cw4 {
+    #[msg(exec)]
+    fn update_admin(&self, ctx: Ctx, admin: Option<String>) -> Result<Response, Error>;
+
+    #[msg(exec)]
```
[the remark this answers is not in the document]
hashedone: Yes, `cw_derive::interface` is basically scanning for those `msg` attributes to generate inferface.

There is a reason not to hardcode the names - becase `Cw1Exec` or `Cw1ExecMsg` is way better than `ExecMsg` which is default. It could be reasonable to have default as `TraitNameExecMsg`/`TraitNameQueryMsg`, but it also may not always work. Making it configurable with some default is now way easier now than complicating generation later.

packages/cw-derive/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+/// # #[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq, schemars::JsonSchema)]
+/// # struct Member;
+///
+/// #[cw_derive::interface(module=msg, exec=Execute, query=Query)]
+/// trait Cw4 {
+///     #[msg(exec)]
+///     fn update_admin(&self, ctx: Ctx, admin: Option<String>) -> Result<Response, Error>;
+///
+///     #[msg(exec)]
+///     fn update_members(&self, ctx: Ctx, remove: Vec<String>, add: Vec<Member>)
```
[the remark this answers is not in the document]
hashedone: I don't understand - if it is equivalent, then how they get different code generated? Schema generator claims it generates same schemas for equivalend serde serializers... The problem is that extracting this to structures may lead to names collisions easly, especially for more complicated structures. It is partially solved by keeping those in separate module (opt-in), but what if both query and execute would have same variant? Two same structures would be generated... You can also try to add suffixes on such case (like numbers) but will it still be easier to use? It is possible, but not sure how much would it improve at this level.

packages/cw-derive/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+///         },
+///         AddHook { addr: String },
+///         RemoveHook { addr: String },
+///     }
+/// }
+/// ```
+///
+/// ## Parameters
+///
+/// `interface` attribute takes optional parameters:
```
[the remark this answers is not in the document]
hashedone: Mostly because of naming collisions. You want to implement many traits on single contract. Probably you want them named differently. And in general - you have more flexibility this way. You can always omit all of those, as they have their defaults.

hashedone:
Ok, I will merge it to cw-derive as it is opened for kind of long, and still be rereviewed. It would let me to make PR for another part. You reviewed the most important part - the usage of it. I will leave name customization for now as I don't think it hurts. Possibly we can change defaults better so name conflicts will not be an issue.

packages/cw-derive/examples/basic.rs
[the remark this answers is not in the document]
ethanfrey: Fair enough.
As I guessed, you were 2 steps ahead of me.

packages/cw-derive/src/lib.rs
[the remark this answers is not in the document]
ethanfrey: NP. I don't see the name collision issue, but no need to do it.

The generated types have the same shape, but those intermediate type definitions (one per variant) can help in other languages. We can also work on the codegen side in other languages here.

That said, we can always change the auto-gen later and it would be functionally equivalent

PR #585, opened 2021-12-17
Make `update_changelog.sh` use the latest version tag by default

maurolacy:
Better `update_changelog` default behaviour to avoid rewriting / overwriting changelog entries.


ueco-jb (approved):
lgtm

PR #586, opened 2021-12-21
Implement range queries for claim controller

orkunkl:
Short PR, implements range queries for claims


packages/controllers/src/claim.rs
```diff
@@ -582,4 +604,50 @@ mod test {
...
+                Addr::unchecked("addr2"),
+                vec![Claim {
+                    amount: (TEST_AMOUNT + 100).into(),
+                    release_at: TEST_EXPIRATION,
+                }],
+            ),
+        ];
+        assert_eq!(cls, expected)
+
+        // no need test intensively, storage range works
```
ueco-jb, line 651: This is range query where you don't test range at all.
There are 3 Option parameters, this is 1 of potentially 9 possible cases (or more, if we'd like to cover order as well).
Why don't you create additional test cases at least for pagination?

I get that probably similar implementation is tested elsewhere, but this is open sourced - IMO we should cover more.

packages/controllers/src/claim.rs
```diff
@@ -90,6 +91,28 @@ impl<'a> Claims<'a> {
         let claims = self.0.may_load(deps.storage, address)?.unwrap_or_default();
         Ok(ClaimsResponse { claims })
     }
+
+    pub fn range_claims(
+        &self,
+        deps: Deps,
+        min: Option<Bound>,
```
ethanfrey, line 98: This is different than most list queries. I assume this would be used by such a query? Or is this for "internal use only"?

I would either:

* use `start_after` and `limit` (assigning default/max to limit if not set) like all `list_*` queries.
* or just expose `self.0.range()` with the identical arguments, not trying to add limits, using `&dyn Storage` not `Deps`, return Iterator, etc

Those are the 2 existing APIs people use. I see no reason to invent a 3rd API for the same approach (the other 2 are internal vs external, just pick one)

Actually, I think the simplest implementation for the 2nd approach is to expose the whole map read-only like, by implementing Deref here (so I can do `CLAIMS.range(...)` and it will auto-call the Map, but it doesn't let me modify the Map (only get a &Map)

ethanfrey (commented):
I understand the motivation.
I propose 2 different approaches to do so.

ueco-jb:
@maurolacy do you think this could be potentially implemented through some new PR?
Do we have need for it?

Issue #551, opened 2021-11-16
Check (and possibly fix) threshold and voting power implementation in `cw3-fixed-multisig`


ueco-jb:
Follow up of pretty much identical issue for `cw3-flex-multisig` - https://github.com/CosmWasm/cw-plus/issues/527.
Minimum threshold required should be `>= 0.5` and voting power must be `>= 1` to allow member to vote.

PR #588, opened 2021-12-21
Fix min threshold and vote power bugs in cw3-fixed-multisig

ueco-jb:
closes https://github.com/CosmWasm/cw-plus/issues/551

Couple hundred lines are mostly move some stuff like `Proposal` and `Ballot` structs from `flex` to `fixed-multisig` contracts, following [Mauro's comment](https://github.com/CosmWasm/cw-plus/pull/588#discussion_r773763853).

~~EDIT: Actually scratch that... It appears that I can't just import common implementation in `cw3-flex-multisig` from `cw3-fixed-multisig`, because `wasm-build` complains during flex compilation about duplicate of symbols for `instantiate`, `execute` and `query` functions.
So in the end, I just copied whole implementation into `fixed` contract and removed tests from `flex` - bit controversial, but they are copied into `fixed`. But I get that duplicated tests are still tests, so I can bring them back anyway.~~
<yoda's voice>Use the `library` feature you must.</yoda's voice>


contracts/cw3-fixed-multisig/src/error.rs
```diff
@@ -1,11 +1,15 @@
...
+use cw3_flex_multisig::error::ContractError as Cw3FlexMultisigError;
 use thiserror::Error;
 
 #[derive(Error, Debug, PartialEq)]
 pub enum ContractError {
     #[error("{0}")]
     Std(#[from] StdError),
 
+    #[error("{0}")]
+    FlexMultisig(#[from] Cw3FlexMultisigError),
```
ueco-jb, line 11: Extracting whole `Threshold` implementation in https://github.com/CosmWasm/cw-plus/pull/590 will make this error clearer.

contracts/cw3-fixed-multisig/Cargo.toml
```diff
@@ -21,6 +21,7 @@ library = []
 utils = { path = "../../packages/utils", version = "0.10.3" }
 cw2 = { path = "../../packages/cw2", version = "0.10.3" }
 cw3 = { path = "../../packages/cw3", version = "0.10.3" }
+cw3-flex-multisig = { path = "../cw3-flex-multisig", version = "0.10.3" }
```
maurolacy, line 24: Not a big deal, but it would be good to do it the other way around, i.e. cw3-flex importing stuff from cw3-fixed.

contracts/cw3-fixed-multisig/src/msg.rs
```diff
@@ -3,12 +3,13 @@ use serde::{Deserialize, Serialize};
...
 use cosmwasm_std::{CosmosMsg, Empty};
 use cw3::Vote;
+use cw3_flex_multisig::msg::Threshold;
 use utils::{Duration, Expiration};
 
 #[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
 pub struct InstantiateMsg {
     pub voters: Vec<Voter>,
-    pub required_weight: u64,
+    pub threshold: Threshold,
```
maurolacy, line 12: We should list / comment on these breaking changes in `MIGRATING.md` now.

contracts/cw3-fixed-multisig/Cargo.toml
ueco-jb: Actually this is because of threshold, so it won't be needed after https://github.com/CosmWasm/cw-plus/pull/590

ueco-jb: I moved common implementation of proposals, ballots etc. from flex to fixed multisig.
Now is how you pointed - flex imports those from fixed.

maurolacy:
You can use the `library` feature to avoid duplicate symbols.

contracts/cw3-fixed-multisig/src/state.rs
```diff
@@ -18,21 +22,26 @@ pub struct Config {
...
+    /// pass requirements
+    pub threshold: Threshold,
+    // the total weight when the proposal started (used to calculate percentages)
+    pub total_weight: u64,
+    // summary of existing votes
+    pub votes: Votes,
 }
 
 impl Proposal {
+    /// current_status is non-mutable and returns what the status should be.
```
maurolacy, line 38: 👍🏼 Thanks for doing this sync-up.

maurolacy (approved):
LGTM.

PR #590, opened 2021-12-21
Move Threshold and coexisting implementations into packages/utils

ueco-jb:
Since both fixed and flex multisig contracts should rely on same threshold mechanics, I decided to move them into common place.
This PR should be delivered before https://github.com/CosmWasm/cw-plus/pull/588. It will simplify then part of refactoring and therefore review.


uint (approved):
Nice

maurolacy (approved):
Looks good!

PR #591, opened 2021-12-21
Add `MIGRATING.md`

### LINKED ISSUE #583: Add `MIGRATING.md`
List migrations needed for 0.10 -> 0.11 breaking changes, with examples.

maurolacy:
Add breaking changes list / migration points for 0.10.x to 0.11.0.

Closes #583.

### DIFF
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -41,14 +41,7 @@ index 48a60083..2f2ef70d 100644
 --- a/contracts/cw3-fixed-multisig/src/contract.rs
 +++ b/contracts/cw3-fixed-multisig/src/contract.rs
 @@ -385,21 +384,20 @@ fn list_votes(
-     let limit = limit.unwrap_or(DEFAULT_LIMIT).min(MAX_LIMIT) as usize;
-     let start = start_after.map(Bound::exclusive);
-
--    let votes: StdResult<Vec<_>> = BALLOTS
-+    let votes = BALLOTS
-         .prefix(proposal_id)
--        .range_raw(deps.storage, start, None, Order::Ascending)
-+        .range(deps.storage, start, None, Order::Ascending)
+         .range(deps.storage, start, None, Order::Ascending)
          .take(limit)
          .map(|item| {
 -            let (key, ballot) = item?;
@@ -59,15 +52,6 @@ index 48a60083..2f2ef70d 100644
                  vote: ballot.vote,
                  weight: ballot.weight,
              })
-         })
--        .collect();
-+        .collect::<StdResult<_>>()?;
-
--    Ok(VoteListResponse { votes: votes? })
-+    Ok(VoteListResponse { votes })
- }
-
- fn query_voter(deps: Deps, voter: String) -> StdResult<VoterResponse> {
 ```
 
 If you don't need key deserialization, just can just rename `range` to `range_raw` and you are good to go.
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -1,0 +1,115 @@
+# Migrating
+
+This guide lists API changes between *cw-plus* major releases.
+
+## v0.10.3 -> v0.11.0
+
+### Breaking Issues / PRs
+
+- Rename cw0 to utils [\#471](https://github.com/CosmWasm/cw-plus/issues/471) / Cw0 rename [\#508](https://github.com/CosmWasm/cw-plus/pull/508)
+
+The `cw0` package was renamed to `utils`. The required changes are straightforward:
+
+```diff
+diff --git a/contracts/cw1-subkeys/Cargo.toml b/contracts/cw1-subkeys/Cargo.toml
+index 1924b655..37af477d 100644
+--- a/contracts/cw1-subkeys/Cargo.toml
++++ b/contracts/cw1-subkeys/Cargo.toml
+@@ -19,7 +19,7 @@ library = []
+ test-utils = []
+
+ [dependencies]
+-cw0 = { path = "../../packages/cw0", version = "0.10.3" }
++utils = { path = "../../packages/utils", version = "0.10.3" }
+ cw1 = { path = "../../packages/cw1", version = "0.10.3" }
+ cw2 = { path = "../../packages/cw2", version = "0.10.3" }
+ cw1-whitelist = { path = "../cw1-whitelist", version = "0.10.3", features = ["library"] }
+```
+
+```diff
+diff --git a/contracts/cw1-subkeys/src/contract.rs b/contracts/cw1-subkeys/src/contract.rs
+index b4852225..f20a65ec 100644
+--- a/contracts/cw1-subkeys/src/contract.rs
++++ b/contracts/cw1-subkeys/src/contract.rs
+@@ -8,7 +8,6 @@ use cosmwasm_std::{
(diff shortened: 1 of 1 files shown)


maurolacy:
Last issue before 0.11 release.

MIGRATING.md
```diff
@@ -0,0 +1,242 @@
...
++                voter: addr.into(),
+                 vote: ballot.vote,
+                 weight: ballot.weight,
+             })
+         })
+-        .collect();
++        .collect::<StdResult<_>>()?;
+
+-    Ok(VoteListResponse { votes: votes? })
++    Ok(VoteListResponse { votes })
```
ueco-jb, line 67: To keep this doc as simple as possible and avoid any confusion I'd only include actual changes needed for migration.
This:
```rust
let votes: StdResult<Vec<_>> = BALLOTS
...
.collect();
Ok(VoteListResponse { votes: votes? })
```
vs this
```rust
let votes = BALLOTS
...
.collect::<StdResult<_>>()?;
Ok(VoteListResponse { votes })
```
does exactly the same, so I'd remove it from diff (keep only one approach).

MIGRATING.md
```diff
@@ -0,0 +1,242 @@
...
+ }
+
+ fn query_voter(deps: Deps, voter: String) -> StdResult<VoterResponse> {
+```
+
+If / when you don't need key deserialization, just can just rename `range` to `range_raw` and you are good to go.
+
+If you need / want key deserialization for **indexes**, you need to specify the primary key type as a last (optional)
```
ueco-jb, line 75: nit: I think "When you don't.../if you want..." variant is good, no need for "if/when ... need/want".

MIGRATING.md
```diff
@@ -0,0 +1,242 @@
...
+argument in the index key specification. If not specified, it defaults to `()`, which means that primary keys will
+not only not be deserialized, but also not provided. This is for backwards-compatibility with current indexes
+specifications, and may change in the future once these features are stabilized.
+See `packages/storage-plus/src/indexed_map.rs` tests for reference.
+
+Also, as part of this issue, `keys` was renamed to `keys_raw`, and `keys_de` to `keys`. `prefix_range` was also renamed,
+to `prefix_range_raw`, and its key deserialization counterpart (`prefix_range_de`) to `prefix_range` in turn.
```
ueco-jb, line 82: Hmm, I think that instead of fiddling with language to not sound repetetive maybe consider just list of renamed types?
```
Renamed types:
- keys -> keys_raw
- keys_de -> keys
- prefix_range -> prefix_range_raw
- prefix_range_de -> prefix_range
```

MIGRATING.md
```diff
@@ -0,0 +1,242 @@
...
+index 022a4504..c7a3bb9d 100644
+--- a/packages/storage-plus/src/indexed_map.rs
++++ b/packages/storage-plus/src/indexed_map.rs
+@@ -295,8 +295,8 @@ mod test {
+     }
+
+     struct DataIndexes<'a> {
+-        // Second arg is for storing pk
+-        pub name: MultiIndex<'a, (String, String), Data, String>,
++        // Last args are for signaling pk deserialization
```
ueco-jb, line 166: If I remember my comment well: https://github.com/CosmWasm/cw-plus/pull/569#discussion_r763850654
this should be a `type parameter` instead of `args`. :)

maurolacy: Good eye. Took it from an old diff.

MIGRATING.md
maurolacy: Will do.

ueco-jb (approved):
lgtm

maurolacy:
Cool. Merging this, and cutting 0.11 next.

MIGRATING.md
```diff
@@ -0,0 +1,261 @@
...
++        // Use the index_key() helper to build the (raw) index key with an empty pk
+         let key = map.idx.name.index_key(key);
+         // Iterate using a bound over the raw key
+         let count = map
+```
+
+- Incorrect I32Key Index Ordering [\#489](https://github.com/CosmWasm/cw-plus/issues/489) /
+Signed int keys order [\#582](https://github.com/CosmWasm/cw-plus/pull/582)
+
+As part of range iterators revamping, we fixed the order of signed integer keys. You shouldn't change anything in your
```
ethanfrey, line 189: This is super important. Can we add this to the top of this MIGRATING.md section?

It can be here as well, but one or two sentences about it on top, then the details below would be nice.

(I know it is merged, but this can go anytime, people read MIGRATING.md from main

MIGRATING.md
```diff
@@ -0,0 +1,261 @@
...
+        return Err(ContractError::CannotMigrate {
+            previous_contract: version.contract,
+        });
+    }
+    // Original map
+    signed_int_map: Map<i8, String> = Map::new("signed_int_map");
+    // New map
+    signed_int_map_new: Map<i8, String> = Map::new("signed_int_map-v2");
+
+    signed_int_map
```
ethanfrey, line 215: This is dangerous (and I am rather sure doesn't compile).
You are writing to the same space you are reading from in an iterator.

let current = signed_int_map.range_raw(deps.storage, None, None, Order::Ascending).collect::<StdResult<Vec<_>>()?;

for (k, _) in current.iter() {
  // ?? how to delete... maybe IntKey8 uses the old format? maybe we can support OldIntKey8 that does it... would make it much easier
}

for (k, v) in current.iter() {
  let signed = i8::from_be_bytes(k);
  signed_int_map_new.save(deps.storage, signed, v)
}
```

This is very important issue, it would be good to have a test case for this. (Can be in 0.11.1)

Comment (maurolacy) [reply MIGRATING.md:215]: Will do. Yes, I wrote that more or less as pseudo-code from memory. Will work on a proper test case / working migration example.

Comment (maurolacy) [reply MIGRATING.md:215]: Please note this only affects signed int keys, which are rather rare.

Comment (maurolacy) [reply MIGRATING.md:215]: Also, all the other breaking changes, including  the one about removing the primary key from the multi index key spec, are non-breaking in terms of storage / data format, which is great.

Comment (maurolacy) [reply MIGRATING.md:215]: > You are writing to the same space you are reading from in an iterator.

Not sure this is what you mean, but notice both maps in the example have different namespace / prefix. It is assumed the new contract code will rely on the new namespace.

Comment (maurolacy) [reply MIGRATING.md:215]: > ```rust
>   // ?? how to delete... maybe IntKey8 uses the old format? maybe we can support OldIntKey8 that does it... would make it much easier
> ```

Good point, I hadn't thought about that. `IntKey` supports the new format as well. Maybe we can change it to support the old format. Or introduce the `OldIntKey` wrappers for it.

The way to do it now would be
```rust
let signed = i8::from_be_bytes(k);
signed_int_map.remove(deps.storage, IntKey8::from(-signed);
```
I think, but it would be good to confirm / test this.

ethanfrey: > > You are writing to the same space you are reading from in an iterator.
> Not sure this is what you mean, but notice both maps in the example have different namespace / prefix. It is assumed the new contract code will rely on the new namespace after migration.

Ah, that is a very good point/design.

In any case, one needs &Storage the other &mut Storage, and you cannot hold them both at the same time. (Thank you Rust)

PR #592, opened 2021-12-21
fix link from coral to sandynet-1

yubrew:
Fix README.md link for public testnet


ueco-jb (commented):
The only thing changed I see is space between pipes `|` - is that right or am I missing something?
GH shows this only as big blob, but as far as I can see everything seems the same?

README.md
```diff
@@ -47,7 +46,7 @@ many custom contracts.
 If you don't know what CosmWasm is, please check out
 [our homepage](https://cosmwasm.com) and
 [our documentation](https://docs.cosmwasm.com) to get more background.
-We are running a [public testnet](https://github.com/CosmWasm/testnets/blob/master/coralnet/README.md)
+We are running a [public testnet](https://github.com/CosmWasm/testnets/blob/master/sandynet-1/README.md)
```
ethanfrey, line 49: Thanks for this update

ethanfrey (commented):
One line here is a good update.

The rest is some "markdown linter" that ran through the code. This could be good, but only if we have a standard that everyone uses.

If you pull out that one line into it's own PR, I will merge it happily.

If you want to make a new PR adding some markdown linter script to https://github.com/CosmWasm/cw-plus/tree/main/scripts and then running it (and checking in the new markdown), I would be happy to review it.

Maybe use something like https://github.com/CosmWasm/cosmwasm/blob/main/devtools/format_md.sh
?

README.md
yubrew: Thanks for parsing through the linter junk to see the 1 liner. I'll look out for linter junk on future PRs.

ethanfrey: Can you update this PR to be that one-line change? (So you can get the commit credit)

Happy for a follow up to ensure consistent linting, but it must be a script, not just your IDE (and mine is different)

yubrew:
the git revert

PR #603, opened 2021-12-23
Adjust order of publishing to handle new deps

ethanfrey:
This died while publishing as cw3-flex depends on cw3-fixed and we didn't wait between publishing them (for crates.io index to update)

This should let new publish go smoother.


ueco-jb (approved):
lgtm

PR #604, opened 2021-12-23
Signed int keys migrate example

### LINKED ISSUE #602: Add signed int keys migration example
As a #591 follow-up, and after https://github.com/CosmWasm/cw-plus/pull/591#discussion_r774183703.

This might require introducing the old keys, so that they may be used for easy migration.

maurolacy:
Closes #602.

### DIFF
--- a/packages/storage-plus/src/de_old.rs
+++ b/packages/storage-plus/src/de_old.rs
@@ -1,0 +1,106 @@
+use std::array::TryFromSliceError;
+use std::convert::TryInto;
+
+use cosmwasm_std::{StdError, StdResult};
+
+use crate::de::KeyDeserialize;
+use crate::keys_old::IntKeyOld;
+
+macro_rules! intkey_old_de {
+    (for $($t:ty),+) => {
+        $(impl KeyDeserialize for IntKeyOld<$t> {
+            type Output = $t;
+
+            #[inline(always)]
+            fn from_vec(value: Vec<u8>) -> StdResult<Self::Output> {
+                Ok(<$t>::from_be_bytes(value.as_slice().try_into()
+                    .map_err(|err: TryFromSliceError| StdError::generic_err(err.to_string()))?))
+            }
+        })*
+    }
+}
+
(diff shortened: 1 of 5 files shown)


packages/storage-plus/src/de_old.rs
```diff
@@ -0,0 +1,106 @@
+use std::array::TryFromSliceError;
+use std::convert::TryInto;
+
+use cosmwasm_std::{StdError, StdResult};
+
+use crate::de::KeyDeserialize;
+use crate::keys_old::IntKeyOld;
```
ethanfrey, line 7: IntKeyOld. I love it

packages/storage-plus/src/map.rs
```diff
@@ -635,6 +638,77 @@ mod test {
...
+            vec![
+                (50, data3.clone()),
+                (-1234, data.clone()),
+                (-56, data2.clone())
+            ]
+        );
+
+        // remove old entries
+        for (k, _) in current.iter() {
+            SIGNED_ID_OLD.remove(&mut store, IntKeyOld::<i32>::from(*k));
```
ethanfrey, line 688: `k.into()` doesn't work here?

I thought `.remove()` is strongly typed arg, which makes this easier to write

packages/storage-plus/src/map.rs
```diff
@@ -635,6 +638,77 @@ mod test {
         assert_eq!(all, vec![(50, data3)]);
     }
 
+    #[test]
```
ethanfrey, line 641: Very nice test/method

ethanfrey (approved):
Looks good. Minor style comment

packages/storage-plus/src/de_old.rs
maurolacy: Yeah. It can be useful for people that are using signed int keys and don't want to migrate, by the way.

We can keep these around for a while, and then delete / clean up the whole thing.

packages/storage-plus/src/map.rs
maurolacy: It does. Changed.

PR #607, opened 2021-12-24
Implement HookMap

orkunkl:
`HookMap` enables developers to set up multiplexed hooks. This opens up flexibility for the hook design.

### DIFF
--- a/packages/controllers/src/hooks.rs
+++ b/packages/controllers/src/hooks.rs
@@ -6,7 +6,7 @@ use thiserror::Error;
 use cosmwasm_std::{
     attr, Addr, Deps, DepsMut, MessageInfo, Response, StdError, StdResult, Storage, SubMsg,
 };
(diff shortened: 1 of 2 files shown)


packages/controllers/src/hooks.rs
```diff
@@ -123,3 +123,102 @@ impl<'a> Hooks<'a> {
 }
 
 // TODO: add test coverage
+
+/// HookMap stores all hook addresses with given primary key. Enables hook multiplexing based on key.
```
ethanfrey, line 127: Why not just have 2 Hooks, one for each type.

We used this in some code, and as the constructor takes a storage key, you can ensure they don't overlap.

Using a different const provides better type safety than a map. Are there cases that need an arbitrary number of types of hooks? I have come across cases that have 1 or 2 types of callbacks, not N.

PR #606, opened 2021-12-24
Implement Expired variant, Scheduled

orkunkl:
This PR implements `Scheduled` type. The reason behind this is, `Expiration` type represents an expiration event although the logic is great for representing future events. 

- Duplicated `Expiration` and deleted `Scheduled::Never`

PR #608, opened 2021-12-24
Implement airdrop start expire

orkunkl:
Blocked by https://github.com/CosmWasm/cw-plus/pull/606

Useful feature.

### DIFF
--- a/contracts/cw20-merkle-airdrop/src/contract.rs
+++ b/contracts/cw20-merkle-airdrop/src/contract.rs
@@ -6,6 +6,7 @@ use cosmwasm_std::{
 };
 use cw2::{get_contract_version, set_contract_version};
 use cw20::Cw20ExecuteMsg;
+use cw_utils::Expiration;
 use sha2::Digest;
 use std::convert::TryInto;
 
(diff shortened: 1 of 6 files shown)


contracts/cw20-merkle-airdrop/src/contract.rs
```diff
@@ -124,12 +140,25 @@ pub fn execute_register_merkle_root(
...
 pub fn execute_claim(
     deps: DepsMut,
-    _env: Env,
+    env: Env,
     info: MessageInfo,
     stage: u8,
     amount: Uint128,
     proof: Vec<String>,
 ) -> Result<Response, ContractError> {
+    // airdrop begun
```
ethanfrey, line 149: I would pull out these lines (the green diff) into a function like:

`is_active(deps: Deps, env: Env, stage: u8) -> StdResult<bool> {}`

which you can call here and easily unit test. You may well want to use the same check somewhere else later.

contracts/cw20-merkle-airdrop/src/contract.rs
```diff
@@ -485,6 +526,88 @@ mod tests {
...
+        let info = mock_info("owner0000", &[]);
+        let msg = ExecuteMsg::RegisterMerkleRoot {
+            merkle_root: "5d4f48f147cb6cb742b376dce5626b2a036f69faec10cd73631c791780e150fc"
+                .to_string(),
+            expiration: None,
+            start: Some(Scheduled::AtHeight(200_000)),
+        };
+        execute(deps.as_mut(), env.clone(), info.clone(), msg).unwrap();
+
+        // can't claim expired
```
ethanfrey, line 594: comment needs updating "can't claim before it starts"

ethanfrey (commented):
Looks good.
Please rebase now that #606 was merged, so this makes a cleaner diff.
I made a small style suggestion but the code looks correct, so it is optional.

ethanfrey:
Please move to https://github.com/CosmWasm/cw-tokens

You cannot transfer PR. You may be able to cherry-pick commits if you are a git super-star. Or simply apply the patch. I kept the same contract names and paths.

PR #614, opened 2021-12-29
Cw derive generic

hashedone:
Proper generation of structs for generic traits - including where clauses propagation where appreciate.

### DIFF
--- a/packages/cw-derive/examples/generics.rs
+++ b/packages/cw-derive/examples/generics.rs
@@ -1,0 +1,16 @@
+#![allow(dead_code)]
+use cosmwasm_std::{CosmosMsg, Response};
+
+struct Ctx;
+struct Error;
+
+#[cw_derive::interface(module=msg, exec=Cw1Exec, query=Cw1Query)]
+trait Cw1<Msg>
+where
+    Msg: std::fmt::Debug + PartialEq + Clone + schemars::JsonSchema,
+{
+    #[msg(exec)]
+    fn execute(&self, ctx: Ctx, msgs: Vec<CosmosMsg<Msg>>) -> Result<Response, Error>;
+}
+
+fn main() {}
--- a/packages/cw-derive/examples/generics.rs
+++ b/packages/cw-derive/examples/generics.rs
@@ -1,11 +1,11 @@
 #![allow(dead_code)]
 use cosmwasm_std::{CosmosMsg, Response};
 
-struct Ctx;
-struct Error;
+pub struct Ctx;
+pub struct Error;
 
 #[cw_derive::interface(module=msg, exec=Cw1Exec, query=Cw1Query)]
-trait Cw1<Msg>
+pub trait Cw1<Msg>
 where
     Msg: std::fmt::Debug + PartialEq + Clone + schemars::JsonSchema,
 {
--- a/packages/cw-derive/tests/messages_generation.rs
+++ b/packages/cw-derive/tests/messages_generation.rs
@@ -1,0 +1,36 @@
+use cosmwasm_std::{Addr, Decimal, Response};
+
+pub struct Ctx;
+pub struct Error;
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq, schemars::JsonSchema)]
+pub struct QueryResult;
+
+#[cw_derive::interface]
+pub trait Interface {
+    #[msg(exec)]
+    fn no_args_execution(&self, ctx: Ctx) -> Result<Response, Error>;
+
+    #[msg(exec)]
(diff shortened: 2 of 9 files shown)


packages/cw-derive/examples/generics.rs
```diff
@@ -0,0 +1,16 @@
+#![allow(dead_code)]
+use cosmwasm_std::{CosmosMsg, Response};
+
+pub struct Ctx;
+pub struct Error;
+
+#[cw_derive::interface(module=msg, exec=Cw1Exec, query=Cw1Query)]
```
ethanfrey, line 7: It is hard to check what these macros actually expand into.

Maybe a simple test case that constructs a valid ExecuteMsg.

```rust
#[cfg(test)]
mod test {
    #[derive(Debug, PartialEq, Clone, JsonSchema)]
    struct MyMsg {
        pub foo: String;
    }

    #[test]
    // this just ensures that the autogenerated message types can be constructed as expected
    fn check_msg() {
        let exec = Cw1Exec<MyMsg>::Execute{
           msgs: vec![
               BankMsg::Burn{ amount: coins(500, "atom") }.into(), 
               CosmosMsg::Custom(MyMsg{ foo: "hello world".to_string() }),
           ],
        };
        cosmwasm_std::to_binary(&exec).unwrap();
    }
}
```

ethanfrey (commented):
Nice approach.
I propose adding some tests to ensure what the autogenerated code looks like.

packages/cw-derive/examples/generics.rs
hashedone: Next thing I am working right now are actual dispatch functions, and there would be tests like you shown, including checking if everythign works on this stage.

I always check what is generated with `cargo expand` - being honest I work with it all the time fixing small generation issues, but I agree it is not good enough that I know what is generated, tests are needed. I just wanted to have API allowing to test behavior on any level (so next small step will include real tests).

packages/cw-derive/tests/messages_generation.rs
```diff
@@ -0,0 +1,36 @@
...
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq, schemars::JsonSchema)]
+pub struct QueryResult;
+
+#[cw_derive::interface]
+pub trait Interface {
+    #[msg(exec)]
+    fn no_args_execution(&self, ctx: Ctx) -> Result<Response, Error>;
+
+    #[msg(exec)]
+    fn argumented_execution(&self, ctx: Ctx, addr: Addr, coef: Decimal, desc: String);
```
ethanfrey, line 15: Shouldn't this have a return value? Shall we/ can we enforce `Result<Response, E>`?

packages/cw-derive/tests/messages_generation.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    #[msg(query)]
+    fn no_args_query(&self, ctx: Ctx) -> Result<QueryResult, Error>;
+
+    #[msg(query)]
+    fn argumented_query(&self, ctx: Ctx, user: Addr) -> Result<QueryResult, Error>;
+}
+
+#[test]
+fn messages_constructible() {
```
ethanfrey, line 25: Nice test for compiling shape.

Another one to test that no other cases are generated would be:

```rust
fn parse_exec(e: ExecMsg) -> i64 {
  match e {
    ExecMsg:NoArgsExecution {} =>1,
    ExecMsg::ArgumentedExecution { .. } => 2,
  } 
}
```

Not sure if so useful, but another idea for a basic "does it compile" test to cover the codegen

packages/cw-derive/tests/messages_renaming.rs
```diff
@@ -0,0 +1,36 @@
+use cosmwasm_std::{Addr, Decimal, Response};
+
+pub struct Ctx;
+pub struct Error;
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq, schemars::JsonSchema)]
+pub struct QueryResult;
+
+#[cw_derive::interface(exec=IfaceExec, query=IfaceQuery)]
```
ethanfrey, line 9: Nice.
I like that you cover these customisations.

ethanfrey (commented):
I like this direction.

packages/cw-derive/tests/messages_generation.rs
hashedone: It should. Corrected in follow up. I will not add it here to avoid rebasing issues.

packages/cw-derive/tests/messages_generation.rs
hashedone: I can add it - not sure about usefullness, but wont hurt. But in followup (to avoid conflicts).

PR #615, opened 2021-12-29
Implement SafeLock controller

orkunkl:
I've come across the usage of this logic a few times. Thought would be good to have it in cw-plus.
Not sure about the naming though. Maybe `KillSwitch`, `SafetyLock`?

### DIFF
--- a/packages/controllers/src/lib.rs
+++ b/packages/controllers/src/lib.rs
@@ -1,7 +1,9 @@
 mod admin;
 mod claim;
 mod hooks;
+mod safe_lock;
 
 pub use admin::{Admin, AdminError, AdminResponse};
 pub use claim::{Claim, Claims, ClaimsResponse};
 pub use hooks::{HookError, Hooks};
+pub use safe_lock::{SafeLock, SafeLockResponse};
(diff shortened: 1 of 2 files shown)


ethanfrey:
This is so simpler, I don't really see the need as is.
Combining it with some logic for authorising this (eg admin) would make more sense to me.

Curious to see a link to a few pieces of code that do use it.

orkunkl:
I've seen this use in Loterra contract and nft-marketplace contract that I implement. I think having a safety lock in the first stages make sense.

ethanfrey:
> I've seen this use in Loterra contract and nft-marketplace contract that I implement. I think having a safety lock in the first stages make sense.

I agree with the use of some safety lock. I think the code extracted here is so small, it doesn't really warrant extraction.

I would look at the whole logic of the other contracts and extract a little bigger piece out. Like maybe wrapping an Admin as well, so this handles auth checks as well.

`let is_enabled = self.0.load(deps.storage)?` doesn't really warrant a type and a method in my mind.

ethanfrey:
Maybe just documenting the pattern is enough here.

Controllers should have enough logic on top of the Map to make it meaningful for reuse

orkunkl:
Agree with you on the simple part. I wanted to leave the authorization part to the developer, to provide flexible usage.

ethanfrey:
This is just an `Item<bool>` with some explanation.
I would just add some docs on the usage, not some type.

orkunkl:
Cool closing the PR.

PR #618, opened 2021-12-30
cw-derive: Dispatching on messages

hashedone:
Actual dispatching on messages to traits method.

~Generic messages still needs a one little trick to be working, but I got tired for today - will be in next batch.~

I was wrong. Dispatching on generics works just fine. I just needed sleep. Test on that also included.

### DIFF
--- a/packages/cw-derive/examples/basic.rs
+++ b/packages/cw-derive/examples/basic.rs
@@ -1,30 +1,41 @@
 #![allow(dead_code)]
-use cosmwasm_std::Response;
-
-pub struct Ctx;
-pub struct Error;
+use cosmwasm_std::{DepsMut, Env, MessageInfo, Response};
 
 #[derive(serde::Serialize, serde::Deserialize, Clone, Debug, PartialEq, schemars::JsonSchema)]
 pub struct Member;
 
 #[cw_derive::interface(module=msg, exec=Execute, query=Query)]
 pub trait Cw4 {
+    type Error;
+
     #[msg(exec)]
-    fn update_admin(&self, ctx: Ctx, admin: Option<String>) -> Result<Response, Error>;
+    fn update_admin(
+        &self,
+        ctx: (DepsMut, Env, MessageInfo),
+        admin: Option<String>,
+    ) -> Result<Response, Self::Error>;
 
     #[msg(exec)]
     fn update_members(
         &self,
-        ctx: Ctx,
+        ctx: (DepsMut, Env, MessageInfo),
         remove: Vec<String>,
         add: Vec<Member>,
-    ) -> Result<Response, Error>;
+    ) -> Result<Response, Self::Error>;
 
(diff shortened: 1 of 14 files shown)


maurolacy:
I lack some context here. Perhaps an initial PR with the basic `cw-derive` changes (new directory, cargo files, etc.) and a nice description / link to an issue would have helped.

packages/cw-derive/examples/basic.rs
```diff
@@ -1,30 +1,41 @@
...
 
 #[cw_derive::interface(module=msg, exec=Execute, query=Query)]
 pub trait Cw4 {
+    type Error;
+
     #[msg(exec)]
-    fn update_admin(&self, ctx: Ctx, admin: Option<String>) -> Result<Response, Error>;
+    fn update_admin(
+        &self,
+        ctx: (DepsMut, Env, MessageInfo),
```
ethanfrey, line 14: Happy if you add a type for this tuple:

`pub type CtxMut = (DepsMut, Env, MessageInfo);`
`pub type Ctx = (Deps, Env);`

Fine for the shape of it, just seems like it would be useful for an alias. Or maybe there are too many variants and it doesn't make sense?

Actually, would be interesting to see how this works in practice when destructing as well...

Would you:

`let (deps, _ info) = ctx;` ?
Or `update_admin(&self, (deps, _info): CtxMut, admin: Option<String>)` ?
Or `ctx.0.storage` ?

packages/cw-derive/src/check_generics.rs
```diff
@@ -17,12 +17,24 @@ impl<'g> CheckGenerics<'g> {
     pub fn used(self) -> Vec<&'g Ident> {
         self.used
     }
+
+    /// Returns split between used and unused generics
+    pub fn used_unused(self) -> (Vec<&'g Ident>, Vec<&'g Ident>) {
```
ethanfrey, line 22: Can you rename to `split_used_unused`?
The name is confusing until reading comment and code.

ethanfrey (commented):
A few questions

maurolacy:
Why don't we stick with parameters? i.e. `deps`, `env`, `info`. I suppose the first thing would be to create those parameters, if they come in a tuple.`ctx.deps` is better than `ctx.0`, but still, I think `deps` is even better.

packages/cw-derive/examples/basic.rs
maurolacy: A mentioned, why not just the classic `deps`, `env` and `info`?

maurolacy (commented):
Looks good to me, though I currently lack the required knowledge about procedural macros this review requires, or merits.

ethanfrey:
Is there a PR open for cw-derive?
I couldn't find it.

It would be nice to rebase that to current main and open a WIP PR for more visibility.

PR #619, opened 2022-01-03
Signed int key benchmarks

maurolacy:
Just an exercise to confirm the different methods of doing `to-cw-bytes` are not incurring a significant performance loss.

Not sure this needs to be merged. Just but published it for reference.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,6 +33,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9b34d609dfbaf33d6889b2b7106d3ca345eacad44200913df5ba02bfd31d2ba9"
 
 [[package]]
+name = "atty"
+version = "0.2.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
+dependencies = [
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -25,6 +25,18 @@ workflows:
...
+            - package_storage_plus
+          filters:
+            branches:
+              only:
+                # Long living branches
+                - main
+                - /^[0-9]+\.[0-9]+$/
+                # 👇Add your branch here if benchmarking matters to your work
+                - benchmarking
+                - signed-int-key-bench
```
ueco-jb, line 39: :+1: for filtering, although I would still cut that list a little bit shorter. 
Maybe remove release tag branches, since they are basically always directly derive from `main` anyway?

ueco-jb (approved):
I like it. 
I wonder if there's some feasible way to present those results in good way? Criterion prints a lot, and I personally use CircleCI as rare as possible.

maurolacy:
> I wonder if there's some feasible way to present those results in good way? Criterion prints a lot, and I personally use CircleCI as rare as possible.

You mean, besides `cargo bench` default output? There a number of options for displaying results (see https://doc.rust-lang.org/cargo/commands/cargo-bench.html#display-options), but now sure how to use them.

That said, `-- --output-format bencher` produces a terser output.

PR #620, opened 2022-01-03
Remove `IntKey` and `TimestampKey`

ueco-jb:
closes https://github.com/CosmWasm/cw-plus/issues/570

### DIFF
--- a/packages/storage-plus/src/de.rs
+++ b/packages/storage-plus/src/de.rs
@@ -3,7 +3,7 @@ use std::convert::TryInto;
 
 use cosmwasm_std::{Addr, StdError, StdResult};
 
-use crate::int_key::IntKey;
+use crate::int_key::CwIntKey;
 
 pub trait KeyDeserialize {
     type Output: Sized;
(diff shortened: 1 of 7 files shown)


packages/storage-plus/src/de.rs
```diff
@@ -1,13 +1,9 @@
...
-#![allow(deprecated)]
-
 use std::array::TryFromSliceError;
 use std::convert::TryInto;
 
 use cosmwasm_std::{Addr, StdError, StdResult};
 
-use crate::int_key::CwIntKey;
-use crate::keys::{IntKey, TimestampKey};
+use crate::int_key::IntKey;
```
maurolacy, line 6: I'm a little worried about this renaming introducing confusion. The module name is different, though. Perhaps that's enough to disambiguate / clarify this change.

maurolacy (approved):
Looks good.

Only doubt I have is with renaming `CwIntKey` in this PR / release round.

We'll leave `IntKeyOld` for a while as well. So, perhaps doing the renaming above when removing `IntKeyOld` is a good idea.

Approving anyway, as I think this is a viable option.

packages/storage-plus/src/de.rs
ueco-jb: Had similar doubts.
All right, I reverted that commit - now it could be cherry-picked some time in the future.

maurolacy: Now tracked in #775.

PR #621, opened 2022-01-03
Expose snapshot map changelog

ben2x4:
Really basic attempt at allowing queries of the underlying changelog of a snapshot map. This is a blocker for a staking rewards contract I am working on. Interested in feedback and alternative approaches to implement this feature.

It feels a little clunky to have to bubble up this function through multiple files due to the snapshots field being private. After writing this code I am starting to think the approach from https://github.com/CosmWasm/cw-plus/pull/600 might make more sense then originally consider.

Thanks everyone!


packages/storage-plus/src/snapshot/map.rs
```diff
@@ -9,7 +9,7 @@ use crate::keys::PrimaryKey;
 use crate::map::Map;
 use crate::path::Path;
 use crate::prefix::{namespaced_prefix_range, Prefix, PrefixBound};
```
maurolacy, line 11: In general, please remove unused imports. Running `cargo clippy -- -D warnings` will flag those and more.

packages/storage-plus/src/snapshot/map.rs
```diff
@@ -262,6 +262,33 @@ where
     }
 }
 
+#[cfg(feature = "iterator")]
+impl<'a, K, T> SnapshotMap<'a, K, T>
+where
+    T: Serialize + DeserializeOwned + Clone,
+    K: PrimaryKey<'a> + Prefixer<'a> + KeyDeserialize,
+{
+    pub fn changelog_range<'c>(
```
maurolacy, line 271: I think this approach is in the right direction, but not quite there yet.

What I recommend is to generalize this, i.e. instead of providing an ad-hoc `range` method that does what you need, provide instead the entire set of functions, both for compatibility and completeness.

packages/storage-plus/src/snapshot/map.rs
```diff
@@ -262,6 +262,33 @@ where
...
+    where
+        T: 'c,
+        'a: 'c,
+        K: 'c,
+        K::Output: 'static,
+    {
+        self.snapshots
+            .changelog
+            .prefix(k.clone())
+            .range(store, min, max, order)
```
maurolacy, line 288: You are combining / hard-coding `prefix` and `range` here.

As mentioned above, instead of doing that it would be better to follow the `range` contract for `changelog_range` as well. And also add a `changelog_prefix` method.

Now, I agree with your comment that making `snapshot` and `changelog` public will basically provide the same, without requiring these wrapper / accessor methods.

Perhaps simply writing a `changelog()` method is the best approach here. It will return the underlying changelog map, and then you can prefix / iterate over it as you see fit.

@ethanfrey , what do you think?

maurolacy: > Perhaps simply writing a `changelog()` method is the best approach here. It will return the underlying changelog map, and then you can prefix / iterate over it as you see fit.

Of course, this would allow writing to the changelog directly, which is wrong. But, you can always do that anyway, by playing with the namespaces. So, I don't see this is a real stopper here.

maurolacy: Best thing would be to return a `MapView` struct, that doesn't have the `save` / `update` / `remove` methods, but I think that's out of scope here.

maurolacy: See #622 for an implementation of this approach.

ben2x4: amazing! Thank you so much for the help! 

I think that is a great approach.

ben2x4:
This was resolved with #622

PR #622, opened 2022-01-04
Publish snapshot map changelog

maurolacy:
Related to #600 and #621.

To allow prefixing / iteration over the changelog entries.

### DIFF
--- a/packages/storage-plus/src/map.rs
+++ b/packages/storage-plus/src/map.rs
@@ -31,6 +31,10 @@ impl<'a, K, T> Map<'a, K, T> {
             key_type: PhantomData,
         }
     }
+
+    pub fn namespace(&self) -> &'a [u8] {
+        self.namespace
+    }
 }
 
 impl<'a, K, T> Map<'a, K, T>
@@ -38,10 +42,6 @@ where
     T: Serialize + DeserializeOwned,
     K: PrimaryKey<'a>,
 {
-    pub fn namespace(&self) -> &'a [u8] {
-        self.namespace
-    }
-
     pub fn key(&self, k: K) -> Path<T> {
         Path::new(
             self.namespace,
--- a/packages/storage-plus/src/snapshot/item.rs
+++ b/packages/storage-plus/src/snapshot/item.rs
@@ -5,13 +5,13 @@ use cosmwasm_std::{StdError, StdResult, Storage};
 
 use crate::snapshot::{ChangeSet, Snapshot};
 use crate::{Item, Map, Strategy};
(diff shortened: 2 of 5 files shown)


ethanfrey (approved):
LGTM.
Does this solve your issue @ben2x4 ?

maurolacy:
Tried to extend this to `SnapshotItem` as well, but hit some issues with Unit `()` deserialization. I think this is good enough for now, and we can extend to item when / if needed.

maurolacy:
> Tried to extend this to `SnapshotItem` as well, but hit some issues with Unit `()` deserialization. I think this is good enough for now, and we can extend to item only when / if needed.

But see latest version.

packages/storage-plus/src/map.rs
```diff
@@ -31,17 +31,17 @@ impl<'a, K, T> Map<'a, K, T> {
             key_type: PhantomData,
         }
     }
+
```
ethanfrey, line 34: good fix regardless of this PR

packages/storage-plus/src/snapshot/item.rs
```diff
@@ -45,6 +46,11 @@ impl<'a, T> SnapshotItem<'a, T> {
     pub fn remove_checkpoint(&self, store: &mut dyn Storage, height: u64) -> StdResult<()> {
         self.snapshots.remove_checkpoint(store, height)
     }
+
+    pub fn changelog(&self) -> Map<u64, ChangeSet<T>> {
+        // Build and return a compatible Map with the proper key type
+        Map::new(unsafe { from_utf8_unchecked(self.snapshots.changelog.namespace()) })
```
ethanfrey, line 52: Why not expose this functionality as a method on Snapshot? (`self.snapshots.changelog()`)

ethanfrey (approved):
Makes sense. Minor style comment

ben2x4:
> LGTM. Does this solve your issue @ben2x4 ?

Yes, I like this approach. Give the user direct access to the changelog without making the whole underlying snapshot item public, very good compromise between the two approaches I suggested.

packages/storage-plus/src/snapshot/item.rs
maurolacy: No, because I'm overriding the `Map` key type here. Our empty (`Unit`) keys have a particular behaviour, in that they are not length-prefixed zero-length elements in tuples (they are just being skipped, when building the key).
 
What I can do to improve on this is to save the changelog namespace in this class, so that I can use it here directly, and avoid the conversion to unicode. Will do.

PR #627, opened 2022-01-10
Type safe bounds

### LINKED ISSUE #462: Make `Bound`s type safe
Current `Bound`s take a `Vec<u8>` as parameter, which is not related to the types the bound will apply to. This forces the user to build the proper bound manually, which requires knowledge of the internal structure used to store the different types. This is particularly annoying with composite keys, where a helper (`joined_key`) must be used to concatenate the key components.

Let's improve on this by making `Bound` type-safe. This may be tricky to do, as currently `Bound` is independent of the map-like structure it can be applied to.

There's a `key()` (`index_key` in the case if indexed maps) method in map-like structs, that can be used as an alternative to building the key manually. But, it only supports the full key, not the prefix / sub-prefix keys. An alternative could be to extend this with a couple of `prefix_key()` and `sub_prefix_key()` helpers.

Let's evaluate first if we can do better than that.

maurolacy:
Closes #462. Implements typed range bounds, following the same approach than `PrefixBound`. Also introduces a couple raw variants, for flexibility and generality.

Did all the changes in this PR, except for some refactoring, that will be published separately.
Lots of changes, but mostly cosmetic. The main impls are in keys.rs, where `Bound` was renamed to `RawBound`, and a new typed `Bound` was introduced. All this will be moved to its own `bound.rs` module for clarity (in a follow-up).

There's also a new `Bounder` trait, to build bounds from the keys (i.e. `key.inclusive_bound()` syntax), but it is mostly accessory / cosmetic.

Some small follow-ups / questions from here:

- Deprecate `prefix_range`? After this new syntax for bounds, `prefix_range` is a little redundant. It's still clearer than using composite keys with empty elements. So, I'm not sure here. 
- Deprecate / adapt raw index key helpers. We either adapt the `index_key` helpers to return typed keys, or deprecate them entirely.
- Missing `prefix_range_raw()` impls. If we don't deprecate it, let's add these for completeness in:
    - `SnapshotMap`
    - `IndexedSnapshotMap`
    - `UniqueIndex`
- Missing `sub_prefix_range/_raw()`. If we don't deprecate it, let's decide if we add these or not (only really useful for triple keys).
- Simplify / consolidate trait bound definitions. There are a lot of trait bounds over these structs, and perhaps some of those can be simplified / consolidated.

**Note**: ~~CI checks are failing due to the no-iterator variant. Solved in #629.~~ (fixed / merged)

### DIFF
--- a/contracts/cw1155-base/src/contract.rs
+++ b/contracts/cw1155-base/src/contract.rs
@@ -3,7 +3,7 @@ use cosmwasm_std::{
     to_binary, Addr, Binary, Deps, DepsMut, Env, MessageInfo, Order, Response, StdResult, SubMsg,
     Uint128,
 };
-use cw_storage_plus::RawBound;
+use cw_storage_plus::Bound;
 
 use cw1155::{
     ApproveAllEvent, ApprovedForAllResponse, BalanceResponse, BatchBalanceResponse,
@@ -494,7 +494,7 @@ fn query_all_approvals(
     limit: Option<u32>,
 ) -> StdResult<ApprovedForAllResponse> {
     let limit = limit.unwrap_or(DEFAULT_LIMIT).min(MAX_LIMIT) as usize;
-    let start = start_after.map(|addr| RawBound::exclusive(addr.as_ref()));
+    let start = start_after.as_ref().map(Bound::exclusive);
 
     let operators = APPROVES
         .prefix(&owner)
@@ -513,7 +513,7 @@ fn query_tokens(
     limit: Option<u32>,
 ) -> StdResult<TokensResponse> {
     let limit = limit.unwrap_or(DEFAULT_LIMIT).min(MAX_LIMIT) as usize;
-    let start = start_after.map(RawBound::exclusive);
+    let start = start_after.as_ref().map(|s| Bound::exclusive(s.as_str()));
 
     let tokens = BALANCES
         .prefix(&owner)
@@ -529,7 +529,7 @@ fn query_all_tokens(
     limit: Option<u32>,
 ) -> StdResult<TokensResponse> {
     let limit = limit.unwrap_or(DEFAULT_LIMIT).min(MAX_LIMIT) as usize;
-    let start = start_after.map(RawBound::exclusive);
+    let start = start_after.as_ref().map(|s| Bound::exclusive(s.as_str()));
     let tokens = TOKENS
         .keys(deps.storage, start, None, Order::Ascending)
         .take(limit)
--- a/contracts/cw1155-base/src/contract.rs
+++ b/contracts/cw1155-base/src/contract.rs
@@ -3,7 +3,7 @@ use cosmwasm_std::{
     to_binary, Addr, Binary, Deps, DepsMut, Env, MessageInfo, Order, Response, StdResult, SubMsg,
     Uint128,
 };
-use cw_storage_plus::Bound;
+use cw_storage_plus::RawBound;
 
 use cw1155::{
     ApproveAllEvent, ApprovedForAllResponse, BalanceResponse, BatchBalanceResponse,
@@ -494,7 +494,7 @@ fn query_all_approvals(
     limit: Option<u32>,
 ) -> StdResult<ApprovedForAllResponse> {
     let limit = limit.unwrap_or(DEFAULT_LIMIT).min(MAX_LIMIT) as usize;
-    let start = start_after.map(|addr| Bound::exclusive(addr.as_ref()));
+    let start = start_after.map(|addr| RawBound::exclusive(addr.as_ref()));
 
     let operators = APPROVES
         .prefix(&owner)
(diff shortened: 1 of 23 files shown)


packages/storage-plus/src/prefix.rs
```diff
@@ -8,40 +8,49 @@ use std::ops::Deref;
...
+    pub fn inclusive<T: Into<K>>(k: T) -> Self {
+        Self::Inclusive((k.into(), PhantomData))
     }
 
-    /// Turns an int, like Option<u32> into an inclusive bound
-    pub fn inclusive_int<T: CwIntKey + Endian>(limit: T) -> Self {
-        Bound::Inclusive(limit.to_cw_bytes().into())
+    pub fn exclusive<T: Into<K>>(k: T) -> Self {
+        Self::Exclusive((k.into(), PhantomData))
     }
```
ueco-jb, line 45: I wonder, wouldn't it make sense to include additional `inclusive_raw` and `exclusive_raw` helpers, which would contain those `.into()` inside implementations? I think it could be used in couple places at top of this PR.

maurolacy: I thought about that, and decided not to. On the one hand, building these is really simple. On the other, it's better to discourage using these. If you need them, it's probably because you are doing something dodgy.

ueco-jb (approved):
lgtm

Being nitpicky as I am, I'd consider changing test names slightly. I'm not sure what exactly `range2_...` means.

Initially I wanted to point out, that `bound` imports in `unique.rs` and `multi.rs` are without `cfg = iteration` mark, but then I noticed that types that uses them have those decorators in `lib.rs`. Nice that you were so precise about that.

Also, iterating over prefix is neat!

maurolacy:
> lgtm
> 
> Being nitpicky as I am, I'd consider changing test names slightly. I'm not sure what exactly `range2_...` means.

Will fix those. Thanks.

contracts/cw1155-base/src/contract.rs
```diff
@@ -513,7 +513,7 @@ fn query_tokens(
     limit: Option<u32>,
 ) -> StdResult<TokensResponse> {
     let limit = limit.unwrap_or(DEFAULT_LIMIT).min(MAX_LIMIT) as usize;
-    let start = start_after.map(Bound::exclusive);
+    let start = start_after.as_ref().map(|s| Bound::exclusive(s.as_str()));
```
ethanfrey, line 516: Curious as to why `as_ref()` is needed.... ah, never mind lifetimes... we cannot consume the original `start_after` until `start` is completed as it maintains a reference there...

contracts/cw20-base/src/enumerable.rs
```diff
@@ -16,7 +16,7 @@ pub fn query_all_allowances(
 ) -> StdResult<AllAllowancesResponse> {
     let owner_addr = deps.api.addr_validate(&owner)?;
     let limit = limit.unwrap_or(DEFAULT_LIMIT).min(MAX_LIMIT) as usize;
-    let start = start_after.map(Bound::exclusive);
+    let start = start_after.map(|s| Bound::ExclusiveRaw(s.as_bytes().into()));
```
ethanfrey, line 19: This makes a clone I think.

Doesn't [`s.into_bytes()`](https://doc.rust-lang.org/stable/std/string/struct.String.html#method.into_bytes) do the same without a clone?

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -287,7 +287,7 @@ fn list_proposals(
     limit: Option<u32>,
 ) -> StdResult<ProposalListResponse> {
     let limit = limit.unwrap_or(DEFAULT_LIMIT).min(MAX_LIMIT) as usize;
-    let start = start_after.map(Bound::exclusive_int);
+    let start = start_after.map(Bound::exclusive);
```
ethanfrey, line 290: Nice this just works

packages/storage-plus/src/indexed_map.rs
```diff
@@ -244,11 +212,34 @@ where
         Box::new(mapped)
     }
 
+    pub fn range_raw<'c>(
+        &self,
+        store: &'c dyn Storage,
+        min: Option<Bound<'a, K>>,
```
ethanfrey, line 218: So this is type-safe, unless they explicitly create a `Bound::ExclusiveRaw` or such, so we still allow a nice backwards-compatible path.

Nice design.

ethanfrey (commented):
Finally got to review it. Looks very nice and impressive you got this to work without much breaking APIs.

A few comments to consider

contracts/cw1155-base/src/contract.rs
maurolacy: It's because the key itself is a reference, so we need an `&Addr` for the bound here. This is a way to obtain a reference to something that is in scope, when it's inside an `Option`.

contracts/cw3-fixed-multisig/src/contract.rs
maurolacy: Finally!

packages/storage-plus/src/indexed_map.rs
maurolacy: Exactly. The raw types are there only for cases when you explicitly want /need to break the "type-safedness".

Thanks. It was fun to do this, even when it involved a lot of side-tracking of the already existing code.

PR #629, opened 2022-01-10
Type safe bounds refactor

maurolacy:
#627 follow-up. Refactored bounds-related stuff into its own module. This helps fix features, and so the no-iterator variant builds properly here.

### DIFF
--- a/packages/storage-plus/src/bound.rs
+++ b/packages/storage-plus/src/bound.rs
@@ -1,0 +1,184 @@
+#![cfg(feature = "iterator")]
+
+use cosmwasm_std::Addr;
+use std::marker::PhantomData;
+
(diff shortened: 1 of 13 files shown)


ueco-jb (approved):
As far as I can see, there are no changes of implementation, right?
Off you go, then.

maurolacy:
This only changes the `range` methods signature. And `Bound` is a different impl, of course. All the rest stays the same.

Then there are some additions / improvements, like the `Bounder` trait impls.

And some removals, like the now unused `RawBound` helpers.

uint (approved):
LGTM

maurolacy:
Thanks. Please review #627 when you find some time. I'm refraining to merge this on top to ease reviewing.

PR #631, opened 2022-01-17
Update ics20 contract

ethanfrey:
Freshen up the contract a bit and add an important improvement to bring it closer to production:

-> Only allowed cw20 tokens can be sent, and they may have a gas limit set to use when transferring them.

As it currently stands, you could send a token from a malicious cw20 contract and when someone sent it back from the destination chain, it would run an infinite loop (or just error). This would kill any relayer trying to send this in a batch with other packets and could (temporarily) make a channel useless until the timeout arrived.

If the cw20 contract is not malicious, the existing code works fine. This allows a gov_contract that can only add new tokens there. This gov_contract cannot disable existing token sends. (Design for minimally intrusive governance that can manage this)

### DIFF
--- a/contracts/cw20-ics20/src/contract.rs
+++ b/contracts/cw20-ics20/src/contract.rs
@@ -122,7 +122,7 @@ pub fn execute_transfer(
     );
     packet.validate()?;
 
-    // prepare message and set proper gas limit
+    // prepare ibc message
     let msg = IbcMsg::SendPacket {
         channel_id: msg.channel,
         data: to_binary(&packet)?,
--- a/contracts/cw20-ics20/src/contract.rs
+++ b/contracts/cw20-ics20/src/contract.rs
@@ -1,8 +1,8 @@
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::entry_point;
 use cosmwasm_std::{
-    from_binary, to_binary, Addr, Binary, Deps, DepsMut, Env, IbcMsg, IbcQuery, MessageInfo, Order,
-    PortIdResponse, Response, StdResult,
+    ensure_eq, from_binary, to_binary, Addr, Binary, Deps, DepsMut, Env, IbcMsg, IbcQuery,
+    MessageInfo, Order, PortIdResponse, Response, StdResult,
 };
 
 use cw2::{get_contract_version, set_contract_version};
@@ -135,13 +135,46 @@ pub fn execute_transfer(
     Ok(res)
 }
 
+/// The gov contract can allow new contracts, or increase the gas limit on existing contracts.
+/// It cannot block or reduce the limit to avoid forcible sticking tokens in the channel.
 pub fn execute_allow(
-    _deps: DepsMut,
+    deps: DepsMut,
     _env: Env,
-    _info: MessageInfo,
-    _allow: AllowMsg,
+    info: MessageInfo,
+    allow: AllowMsg,
 ) -> Result<Response, ContractError> {
-    unimplemented!()
+    let cfg = CONFIG.load(deps.storage)?;
+    ensure_eq!(info.sender, cfg.gov_contract, ContractError::Unauthorized);
+
+    let contract = deps.api.addr_validate(&allow.contract)?;
+    let set = AllowInfo {
+        gas_limit: allow.gas_limit,
+    };
+    ALLOW_LIST.update(deps.storage, &contract, |old| {
+        if let Some(old) = old {
+            // we must ensure it increases the limit
+            match (old.gas_limit, set.gas_limit) {
+                (None, Some(_)) => return Err(ContractError::CannotLowerGas),
+                (Some(old), Some(new)) if new < old => return Err(ContractError::CannotLowerGas),
+                _ => {}
+            };
+        }
+        Ok(AllowInfo {
+            gas_limit: allow.gas_limit,
+        })
+    })?;
+
+    let gas = if let Some(gas) = allow.gas_limit {
+        gas.to_string()
+    } else {
+        "None".to_string()
+    };
+
+    let res = Response::new()
+        .add_attribute("action", "allow")
+        .add_attribute("contract", allow.contract)
+        .add_attribute("gas_limit", gas);
+    Ok(res)
 }
 
 #[cfg_attr(not(feature = "library"), entry_point)]
--- a/contracts/cw20-ics20/src/contract.rs
+++ b/contracts/cw20-ics20/src/contract.rs
@@ -2,7 +2,7 @@
 use cosmwasm_std::entry_point;
 use cosmwasm_std::{
     ensure_eq, from_binary, to_binary, Addr, Binary, Deps, DepsMut, Env, IbcMsg, IbcQuery,
-    MessageInfo, Order, PortIdResponse, Response, StdResult,
+    MessageInfo, Order, PortIdResponse, Response, StdResult, SubMsg,
 };
 
 use cw2::{get_contract_version, set_contract_version};
@@ -93,10 +93,20 @@ pub fn execute_transfer(
         return Err(ContractError::NoFunds {});
     }
     // ensure the requested channel is registered
-    // FIXME: add a .has method to map to make this faster
-    if CHANNEL_INFO.may_load(deps.storage, &msg.channel)?.is_none() {
+    if CHANNEL_INFO.has(deps.storage, &msg.channel) {
         return Err(ContractError::NoSuchChannel { id: msg.channel });
     }
+
+    // if cw20 token, ensure it is whitelisted, and use the registered gas limit
+    let gas_limit = if let Amount::Cw20(coin) = &amount {
+        let addr = deps.api.addr_validate(&coin.address)?;
+        let allow = ALLOW_LIST
+            .may_load(deps.storage, &addr)?
+            .ok_or(ContractError::NotOnAllowList)?;
+        allow.gas_limit
+    } else {
+        None
+    };
 
     // delta from user is in seconds
     let timeout_delta = match msg.timeout {
@@ -115,19 +125,20 @@ pub fn execute_transfer(
     );
     packet.validate()?;
 
-    // prepare message
-    let msg = IbcMsg::SendPacket {
+    // prepare message and set proper gas limit
+    let mut msg = SubMsg::new(IbcMsg::SendPacket {
         channel_id: msg.channel,
         data: to_binary(&packet)?,
         timeout: timeout.into(),
-    };
+    });
+    msg.gas_limit = gas_limit;
 
     // Note: we update local state when we get ack - do not count this transfer towards anything until acked
     // similar event messages like ibctransfer module
 
     // send response
     let res = Response::new()
-        .add_message(msg)
+        .add_submessage(msg)
         .add_attribute("action", "transfer")
         .add_attribute("sender", &packet.sender)
         .add_attribute("receiver", &packet.receiver)
--- a/contracts/cw20-ics20/src/contract.rs
+++ b/contracts/cw20-ics20/src/contract.rs
@@ -12,10 +12,10 @@ use crate::amount::Amount;
 use crate::error::ContractError;
 use crate::ibc::Ics20Packet;
 use crate::msg::{
-    ChannelResponse, ExecuteMsg, InitMsg, ListChannelsResponse, MigrateMsg, PortResponse, QueryMsg,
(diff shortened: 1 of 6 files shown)


contracts/cw20-ics20/src/contract.rs
```diff
@@ -103,7 +122,7 @@ pub fn execute_transfer(
     );
     packet.validate()?;
 
-    // prepare message
+    // prepare message and set proper gas limit
```
hashedone, line 125: I see the comment about setting gas, but I don't see single line of code changed. Is it just encapsulated in the `packet`? Kind of guess, but I want to understand the comment change.

hashedone (commented):
I don't see two things:
* Setting the gas limit on message execution (it is stored in config, but I don't see where is it applied)
* Test failing if set gas costs affects anything. I think it is doable in MT - create a fake contract which takes very long to exec (some infinite or very long loop of  state updates I guess) and then used it with the `Transfer`. Transfer should fail with with proper error message.

contracts/cw20-ics20/src/contract.rs
ethanfrey: you are right. I need to remove it.

I started adding gas limit here, then realized I need to add it when seeing the cw20 token (2 places in ibc.rs) not when making the ibc packet. It should be in ibc_packet_receive and in ibc_packet_failure

contracts/cw20-ics20/src/msg.rs
```diff
@@ -37,6 +49,7 @@ pub struct TransferMsg {
     pub timeout: Option<u64>,
 }
 
+// TODO: query config, query allow list
```
maurolacy, line 52: Remove this line.

contracts/cw20-ics20/src/contract.rs
hashedone: Ok, then just remove commend and it is fine. I accept the PR then as this is not a functional thing.

hashedone: Ok, then just remove commend and it is fine. I accept the PR then as this is not a functional thing.

contracts/cw20-ics20/src/ibc.rs
```diff
@@ -243,12 +220,41 @@ fn do_ibc_packet_receive(deps: DepsMut, packet: &IbcPacket) -> Result<Ics20Packe
...
+    match amount {
+        Amount::Cw20(coin) => {
+            // if cw20 token, use the registered gas limit, or error if not whitelisted
+            let addr = deps.api.addr_validate(&coin.address)?;
+            Ok(ALLOW_LIST
+                .may_load(deps.storage, &addr)?
+                .ok_or(ContractError::NotOnAllowList)?
+                .gas_limit)
+        }
+        _ => Ok(None),
```
maurolacy, line 256: What about these? Just curious.

hashedone (approved):
lgtm

contracts/cw20-ics20/src/ibc.rs
```diff
@@ -348,7 +355,9 @@ fn send_amount(amount: Amount, recipient: String) -> SubMsg {
                 msg: to_binary(&msg).unwrap(),
                 funds: vec![],
             };
-            SubMsg::reply_on_error(exec, SEND_TOKEN_ID)
+            let mut sub = SubMsg::reply_on_error(exec, SEND_TOKEN_ID);
+            sub.gas_limit = gas_limit;
```
maurolacy, line 359: 👍🏼

contracts/cw20-ics20/src/ibc.rs
```diff
@@ -399,7 +413,9 @@ mod test {
             msg: to_binary(&msg).unwrap(),
             funds: vec![],
         };
-        SubMsg::reply_on_error(exec, SEND_TOKEN_ID)
+        let mut msg = SubMsg::reply_on_error(exec, SEND_TOKEN_ID);
```
maurolacy, line 416: Suggestion: Leet's add a `gas_limit` param to the `reply_on` fns for ergonomics.

contracts/cw20-ics20/src/contract.rs
```diff
@@ -124,6 +143,48 @@ pub fn execute_transfer(
...
+
+    let contract = deps.api.addr_validate(&allow.contract)?;
+    let set = AllowInfo {
+        gas_limit: allow.gas_limit,
+    };
+    ALLOW_LIST.update(deps.storage, &contract, |old| {
+        if let Some(old) = old {
+            // we must ensure it increases the limit
+            match (old.gas_limit, set.gas_limit) {
+                (None, Some(_)) => return Err(ContractError::CannotLowerGas),
```
ueco-jb, line 165: If I read this condition correctly - there was no gas limit, now it is - then:
> ensure it increases the limit

it looks like it increases the limit (from 0 to anything above)?

contracts/cw20-ics20/src/contract.rs
```diff
@@ -183,6 +249,59 @@ pub fn query_channel(deps: Deps, id: String) -> StdResult<ChannelResponse> {
     })
 }
 
+fn query_config(deps: Deps) -> StdResult<ConfigResponse> {
+    let cfg = CONFIG.load(deps.storage)?;
+    let res = ConfigResponse {
+        default_timeout: cfg.default_timeout,
+        gov_contract: cfg.gov_contract.into(),
+    };
```
ueco-jb, line 257: Why not just returning `Config` itself? Those are the same two only fields.

contracts/cw20-ics20/src/contract.rs
```diff
@@ -81,11 +93,18 @@ pub fn execute_transfer(
         return Err(ContractError::NoFunds {});
     }
     // ensure the requested channel is registered
-    // FIXME: add a .has method to map to make this faster
-    if CHANNEL_INFO.may_load(deps.storage, &msg.channel)?.is_none() {
+    if !CHANNEL_INFO.has(deps.storage, &msg.channel) {
```
maurolacy, line 96: Nice.

contracts/cw20-ics20/src/contract.rs
```diff
@@ -124,6 +143,48 @@ pub fn execute_transfer(
     Ok(res)
 }
 
+/// The gov contract can allow new contracts, or increase the gas limit on existing contracts.
+/// It cannot block or reduce the limit to avoid forcible sticking tokens in the channel.
```
maurolacy, line 147: Nice comment.

contracts/cw20-ics20/src/error.rs
```diff
@@ -49,6 +49,15 @@ pub enum ContractError {
 
     #[error("Got a submessage reply with unknown id: {id}")]
     UnknownReplyId { id: u64 },
+
+    #[error("You cannot lower the gas limit for a contract on the allow list")]
+    CannotLowerGas,
+
+    #[error("Only the governance contract can do this")]
+    Unauthorized,
```
ueco-jb, line 57: I'd incline to adding some `(String)` with context of operation which failed etc.

I stumbled upon many such errors and each time probably you need to look for answer in specific place in code to understand it
```rust
"Only the governance contract can do this: ", {context: contract update}
```

contracts/cw20-ics20/src/contract.rs
```diff
@@ -124,6 +143,48 @@ pub fn execute_transfer(
...
+                (Some(old), Some(new)) if new < old => return Err(ContractError::CannotLowerGas),
+                _ => {}
+            };
+        }
+        Ok(AllowInfo {
+            gas_limit: allow.gas_limit,
+        })
+    })?;
+
+    let gas = if let Some(gas) = allow.gas_limit {
```
maurolacy, line 175: Not really relevant, but `From<Option>>` for `String` for ergonomics?

maurolacy (approved):
LGTM.

contracts/cw20-ics20/src/contract.rs
ethanfrey: None means "infinite" - no limit
Some(x) means we apply a limit.

Probably should add a doc string here https://github.com/CosmWasm/cosmwasm/blob/main/packages/std/src/results/submessages.rs#L36

But seeing that None is the default, so one can imagine no limit is applied.

contracts/cw20-ics20/src/contract.rs
ethanfrey: Change from Addr to String.

I kind of got the feel we want to only use String in public APIs and Addr in stuff we have personally verified (in logic or in storage). But maybe this is me just being pedantic

contracts/cw20-ics20/src/ibc.rs
ethanfrey: We do, but it takes a non-option (u64) there.
Which makes sense usually when you know you want to set limit to eg. 100.000, but not here

contracts/cw20-ics20/src/contract.rs
ethanfrey: Sure. But how do I implement that without forking the stdlib (those are all standard types).
I guess I could pull this display logic into a function.

contracts/cw20-ics20/src/ibc.rs
ethanfrey: Native token transfers go to bank. No need to enforce limits.
That logic is trusted cosmos sdk code.

This is defines about calling into a cw20 token contract that might do ANYTHING, especially burning all gas in this transaction which would cause trouble with the relayer

ethanfrey (commented):
Thanks for the feedback everyone

contracts/cw20-ics20/src/contract.rs
ueco-jb: Okay, I actually though this is needed internally.

contracts/cw20-ics20/src/contract.rs
hashedone: I read it as `None` being "no-limit", so infinite limit, so any limitation is decreasing limit, so its illegal. "Zero" limit doesn't make any sense even remotely, so if `None` would be "zero" it could be just never be an optional as it is never a value you want to keep.

contracts/cw20-ics20/src/ibc.rs
maurolacy: Ah, I see there's a `with_gas_limit()`. Perhaps removing that and taking an `Option<u64>` `gas_limit` in these instead?

ethanfrey: It's in cosmwasm_std... 🤷‍♂️

giansalex:
Is possible to change denom from `cw20:contract` to `cw20:contract:denom`? 
keplr uses this [format](https://github.com/chainapsis/keplr-wallet/blob/c1976b522f610a9d894de7860dc1889486820bc8/packages/types/src/currency.ts#L19) , i am testing cw20-ics20 in osmosis-frontend and it works perfectly when i use their format .

```json
{
  "denom_traces": [
    {
      "path": "transfer/channel-2",
      "base_denom": "cw20:juno1jue5rlc9dkurt3etr57duutqu7prchqrk2mes2227m52kkrual3qa9lzdd:haz"
    }
  ]
}
```

- another question, is `gas_limit` used to prevent if Gov registers a malicious contract?

ethanfrey:
Fair suggestion, I can update the text format.
As long as the contract address is in there, it is safe.
BTW, most coins use uppercase denoms this is lower-case. Should I auto-lowercase it?

ethanfrey:
> another question, is gas_limit used to prevent if Gov registers a malicious contract?

So gov prevents intentionally malicious contracts that might try bitcoin mining til the tx runs out of gas, thus blocking the relayer from relaying packets.

However, imagine your cw20 token is a complex thing, like cETH from compound, which could call into many more contracts on a transfer. And imagine one of those was doing something bad. Adding a second limit here (beyond blanket whitelist) allows some more control. Like setting to 200.000 to start and be surprised if it ever hits that

giansalex:
> Fair suggestion, I can update the text format. As long as the contract address is in there, it is safe. BTW, most coins use uppercase denoms this is lower-case. Should I auto-lowercase it?

this is the [example contract ](https://github.com/giansalex/cw-plus/blob/ics20-updates/contracts/cw20-ics20/src/contract.rs), denom is a param in "RegisterCw20" msg, but can get the original denom.

ethanfrey:
@giansalex comments on closed PRs are hard to track and I usually just mark all closed PRs as read in my notifications.
Please make a message

PR #634, opened 2022-01-21
Fix tag consolidation for matching CHANGELOG entries

maurolacy:


ethanfrey (approved):
seen this before in another repo. :+1:

PR #635, opened 2022-01-25
Refactor cw2 spec readme

orkunkl:
I think removing migration reference cw2 spec makes sense. This spec is used for code inspection and verification too. Migration name in the title makes it hard for someone to find this.


ethanfrey (approved):
looks good.

PR #641, opened 2022-01-31
Allow modifying admin of Ics20 contract

ethanfrey:
- [x] Use ADMIN rather then config
- [x] Support standard ADMIN query and update
- [x] Add migration support from current version to new version

### DIFF
--- a/contracts/cw20-ics20/src/contract.rs
+++ b/contracts/cw20-ics20/src/contract.rs
@@ -1,8 +1,8 @@
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::entry_point;
 use cosmwasm_std::{
-    ensure_eq, from_binary, to_binary, Addr, Binary, Deps, DepsMut, Env, IbcMsg, IbcQuery,
-    MessageInfo, Order, PortIdResponse, Response, StdResult,
+    from_binary, to_binary, Addr, Binary, Deps, DepsMut, Env, IbcMsg, IbcQuery, MessageInfo, Order,
+    PortIdResponse, Response, StdResult,
 };
 
 use cw2::{get_contract_version, set_contract_version};
@@ -16,7 +16,7 @@ use crate::msg::{
     AllowMsg, AllowedInfo, AllowedResponse, ChannelResponse, ConfigResponse, ExecuteMsg, InitMsg,
     ListAllowedResponse, ListChannelsResponse, MigrateMsg, PortResponse, QueryMsg, TransferMsg,
 };
-use crate::state::{AllowInfo, Config, ALLOW_LIST, CHANNEL_INFO, CHANNEL_STATE, CONFIG};
+use crate::state::{AllowInfo, Config, ADMIN, ALLOW_LIST, CHANNEL_INFO, CHANNEL_STATE, CONFIG};
 use cw_utils::{maybe_addr, nonpayable, one_coin};
 
 // version info for migration info
@@ -25,7 +25,7 @@ const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
 
 #[cfg_attr(not(feature = "library"), entry_point)]
 pub fn instantiate(
-    deps: DepsMut,
+    mut deps: DepsMut,
     _env: Env,
     _info: MessageInfo,
     msg: InitMsg,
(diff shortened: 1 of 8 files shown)


contracts/cw20-ics20/src/contract.rs
```diff
@@ -186,14 +192,54 @@ pub fn execute_allow(
...
+    // existing one is newer
+    if storage_version > version {
+        return Err(ContractError::CannotMigrateVersion {
+            previous_version: stored.version,
+        });
+    }
+
+    // Then, run the proper migration
+    // unsupported old version
+    if storage_version < "0.11.1".parse().map_err(from_semver)? {
```
maurolacy, line 216: Use a `MIGRATE_MIN_VERSION` constant?

contracts/cw20-ics20/src/contract.rs
```diff
@@ -186,14 +192,54 @@ pub fn execute_allow(
...
+        return Err(ContractError::CannotMigrateVersion {
+            previous_version: stored.version,
+        });
+    }
+    // for 0.12.0-alpha1 or earlier, migrate from the config.gov_contract to ADMIN
+    if storage_version <= "0.12.0-alpha1".parse().map_err(from_semver)? {
```
maurolacy, line 222: Use a `MIGRATE_V1_VERSION` constant?

contracts/cw20-ics20/src/contract.rs
```diff
@@ -186,14 +192,54 @@ pub fn execute_allow(
...
+
+    if storage_version < version {
+        // we don't need to save anything if migrating from the same version
+        set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+    }
+
+    Ok(Response::new())
+}
+
+fn from_semver(err: semver::Error) -> StdError {
```
maurolacy, line 241: Nice.

contracts/cw20-ics20/src/contract.rs
```diff
@@ -251,9 +298,10 @@ pub fn query_channel(deps: Deps, id: String) -> StdResult<ChannelResponse> {
 
 fn query_config(deps: Deps) -> StdResult<ConfigResponse> {
     let cfg = CONFIG.load(deps.storage)?;
+    let admin = ADMIN.get(deps)?.unwrap_or_else(|| Addr::unchecked(""));
```
maurolacy, line 301: Nice. Wouldn't `"None"` be better here? I think we used that in the past.

contracts/cw20-ics20/src/contract.rs
```diff
@@ -186,14 +192,54 @@ pub fn execute_allow(
...
+    // unsupported old version
+    if storage_version < "0.11.1".parse().map_err(from_semver)? {
+        return Err(ContractError::CannotMigrateVersion {
+            previous_version: stored.version,
+        });
+    }
+    // for 0.12.0-alpha1 or earlier, migrate from the config.gov_contract to ADMIN
+    if storage_version <= "0.12.0-alpha1".parse().map_err(from_semver)? {
+        // Do v1 migration
+        let old_config = crate::migrations::v1::CONFIG.load(deps.storage)?;
```
maurolacy, line 224: 👍🏼

contracts/cw20-ics20/src/contract.rs
```diff
@@ -186,14 +192,54 @@ pub fn execute_allow(
...
+    }
+    // otherwise no migration (yet)
+
+    if storage_version < version {
+        // we don't need to save anything if migrating from the same version
+        set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+    }
+
+    Ok(Response::new())
+}
```
maurolacy, line 239: Not for now, but I wonder how much of this logic can be turned into a helper function.

maurolacy (approved):
LGTM.

contracts/cw20-ics20/src/contract.rs
ethanfrey: In this contract, we don't allow None, so it doesn't map much to the ADMIN, I kind of force it

contracts/cw20-ics20/src/migrations.rs
```diff
@@ -0,0 +1,16 @@
+// v1 format is anything older than 0.12.0
+pub mod v1 {
+    use schemars::JsonSchema;
+    use serde::{Deserialize, Serialize};
+
+    use cosmwasm_std::Addr;
+    use cw_storage_plus::Item;
+
+    #[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
+    pub struct ConfigV1 {
```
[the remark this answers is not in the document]
ethanfrey: I will do the other naming

contracts/cw20-ics20/src/contract.rs
ethanfrey: some of this is in poe-contracts/tg-utils

but matching on versions would be a good thing to abstract out (later, when we have some good examples/experience)

PR #642, opened 2022-01-31
Set version: 0.12.0-alpha2

ethanfrey:
Using these alphas for the ics20 contract. But also, this exposes the new Bounds type if we want to ensure other repos can update well before making a final tag


maurolacy (approved):
LGTM.

PR #643, opened 2022-01-31
Fix status/execution bugs in flex-multisig

### LINKED ISSUE #630: [cw3-flex/fixed-multisig] Status changes after voting and proposal expiration
A number of bugs were discovered in voting contracts, part of ~~soon to be~~ open-sourced poe-contracts. (See [#12](https://github.com/confio/poe-contracts/issues/12), [#14](https://github.com/confio/poe-contracts/issues/14), and [#16](https://github.com/confio/poe-contracts/issues/16)).

Let's backport / adapt the fixes and tests to cw3-flex-multisig and cw3-fixed-multisig (if applicable).

uint:
Closes #630, ports tests/fixes from https://github.com/confio/poe-contracts.

https://github.com/confio/poe-contracts/issues/12 was not an issue here, but I added a test for it.

https://github.com/confio/poe-contracts/issues/14 is indeed a bug here, fixing in this PR.

https://github.com/confio/poe-contracts/issues/16 was a false alarm in the first place.

### DIFF
--- a/contracts/cw3-flex-multisig/src/contract.rs
+++ b/contracts/cw3-flex-multisig/src/contract.rs
@@ -1146,12 +1146,7 @@ mod tests {
 
         // Trying to execute something that was already executed fails
         let err = app
-            .execute_contract(
-                Addr::unchecked(SOMEBODY),
-                flex_addr.clone(),
-                &execution,
-                &[],
-            )
+            .execute_contract(Addr::unchecked(SOMEBODY), flex_addr, &execution, &[])
             .unwrap_err();
         assert_eq!(
             ContractError::WrongExecuteStatus {},
@@ -1225,7 +1220,7 @@ mod tests {
         let res = app
             .execute_contract(
                 Addr::unchecked(SOMEBODY),
-                flex_addr.clone(),
+                flex_addr,
                 &ExecuteMsg::Execute { proposal_id },
                 &[],
             )
--- a/contracts/cw3-flex-multisig/src/contract.rs
+++ b/contracts/cw3-flex-multisig/src/contract.rs
@@ -188,7 +188,7 @@ pub fn execute_vote(
 
 pub fn execute_execute(
     deps: DepsMut,
-    _env: Env,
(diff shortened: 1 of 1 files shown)


uint:
The failing test exposes the bug: https://github.com/confio/poe-contracts/issues/14

Now for the fix...

maurolacy (commented):
Looks good.

Same comments than in https://github.com/confio/poe-contracts/pull/60 apply:
- Why not use `update_status()` in execute (mutable) calls everywhere?

Perhaps take the opportunity and fix the other issues with state here as well. Or, in a follow-up, your call.

uint:
> Looks good.
> 
> Same comments than in [confio/poe-contracts#60](https://github.com/confio/poe-contracts/pull/60) apply:
> 
>     * Why not use `update_status()` in execute (mutable) calls everywhere, and leave `current_status()` just for queries?

As in here? https://github.com/CosmWasm/cw-plus/pull/643/files#diff-f395b2761fa612ba44398ada5f0a4d9a9fc7333ce4565281fafd99329c4dcb23R200-R206

I don't think there's a point using `update_status()` if it's changed to `Execute` in the next line anyway. But we could add something like `mark_executed()` to the `Proposal` type if you think that's clearer.

> Perhaps take the opportunity and fix the other issues with state here as well. Or, in a follow-up. Your call.

Nah, I'd rather see another PR for that.

maurolacy:
> I don't think there's a point using `update_status()` if it's changed to `Executed` in the next line anyway.

I agree. In fact, `update_status` makes sense only in a limited number of cases, as the status changes will be discarded in case of error.

> But we could add something like `mark_executed()` to the `Proposal` type if you think that's clearer.

I think that will be good, but that it will better to do it in another iteration.

> 
> > Perhaps take the opportunity and fix the other issues with state here as well. Or, in a follow-up. Your call.
> 
> Nah, I'd rather see another PR for that.

OK.

maurolacy (approved):

PR #646, opened 2022-02-02
Use ContractInfoResponse from cosmwasm_std

webmaster128:
If I understand correctly, "due to non_exhaustive" really means _we have no way to construct it_. This is a feature initially because ContractInfoResponse instances are always provided to the contract developer and it should be hard to create them.

In https://github.com/CosmWasm/cosmwasm/pull/1177 a constructor was added sets the two fields which have no default and use default values for the rest. This can be used now to remove the copied type.

### DIFF
--- a/packages/multi-test/src/wasm.rs
+++ b/packages/multi-test/src/wasm.rs
@@ -4,9 +4,9 @@ use std::ops::Deref;
 
 use cosmwasm_std::{
     to_binary, Addr, Api, Attribute, BankMsg, Binary, BlockInfo, Coin, ContractInfo,
-    ContractResult, CustomQuery, Deps, DepsMut, Env, Event, MessageInfo, Order, Querier,
-    QuerierWrapper, Reply, ReplyOn, Response, StdResult, Storage, SubMsg, SubMsgExecutionResponse,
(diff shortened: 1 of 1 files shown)


packages/multi-test/src/wasm.rs
```diff
@@ -25,20 +25,6 @@ use cosmwasm_std::testing::mock_wasmd_attr;
 
 use anyhow::{bail, Context, Result as AnyResult};
 
-// TODO: we should import this from cosmwasm-std, but cannot due to non_exhaustive so copy here
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
-pub struct ContractInfoResponse {
```
webmaster128, line 30: This `pub` was never used, right? At least I did not find something.

ethanfrey: It was not. Maybe pub for test cases?

ethanfrey (approved):
Good cleanup now we updated the code.

Thanks.

packages/multi-test/src/wasm.rs
webmaster128: Probably just copy/paste from the cosmwasm repo

PR #650, opened 2022-02-07
add paging options struct

wrjanan:
This PR is created with reference to https://github.com/CosmWasm/cw-nfts/pull/50.

Linter complaining of too many variables in function call. 
Hence grouping paging options together seems like the natural cause of action.

Adding to cw-plus utils, since paging options might be a struct that any cw-storage structs can use.
Considering that the `skip` and `take` functions are useful and implemented within query functions.


ueco-jb:
This would require a new PR anyway.

PR #652, opened 2022-02-08
packages: support custom queries

uint:
In order to resolve https://github.com/confio/poe-contracts/issues/36, it looks like we need the `tgrade-valset` contract to use `TgradeQuery` as the custom query type for `Deps` and `DepsMut` (currently it's `Empty`).

In order to do that, `Admin` needs to support custom query types that are not `Empty` in `Deps`/`DepsMut`.

### DIFF
--- a/packages/controllers/src/admin.rs
+++ b/packages/controllers/src/admin.rs
@@ -3,7 +3,9 @@ use serde::{Deserialize, Serialize};
 use std::fmt;
 use thiserror::Error;
 
-use cosmwasm_std::{attr, Addr, Deps, DepsMut, MessageInfo, Response, StdError, StdResult};
+use cosmwasm_std::{
+    attr, Addr, CustomQuery, Deps, DepsMut, MessageInfo, Response, StdError, StdResult,
+};
 use cw_storage_plus::Item;
 
 // TODO: should the return values end up in utils, so eg. cw4 can import them as well as this module?
@@ -32,17 +34,17 @@ impl<'a> Admin<'a> {
         Admin(Item::new(namespace))
     }
 
-    pub fn set(&self, deps: DepsMut, admin: Option<Addr>) -> StdResult<()> {
+    pub fn set<Q: CustomQuery>(&self, deps: DepsMut<Q>, admin: Option<Addr>) -> StdResult<()> {
         self.0.save(deps.storage, &admin)
     }
 
-    pub fn get(&self, deps: Deps) -> StdResult<Option<Addr>> {
+    pub fn get<Q: CustomQuery>(&self, deps: Deps<Q>) -> StdResult<Option<Addr>> {
         self.0.load(deps.storage)
     }
 
     /// Returns Ok(true) if this is an admin, Ok(false) if not and an Error if
     /// we hit an error with Api or Storage usage
-    pub fn is_admin(&self, deps: Deps, caller: &Addr) -> StdResult<bool> {
+    pub fn is_admin<Q: CustomQuery>(&self, deps: Deps<Q>, caller: &Addr) -> StdResult<bool> {
         match self.0.load(deps.storage)? {
             Some(owner) => Ok(caller == &owner),
             None => Ok(false),
@@ -51,7 +53,11 @@ impl<'a> Admin<'a> {
 
     /// Like is_admin but returns AdminError::NotAdmin if not admin.
     /// Helper for a nice one-line auth check.
-    pub fn assert_admin(&self, deps: Deps, caller: &Addr) -> Result<(), AdminError> {
+    pub fn assert_admin<Q: CustomQuery>(
+        &self,
+        deps: Deps<Q>,
+        caller: &Addr,
+    ) -> Result<(), AdminError> {
         if !self.is_admin(deps, caller)? {
             Err(AdminError::NotAdmin {})
         } else {
@@ -59,9 +65,9 @@ impl<'a> Admin<'a> {
         }
     }
 
-    pub fn execute_update_admin<C>(
+    pub fn execute_update_admin<C, Q: CustomQuery>(
         &self,
-        deps: DepsMut,
+        deps: DepsMut<Q>,
         info: MessageInfo,
         new_admin: Option<Addr>,
     ) -> Result<Response<C>, AdminError>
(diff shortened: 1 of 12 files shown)


ethanfrey:
Ah, yes, now that I made Deps/DepsMut generic over the custom query type, we need to make controllers generic over that as well, not just Response<T>.

I agree. Can you also update other controllers? (Possibly as a follow up)

ethanfrey (approved):
nice find

ethanfrey:
Note that main is breaking, so after merge, you should make a new branch off of v0.11.1, back port this fix (and for the other controllers) and release a v0.11.2.

I don't think v0.12.0 is coming out so fast, so let's just patch the last stable version

uint:
> I agree. Can you also update other controllers? (Possibly as a follow up)

Happy to do that, should be easy enough.

maurolacy:
OK with me. Only thing is that v0.12 will include type safe bounds, and it will require code migration.

Migration should be easy, though. And, there's already a MIGRATING entry for this.

maurolacy (approved):
LGTM.

maurolacy:
As @ethanfrey says, you can either cherry pick this into a `0.11` branch (branched off from the latest 0.11 version tag) and release a new 0.11, or release 0.12 early. I for one would do the 2nd, even when it'll require some code migration on the client repo.

ethanfrey:
Speaking of.. let me tag the v0.12.0-alpha2 release I prepared.
And then I will make one more PR I would like in a final v0.12.0. I will try to finish tonight, be able to review/merge tomorrow

maurolacy:
Great. I can help with releasing 0.12 final if needed.

uint:
I realized we might also need to update multitest with custom queries and I'm taking a shot at that, but it's not trivial.

ethanfrey:
This is the relevant PR: https://github.com/CosmWasm/cw-plus/pull/653

I would like that in before a v0.12.0 is tagged

uint:
Okay, I tried to cover all of the `/packages`, `multi-test` included. Hopefully I didn't miss anything. Can you please take another look?

uint:
> As @ethanfrey says, you can either cherry pick this into a 0.11 branch (branched off from the latest 0.11 version tag) and release a new 0.11, or release 0.12 early. I for one would go for the 2nd, even when it'll require some code migration on the client repo(s).

I'd rather cut the new release than backport things. I guess for the migration there's nothing to do on my side if there's already a `MIGRATING.md` entry? Just that this is breaking for users?

packages/multi-test/src/lib.rs
```diff
@@ -6,6 +6,8 @@
 //!
 //! To understand the design of this module, please refer to `../DESIGN.md`
 
+#![allow(clippy::type_complexity)]
```
maurolacy, line 9: I understand this is needed on all three modules below, right?

packages/controllers/src/admin.rs
```diff
@@ -32,17 +34,17 @@ impl<'a> Admin<'a> {
         Admin(Item::new(namespace))
     }
 
-    pub fn set(&self, deps: DepsMut, admin: Option<Addr>) -> StdResult<()> {
+    pub fn set<Q: CustomQuery>(&self, deps: DepsMut<Q>, admin: Option<Addr>) -> StdResult<()> {
```
maurolacy, line 37: Why not defining `Q` over the struct? That way it can have a default of `Empty`.

You would require a `PhantomData`, but it's probably worth it.

The same with the other structs.

maurolacy: It seems doing that is more work than I anticipated... your call.

packages/multi-test/src/contracts.rs
```diff
@@ -4,51 +4,54 @@ use std::error::Error;
...
-    from_slice, Binary, CosmosMsg, Deps, DepsMut, Empty, Env, MessageInfo, Reply, Response, SubMsg,
+    from_slice, Binary, CosmosMsg, CustomQuery, Deps, DepsMut, Empty, Env, MessageInfo, Reply,
+    Response, SubMsg,
 };
 
 use anyhow::{anyhow, bail, Context, Result as AnyResult};
 
 /// Interface to call into a Contract
-pub trait Contract<T>
+pub trait Contract<T, Q = Empty>
```
maurolacy, line 14: 👍🏼

packages/multi-test/src/contracts.rs
```diff
@@ -130,8 +136,8 @@ where
     }
 }
 
-impl<T1, T2, T3, E1, E2, E3, C, T4, E4, E5, T6, E6>
-    ContractWrapper<T1, T2, T3, E1, E2, E3, C, T4, E4, E5, T6, E6>
+impl<T1, T2, T3, E1, E2, E3, C, Q, T4, E4, E5, T6, E6>
+    ContractWrapper<T1, T2, T3, E1, E2, E3, C, Q, T4, E4, E5, T6, E6>
```
maurolacy, line 140: Type complexity anyone?

packages/multi-test/src/lib.rs
uint: No, just `contracts`. We can make it more granular if we want to. The lint seems like it's sometimes broken though:

https://github.com/rust-lang/rust-clippy/issues/1013#issuecomment-587054810

packages/controllers/src/admin.rs
uint: I'll take a look!

maurolacy (approved):
LGTM.

packages/multi-test/src/lib.rs
maurolacy: Let's do that if possible, yes. That will give us the opportunity to at least flag upcoming and take a look at future type complexities.

packages/controllers/src/admin.rs
uint: I wonder about this. From another point of view, I feel like `Admin` shouldn't be concerned with what the custom query type is.

And I'm not sure the default of `Empty` is such a big benefit - in typical calls of these functions type inference should do the trick without the need to write something like `set::<TgradeQuery>(...)`... I think?

uint: For example, this line still works just fine after adding the type parameter to `set`:

https://github.com/CosmWasm/cw-plus/blob/controllers-custom-query-type/contracts/cw20-ics20/src/contract.rs#L42

uint: Yeah, I'm keeping this as is.

packages/multi-test/src/contracts.rs
uint: Are you suggesting simplifying this somehow? I'd definitely like to, I just have no idea how as of right now.

packages/multi-test/src/lib.rs
uint: That's fair, done.

packages/multi-test/src/contracts.rs
maurolacy: No. Just couldn't help noticing the sheer number of type params.

uint: Ah, yeah, it's... something :grin:

PR #653, opened 2022-02-08
Ics20 same ack handling as ibctransfer

ethanfrey:
Use the same handling as ibctransfer, where we update local balancer counter when sending, rather than wait for a success ack.

This works for the cases where relayer are slow to relay success acks (or never do it). It also allows a much more reliable and quicker round trip of send, return without waiting for the acks to be relayed even in the normal case.

Same behavior as ibctransfer, I don't need to be more conservative than that.

### DIFF
--- a/contracts/cw20-ics20/src/ibc.rs
+++ b/contracts/cw20-ics20/src/ibc.rs
@@ -386,9 +386,11 @@ mod test {
     use super::*;
     use crate::test_helpers::*;
 
-    use crate::contract::query_channel;
-    use cosmwasm_std::testing::mock_env;
-    use cosmwasm_std::{coins, to_vec, IbcAcknowledgement, IbcEndpoint, IbcTimeout, Timestamp};
+    use crate::contract::{execute, query_channel};
+    use crate::msg::{ExecuteMsg, TransferMsg};
+    use cosmwasm_std::testing::{mock_env, mock_info};
+    use cosmwasm_std::{coins, to_vec, IbcEndpoint, IbcTimeout, Timestamp};
+    use cw20::Cw20ReceiveMsg;
 
     #[test]
     fn check_ack_json() {
@@ -447,6 +449,7 @@ mod test {
         )
     }
 
+    #[allow(dead_code)]
     fn mock_sent_packet(my_channel: &str, amount: u128, denom: &str, sender: &str) -> IbcPacket {
         let data = Ics20Packet {
             denom: denom.into(),
@@ -468,6 +471,7 @@ mod test {
             IbcTimeout::with_timestamp(Timestamp::from_seconds(1665321069)),
         )
     }
+
     fn mock_receive_packet(
         my_channel: &str,
         amount: u128,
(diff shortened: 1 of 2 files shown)


contracts/cw20-ics20/src/ibc.rs
```diff
@@ -338,6 +336,9 @@ fn on_packet_failure(
 ) -> Result<IbcBasicResponse, ContractError> {
     let msg: Ics20Packet = from_binary(&packet.data)?;
 
+    // undo the balance update
+    reduce_channel_balance(deps.storage, &packet.src.channel_id, &msg.denom, msg.amount)?;
```
maurolacy, line 340: My concern would be: What if, for whatever reason, the packet failure message never arrives?

Not sure if this can be abused, as the successful deposit on the other side is the one crediting the funds. But still, if the packet failure never arrives, the balance would be left in an inconsistent state, resulting in lost funds.

contracts/cw20-ics20/src/ibc.rs
```diff
@@ -324,9 +325,6 @@ fn on_packet_success(deps: DepsMut, packet: IbcPacket) -> Result<IbcBasicRespons
         attr("success", "true"),
     ];
 
-    // we have made a proper transfer, record this
```
maurolacy, line 327: It will be nice to have something like a double check here. If this arrives, we confirm the balance adjustment. If this "never" arrives, we reject it.

This would require something like a queue / pool of pending balance adjustment confirmations... also, what happens if the deposit is successful in the other side, but the success packet never arrives. Does IBC has something like a "guaranteed delivery" mechanism for these packets?

maurolacy (approved):
Approving on your behalf, as I don't know enough about IBC yet for doing a proper review here.

Left a couple of comments with my main concerns. My take is that, if there's no mechanism for retries / messages confirmation, there can be problems.

contracts/cw20-ics20/src/ibc.rs
ethanfrey: It should eventually arrive. If nothing out via "timeout".

If this never gets relayed, the senders tokens remained locked in the contract. Once the timeout arrives, they are paid back.

contracts/cw20-ics20/src/ibc.rs
ethanfrey: There is no guarantee they ever arrive if the relayers are down, but an ack will remain waiting on the remote queue until it is handled.

contracts/cw20-ics20/src/ibc.rs
```diff
@@ -32,15 +32,20 @@ pub struct Ics20Packet {
...
+const V2: u32 = 2;
+
 impl Ics20Packet {
     pub fn new<T: Into<String>>(amount: Uint128, denom: T, sender: &str, receiver: &str) -> Self {
         Ics20Packet {
             denom: denom.into(),
             amount,
             sender: sender.to_string(),
             receiver: receiver.to_string(),
+            v: Some(V2),
```
giansalex, line 48: I created an issue about this #662

contracts/cw20-ics20/src/ibc.rs
```diff
@@ -32,15 +32,20 @@ pub struct Ics20Packet {
     pub receiver: String,
     /// the sender address
     pub sender: String,
+    /// used only by us to control ack handling
+    pub v: Option<u32>,
```
giansalex, line 36: I created an issue about this #662

ethanfrey: Yeah, I found this and had to revert it.
Turns out ibctransfer packet uses a super-strict JSON parser that fails on any unknown fields.
It seems to me that this is undesired behavior, but that is another discussion.

I need a different format and migration step to do a new release of this, very correct. I was out for a while.

PR #655, opened 2022-02-14
Fix missing prefix bound types

maurolacy:
Fixes a couple of errors with missing bound annotations for `IndexedMap`s and `MultiIndex`.

### DIFF
--- a/packages/storage-plus/src/indexed_map.rs
+++ b/packages/storage-plus/src/indexed_map.rs
@@ -172,11 +172,11 @@ where
     K: PrimaryKey<'a>,
     I: IndexList<T>,
 {
-    pub fn sub_prefix(&self, p: K::SubPrefix) -> Prefix<K::SuperSuffix, T> {
+    pub fn sub_prefix(&self, p: K::SubPrefix) -> Prefix<K::SuperSuffix, T, K::SuperSuffix> {
         Prefix::new(self.pk_namespace, &p.prefix())
     }
 
-    pub fn prefix(&self, p: K::Prefix) -> Prefix<K::Suffix, T> {
+    pub fn prefix(&self, p: K::Prefix) -> Prefix<K::Suffix, T, K::Suffix> {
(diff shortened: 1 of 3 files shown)


packages/storage-plus/src/indexed_map.rs
```diff
@@ -172,11 +172,11 @@ where
     K: PrimaryKey<'a>,
     I: IndexList<T>,
 {
-    pub fn sub_prefix(&self, p: K::SubPrefix) -> Prefix<K::SuperSuffix, T> {
+    pub fn sub_prefix(&self, p: K::SubPrefix) -> Prefix<K::SuperSuffix, T, K::SuperSuffix> {
```
ueco-jb, line 175: Ugh, so what is 3rd type argument in `Prefix` again?
```rust
pub struct Prefix<K, T, B = Vec<u8>>
where
    K: KeyDeserialize,
{
    storage_prefix: Vec<u8>,
    data: PhantomData<(T, B)>,
    pk_name: Vec<u8>,
    de_fn_kv: DeserializeKvFn<K, T>,
    de_fn_v: DeserializeVFn<T>,
}
```
It's the actual content of prefix, right?

maurolacy: The 3rd type is precisely the new `Bound` type annotations. You can compare with `Map` to get the idea.

While implementing this, made it work for `Map`, and then forgot about these here and elsewhere (`MultiIndex`). These bounds have a default of `Vec<u8>`, that's why it "worked".

Adding explicit tests for these typed bounds would be good I guess, at some point.

ueco-jb (approved):
> Adding explicit tests for these typed bounds would be good I guess, at some point.

I think this is worth an extra issue as follow up.

packages/storage-plus/src/indexed_map.rs
maurolacy: It's a little tricky how all this works. I can give a small overview / talk at some point, if there's interest.

PR #656, opened 2022-02-14
Forward original errors in multitest `instantiate`, `execute` and `query`

maurolacy:
Instead of rewriting errors, just forward them so that they can be used for error matching / patterns.

### DIFF
--- a/packages/multi-test/src/app.rs
+++ b/packages/multi-test/src/app.rs
@@ -2604,9 +2604,9 @@ mod test {
                 panic!("wrong StdError variant");
             }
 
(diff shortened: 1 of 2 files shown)


ueco-jb (approved):
You basically get rid of annotation "Contract returned an error..." or is something more sophisticated going on here?

maurolacy:
> You basically get rid of annotation "Contract returned an error..." or is something more sophisticated going on here?

Just that. If not, we cannot pattern / text match the original errors anymore; which is useful for tests.

PR #659, opened 2022-02-15
cw20-base: validate addresses are unique in initial balances

### LINKED ISSUE #626: Duplicate accounts in cw20 initial balances causes unrecoverable inconsistent state
If `initial_balances` contains one or more duplicate accounts, these accounts will have the balance of the final balance saved to state, whilst the `total_supply` will be the sum over all initial balances provided.

https://github.com/CosmWasm/cw-plus/blob/501efe8fb5151f658380a87c30091a2851e556e5/contracts/cw20-base/src/contract.rs#L154-L162

Example:

```rust
#[test]
fn inconsistent_total_supply_with_duplicate_accounts_in_initial_balances() {
    let mut deps = mock_dependencies(&[]);
    let instantiate_msg = InstantiateMsg {
        name: "Cash Token".to_string(),
        symbol: "CASH".to_string(),
        decimals: 9,
        initial_balances: vec![
            Cw20Coin {
                address: String::from("addr0000"),
                amount: Uint128::from(1u128),
            },
            Cw20Coin {
                address: String::from("addr0000"),
                amount: Uint128::from(1u128),
            },
        ],
        mint: None,
        marketing: None,
    };
    let info = mock_info("creator", &[]);
    let env = mock_env();
    instantiate(deps.as_mut(), env, info, instantiate_msg).unwrap();

    let token_info = query_token_info(deps.as_ref()).unwrap();
    assert_eq!(token_info.total_supply, Uint128::new(2));
    assert_eq!(get_balance(deps.as_ref(), "addr0000"), Uint128::new(1));
}
```

Should we assert that all accounts in `initial_balances` are unique? Happy to make a PR for this if there is interest.

NB this was found in an audit of Mars protocol in a contract that was forked from the cw-plus cw20-base contract.

harryscholes:
Fixes #626

### DIFF
--- a/contracts/cw20-base/src/contract.rs
+++ b/contracts/cw20-base/src/contract.rs
@@ -151,14 +151,32 @@ pub fn instantiate(
     Ok(Response::default())
 }
 
-pub fn create_accounts(deps: &mut DepsMut, accounts: &[Cw20Coin]) -> StdResult<Uint128> {
+pub fn create_accounts(
+    deps: &mut DepsMut,
+    accounts: &[Cw20Coin],
+) -> Result<Uint128, ContractError> {
+    validate_accounts(accounts)?;
+
     let mut total_supply = Uint128::zero();
     for row in accounts {
         let address = deps.api.addr_validate(&row.address)?;
         BALANCES.save(deps.storage, &address, &row.amount)?;
         total_supply += row.amount;
     }
+
     Ok(total_supply)
+}
+
+pub fn validate_accounts(accounts: &[Cw20Coin]) -> Result<(), ContractError> {
+    let mut addresses = accounts.iter().map(|c| &c.address).collect::<Vec<_>>();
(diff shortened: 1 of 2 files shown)


contracts/cw20-base/src/contract.rs
```diff
@@ -151,16 +151,34 @@ pub fn instantiate(
...
         total_supply += row.amount;
     }
+
     Ok(total_supply)
 }
 
+pub fn validate_accounts(accounts: &[Cw20Coin]) -> Result<(), ContractError> {
+    let mut addresses = accounts.iter().map(|c| &c.address).collect::<Vec<_>>();
+    addresses.sort();
+    addresses.dedup();
```
ethanfrey, line 173: Seems like a nice way to calculate this.

ethanfrey (approved):
Nice one

PR #660, opened 2022-02-16
Properly handle generic queries in multi-test

ethanfrey:
Make ContractWrapper::new_with_empty convert Deps<Empty> to Deps<Q>

### DIFF
--- a/packages/multi-test/src/contracts.rs
+++ b/packages/multi-test/src/contracts.rs
@@ -269,11 +269,14 @@ where
     C: Clone + fmt::Debug + PartialEq + JsonSchema + 'static,
     Q: CustomQuery + DeserializeOwned + 'static,
 {
-    let customized =
-        move |mut deps: DepsMut<Q>, env: Env, info: MessageInfo, msg: T| -> Result<Response<C>, E> {
-            let deps = decustomize_deps_mut(&mut deps);
-            raw_fn(deps, env, info, msg).map(customize_response::<C>)
-        };
+    let customized = move |mut deps: DepsMut<Q>,
+                           env: Env,
+                           info: MessageInfo,
+                           msg: T|
+          -> Result<Response<C>, E> {
+        let deps = decustomize_deps_mut(&mut deps);
+        raw_fn(deps, env, info, msg).map(customize_response::<C>)
+    };
     Box::new(customized)
 }
 
--- a/packages/multi-test/src/contracts.rs
+++ b/packages/multi-test/src/contracts.rs
@@ -2,10 +2,11 @@ use schemars::JsonSchema;
 use serde::de::DeserializeOwned;
 use std::error::Error;
 use std::fmt::{self, Debug, Display};
+use std::ops::Deref;
 
 use cosmwasm_std::{
-    from_slice, Binary, CosmosMsg, CustomQuery, Deps, DepsMut, Empty, Env, MessageInfo, Reply,
-    Response, SubMsg,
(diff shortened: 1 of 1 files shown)


ethanfrey:
@maurolacy this is a start.
fails due to lifetime issue that may require a change  to cosmwasm-std

maurolacy:
Will give it a try tomorrow. Ended up avoiding the need for this by removing the custom message from `tg4-group` entirely. See https://github.com/confio/poe-contracts/pull/109.

maurolacy:
Trying this, but needs to do an upgrade of dependencies / patches locally to properly assess it works.

That said, I think this is a good idea, and we should merge this, after confirming it works, or making the needed changes.

ueco-jb (approved):
LGTM
I think this is a way to go.
I only wonder, is this change breaking? From generic type to concrete (`Empty`) one?

uint (approved):
Seems good enough!

ethanfrey:
> I only wonder, is this change breaking? From generic type to concrete (Empty) one?

In theory, but everyone using `Response<Empty>` is also using `Deps<Empty>`
If there are some using `Response<T>` with `Deps<Empty>` or `Response<Empty>` with `Deps<T>` that will be breaking.

Maybe we can add two extra functions for those two (less likely cases). I am sure they will pop up eventually

ethanfrey:
> Anyway: I have a strong feeling that the message type (custom or empty) should be orthogonal to the multi-test blockchain / app. And that we are running into all these clashes, in the end, because it isn't.

It would be much easier if they were. But I think they cannot be.

Basically, we need to ensure the Response returned is a subset of the possible responses managed by the app.
And the Querier needed is a subset of the Querier provided.

The other option is just use `Vec<u8>` everywhere, but we couldn't even parse the Response from a contract without knowing the Custom type

ethanfrey:
Can you make a PR to the case you mention that *almost* works and I will look into it?

ethanfrey:
I would like to:

1. Improve multi-test generics enough to make this work well for now.
2. Later, make multi-test more robust / easy to use

maurolacy:
> Can you make a PR to the case you mention that _almost_ works and I will look into it?

There you go: https://github.com/confio/tfi/pull/68

maurolacy:
> > Anyway: I have a strong feeling that the message type (custom or empty) should be orthogonal to the multi-test blockchain / app. And that we are running into all these clashes, in the end, because it isn't.
> 
> It would be much easier if they were. But I think they cannot be.
> 
> Basically, we need to ensure the Response returned is a subset of the possible responses managed by the app. And the Querier needed is a subset of the Querier provided.
> 
> The other option is just use `Vec<u8>` everywhere, but we couldn't even parse the Response from a contract without knowing the Custom type

Can't we define the message type at the calls / `store_code`, etc. levels? Instead of having everything pre-defined as `Empty` or some custom type at the `App` level.

It's a matter of decoupling the message type from the app. I understand it will require some major overhauling, though.

ethanfrey:
The issue is the App has a custom module (not contract) that can handle custom messages C and queries Q.

The contracts must all return C or Empty so they can be dispatched.

maurolacy:
I see. Some thoughts or ideas that come to mind:

- Use serialisation / deserialisation tricks to handle / dispatch multiple message types internally. Everyone knows what to expect / receive, so it just optimistically tries to deserialise to it.
- Instead of a `Custom(T)` enum, use the / a `CustomMessage` trait to abstract away different message types, and `dyn` our way around it. Not sure we can have something like a `Custom(dyn CustomMessage)`, but you get the idea.

These are just off the top of my head. I don't know how relevant / practical / possible / sensible they are.

Issue #665, opened 2022-02-21
[cw3-flex/fixed-multisig] Reject proposals early


maurolacy:
After a similar issue in dao-contracts: https://github.com/DA0-DA0/dao-contracts/issues/169.

The task is then backporting / adapting that logic here, and adding the equivalent functionality to fixed-multisig as well.

Callum-A:
My fix for DAO DAO is available [here](https://github.com/DA0-DA0/dao-contracts/pull/181). I am happy to undertake this myself or if anyone else wants to pick this up and cannot understand my solution, feel free to reach out!

maurolacy:
Please submit a PR, and we review it and merge it.

maurolacy:
Closed by #668.

PR #668, opened 2022-02-22
Reject proposals early

Callum-A:
Addresses Issue #665 

Implemented what I implemented for DAO DAO here for the multisigs. If I have missed any functionality or coverage do say.

- Proposals are now marked as rejected when no votes reach their threshold.
- Added tests for proposal struct and unit tests for functionality.

Feedback appreciated.

### DIFF
--- a/contracts/cw3-fixed-multisig/src/contract.rs
+++ b/contracts/cw3-fixed-multisig/src/contract.rs
@@ -778,7 +778,7 @@ mod tests {
             msgs,
             latest: None,
         };
-        let res = execute(deps.as_mut(), mock_env(), info.clone(), proposal).unwrap();
+        let res = execute(deps.as_mut(), mock_env(), info, proposal).unwrap();
 
         // Get the proposal id from the logs
         let proposal_id: u64 = res.attributes[2].value.parse().unwrap();
--- a/contracts/cw3-fixed-multisig/src/contract.rs
+++ b/contracts/cw3-fixed-multisig/src/contract.rs
@@ -764,6 +764,56 @@ mod tests {
         let info = mock_info(VOTER5, &[]);
         let err = execute(deps.as_mut(), mock_env(), info, yes_vote).unwrap_err();
         assert_eq!(err, ContractError::NotOpen {});
+
+        // Propose
+        let info = mock_info(OWNER, &[]);
+        let bank_msg = BankMsg::Send {
+            to_address: SOMEBODY.into(),
+            amount: vec![coin(1, "BTC")],
+        };
+        let msgs = vec![CosmosMsg::Bank(bank_msg)];
+        let proposal = ExecuteMsg::Propose {
+            title: "Pay somebody".to_string(),
+            description: "Do I pay her?".to_string(),
+            msgs,
+            latest: None,
+        };
+        let res = execute(deps.as_mut(), mock_env(), info.clone(), proposal).unwrap();
+
+        // Get the proposal id from the logs
+        let proposal_id: u64 = res.attributes[2].value.parse().unwrap();
+
+        // Cast a No vote
+        let no_vote = ExecuteMsg::Vote {
+            proposal_id,
+            vote: Vote::No,
+        };
+        // Voter1 vote no
+        let info = mock_info(VOTER1, &[]);
+        let res = execute(deps.as_mut(), mock_env(), info, no_vote.clone()).unwrap();
+
+        // Verify it is not enough to reject yet
+        assert_eq!(
+            res,
+            Response::new()
+                .add_attribute("action", "vote")
+                .add_attribute("sender", VOTER1)
+                .add_attribute("proposal_id", proposal_id.to_string())
+                .add_attribute("status", "Open")
+        );
+
+        let info = mock_info(VOTER4, &[]);
+        let res = execute(deps.as_mut(), mock_env(), info, no_vote).unwrap();
+
+        // Verify it is now rejected due to reaching threshold
+        assert_eq!(
+            res,
+            Response::new()
+                .add_attribute("action", "vote")
+                .add_attribute("sender", VOTER4)
+                .add_attribute("proposal_id", proposal_id.to_string())
+                .add_attribute("status", "Rejected")
+        );
     }
 
     #[test]
--- a/contracts/cw3-fixed-multisig/src/contract.rs
+++ b/contracts/cw3-fixed-multisig/src/contract.rs
@@ -788,7 +788,7 @@ mod tests {
             proposal_id,
             vote: Vote::No,
         };
-        // Voter1 vote no
+        // Voter1 vote no, weight 1
         let info = mock_info(VOTER1, &[]);
         let res = execute(deps.as_mut(), mock_env(), info, no_vote.clone()).unwrap();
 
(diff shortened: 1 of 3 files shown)


contracts/cw3-fixed-multisig/src/state.rs
```diff
@@ -47,6 +47,9 @@ impl Proposal {
         if status == Status::Open && self.expires.is_expired(block) {
             status = Status::Rejected;
         }
+        if status == Status::Open && self.is_rejected(block) {
+            status = Status::Rejected;
+        }
```
ueco-jb, line 52:
```suggestion
        if status == Status::Open && (self.expires.is_expired(block) || self.is_rejected(block)) {
            status = Status::Rejected;
        }
```

contracts/cw3-fixed-multisig/src/state.rs
```diff
@@ -76,18 +86,30 @@ impl Proposal {
...
                     let possible_opinions = self.total_weight - self.votes.abstain;
-                    self.votes.yes >= votes_needed(possible_opinions, threshold)
+                    vote_count >= votes_needed(possible_opinions, threshold)
                 }
             }
         }
     }
+
+    /// returns true iff this proposal is sure to pass (even before expiration if no future
+    /// sequence of possible votes can cause it to fail)
```
ueco-jb, line 103:
```suggestion
    /// returns true if this proposal is sure to pass (even before expiration, if no future
    /// sequence of possible votes could cause it to fail)
```

contracts/cw3-fixed-multisig/src/state.rs
```diff
@@ -76,18 +86,30 @@ impl Proposal {
...
+    pub fn is_passed(&self, block: &BlockInfo) -> bool {
+        self.does_vote_count_reach_threshold(self.votes.yes, block)
+    }
+
+    /// As above for the rejected check, used to check if a proposal is
+    /// already rejected.
```
ueco-jb, line 109: IMO doc comments shouldn't state "as above`, since "above" part won't be shown by IDE, rust compiled docs etc.

contracts/cw3-fixed-multisig/src/state.rs
```diff
@@ -57,17 +60,24 @@ impl Proposal {
...
-    pub fn is_passed(&self, block: &BlockInfo) -> bool {
+    /// Helper function to check if a certain vote count has reached threshold.
+    /// Only called from is_rejected and is_passed for no and yes votes
+    /// Handles the different threshold types accordingly.
+    /// This function returns true if and only if vote_count is greater than the threshold which
+    /// is calculated.
+    /// In the case where we use yes votes, this function will return true if and only if the
+    /// proposal will pass.
+    /// In the case where we use no votes, this function will return true if and only if the
+    /// proposal will be rejected regardless of other votes.
```
ueco-jb, line 71:
```suggestion
    /// This function returns true if and only if vote_count is greater than the threshold which
    /// is calculated.
    /// In the case where we use yes votes, this function will return true if the proposal will pass.
    /// In the case where we use no votes, this function will return true if the
    /// proposal will be rejected regardless of other votes.
```

ueco-jb (commented):
I'll be back to review actual implementation, commented only obvious stuff so far.

contracts/cw3-flex-multisig/src/contract.rs
```diff
@@ -1056,6 +1056,39 @@ mod tests {
...
+        // Cast a No vote
+        let no_vote = ExecuteMsg::Vote {
+            proposal_id,
+            vote: Vote::No,
+        };
+        let _ = app
+            .execute_contract(Addr::unchecked(VOTER2), flex_addr.clone(), &no_vote, &[])
+            .unwrap();
+
+        // Powerful voter supports it, so it rejects
```
maurolacy, line 1078:
```suggestion
        // Powerful voter opposes it, so it rejects
```

contracts/cw3-fixed-multisig/src/state.rs
```diff
@@ -76,18 +82,30 @@ impl Proposal {
...
     }
+
+    /// Returns true if this proposal is sure to pass (even before expiration, if no future
+    /// sequence of possible votes could cause it to fail).
+    pub fn is_passed(&self, block: &BlockInfo) -> bool {
+        self.does_vote_count_reach_threshold(self.votes.yes, block)
+    }
+
+    /// Returns true if this proposal is sure to be rejected (even before expiration, if
+    /// no future sequence of possible votes could cause it to fail).
```
maurolacy, line 105:
```suggestion
    /// no future sequence of possible votes could cause it to pass).
```

maurolacy (approved):
LGTM. Thanks for the contribution.

contracts/cw3-fixed-multisig/src/state.rs
```diff
@@ -57,17 +57,23 @@ impl Proposal {
...
+    /// is calculated.
+    /// In the case where we use yes votes, this function will return true if the proposal will pass.
+    /// In the case where we use no votes, this function will return true if the
+    /// proposal will be rejected regardless of other votes.
+    fn does_vote_count_reach_threshold(&self, vote_count: u64, block: &BlockInfo) -> bool {
         match self.threshold {
             Threshold::AbsoluteCount {
                 weight: weight_needed,
-            } => self.votes.yes >= weight_needed,
+            } => vote_count >= weight_needed,
```
ethanfrey, line 72: This check doesn't work properly of no votes. I would implement it sepately rather than combining.

Assume a threshold of 67%.

If we have 40% yes and 40% no, it would return false for both.
However, this could be safely marked as rejected.

Your logic works fine when the threshold is 50% (same for both).

I would propose not renaming the function and using it for both cases (as done here), but rather copying is_passed to is_rejected and make the various logic changes there needed (often 1-threshold or such). This is different for each branch and hard to generalise. Better to have two separate functions than combine them

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -764,6 +764,56 @@ mod tests {
...
+                .add_attribute("action", "vote")
+                .add_attribute("sender", VOTER1)
+                .add_attribute("proposal_id", proposal_id.to_string())
+                .add_attribute("status", "Open")
+        );
+
+        let info = mock_info(VOTER4, &[]);
+        let res = execute(deps.as_mut(), mock_env(), info, no_vote).unwrap();
+
+        // Verify it is now rejected due to reaching threshold
```
ethanfrey, line 808: This is actually a test showing the code is broken.

`setup_test_case` adds 6 voters with a total weight of 16.
This test cases constructs an example with the odd configuration of 3 votes to reach threshold.

4 no votes should not mark it as rejected. You should require 14 no votes to ensure it is rejected.

I would propose a separate test function to cover this.

contracts/cw3-fixed-multisig/src/state.rs
```diff
@@ -318,6 +404,95 @@ mod test {
         assert!(check_is_passed(quorum, passing, 16, false));
     }
 
+    #[test]
+    fn proposal_rejected_quorum() {
+        let quorum = Threshold::ThresholdQuorum {
+            threshold: Decimal::percent(50),
```
ethanfrey, line 410: Let's use threshold = Decimal::percent(60) so we can test the real case. Then fix the logic to match

ethanfrey (changes requested):
Is rejected doesn't work if the threshold deviates from 50%.

Added suggestions how to fix code and tests

contracts/cw3-fixed-multisig/src/state.rs
```diff
@@ -84,27 +78,53 @@ impl Proposal {
...
-
     /// Returns true if this proposal is sure to be rejected (even before expiration, if
     /// no future sequence of possible votes could cause it to pass).
     pub fn is_rejected(&self, block: &BlockInfo) -> bool {
-        self.does_vote_count_reach_threshold(self.votes.no, block)
+        match self.threshold {
+            Threshold::AbsoluteCount {
+                weight: weight_needed,
+            } => {
+                let weight = self.total_weight - weight_needed;
```
ethanfrey, line 99: This is good

contracts/cw3-fixed-multisig/src/state.rs
```diff
@@ -84,27 +78,53 @@ impl Proposal {
...
+                self.votes.no > weight
+            }
+            Threshold::AbsolutePercentage {
+                percentage: percentage_needed,
+            } => {
+                self.votes.no
+                    > votes_needed(
+                        self.total_weight - self.votes.abstain,
+                        Decimal::one() - percentage_needed,
```
ethanfrey, line 108: Yes, this is proper place. Nice eye

ethanfrey (approved):
Nice fix.
Just request two minor changes, which don't affect correctness

contracts/cw3-fixed-multisig/src/state.rs
```diff
@@ -464,12 +461,37 @@ mod test {
...
+        ));
+        // Can reject when expired
+        assert!(check_is_rejected(
+            quorum.clone(),
+            rejecting.clone(),
+            50,
+            true
+        ));
+
+        // Check edgecase where quorum is not met but we can reject
```
ethanfrey, line 480: 💯

ethanfrey (approved):
great work. merging

maurolacy:
> Is rejected doesn't work if the threshold deviates from 50%.

Good catch!

PR #671, opened 2022-02-23
Update storage plus docs / Remove `MultiIndex` PK default

### LINKED ISSUE #669: Fix `MultiIndex` last type param default / docs
With the latest implementation of type-safe bounds, the definition of a `MultiIndex` `PK` type is crucial, as it now serves to define the type of the (partial) bounds over the index key (the part that corresponds to the pk, that is). See https://github.com/confio/poe-contracts/pull/114/files#r812029191 for reference / an example.

Let's fix the docs to reflect this use case, and its importance.

Also, let's remove the default over `PK` that we introduced for backwards compatibility, as this type definition is now really important for specifying (type-safe) bounds.

maurolacy:
Closes #669.

### DIFF
--- a/packages/storage-plus/README.md
+++ b/packages/storage-plus/README.md
@@ -104,7 +104,7 @@ we support not only simple binary keys (`&[u8]`), but tuples, which are
 combined. This allows us to store allowances as composite keys
 eg. `(owner, spender)` to look up the balance.
 
-Beyond direct lookups, we have a super power not found in Ethereum -
+Beyond direct lookups, we have a super-power not found in Ethereum -
 iteration. That's right, you can list all items in a `Map`, or only
 part of them. We can efficiently allow pagination over these items as
 well, starting at the point the last query ended, with low gas costs.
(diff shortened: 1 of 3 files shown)


packages/storage-plus/README.md
```diff
@@ -211,22 +211,25 @@ legitimate address, and not random text which will fail later.
...
 
-Here how we use it with composite keys. Just define a tuple as a key and use that
+Here's how we use it with composite keys. Just define a tuple as a key and use that
 everywhere you used a byte slice above.
 
 ```rust
-// Note the tuple for primary key. We support one slice, or a 2 or 3-tuple
-// adding longer tuples is quite easy but unlikely to be needed.
+// Note the tuple for primary key. We support one slice, or a 2 or 3-tuple.
+// Adding longer tuples is possible, but unlikely to be needed.
```
ueco-jb, line 232: :) Points for honesty

packages/storage-plus/README.md
```diff
@@ -314,7 +317,7 @@ fn demo() -> StdResult<()> {
 
 In addition to getting one particular item out of a map, we can iterate over the map
 (or a subset of the map). This let us answer questions like "show me all tokens",
-and we provide some nice [`Bound`](#Bound) helpers to easily allow pagination or custom ranges.
+and we provide some nice [`Bound`](#bound) helpers to easily allow pagination, or custom ranges.
```
ueco-jb, line 320: Actually I think it should be without ','.

packages/storage-plus/README.md
```diff
@@ -323,7 +326,7 @@ over all items with `range(store, min, max, order)`. It supports `Order::Ascendi
 `min` is the lower bound and `max` is the higher bound.
 
 If the `min` and `max` bounds are `None`, `range` will return all items under the prefix. You can use `.take(n)` to
-limit the results to `n` items and start doing pagination. You can also set the `min` bound to
+limit the results to `n` items, and start doing pagination. You can also set the `min` bound to
```
ueco-jb, line 329: As above.

ueco-jb (approved):
lgtm
I salute you for your commitment, since this PR could be a one liner after all.

packages/storage-plus/README.md
maurolacy: :)

PR #675, opened 2022-03-09
Update changelog add upcoming

maurolacy:
Adds a `-u|--upcoming-tag <tag>` option to `update_changelog.sh`, so that a new entry is created in CHANGELOG.md for the upcoming version tag.

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -1,6 +1,17 @@
 #!/bin/bash
 set -o errexit -o nounset -o pipefail
-command -v shellcheck > /dev/null && shellcheck "$0"
+command -v shellcheck >/dev/null && shellcheck "$0"
+
+function print_usage() {
+  echo "Usage: $0 [-h|--help]"
+  echo "Publishes crates to crates.io."
+}
+
+if [ "$1" = "-h" ] || [ "$1" = "--help" ]
+then
+    print_usage
+    exit 1
+fi
 
 # this should really more to cosmwasm...
 STORAGE_PACKAGES="storage-plus"


scripts/update_changelog.sh
```diff
@@ -50,4 +65,11 @@ sed -i -n "/^## \\[${TAG}[^]]*\\]/,\$p" CHANGELOG.md
 
 github_changelog_generator -u CosmWasm -p cw-plus --base CHANGELOG.md $ORIGINAL_OPTS || cp /tmp/CHANGELOG.md.$$ CHANGELOG.md
 
+if [ -n "$UPCOMING_TAG" ]
+then
+  # Add "upcoming" version tag
+  TODAY=$(date "+%Y-%m-%d")
+  sed -i "s+\[Full Changelog\](https://github.com/CosmWasm/cw-plus/compare/\(.*\)\.\.\.HEAD)+[Full Changelog](https://github.com/CosmWasm/cw-plus/compare/$UPCOMING_TAG...HEAD)\n\n## [$UPCOMING_TAG](https://github.com/CosmWasm/cw-plus/tree/$UPCOMING_TAG) ($TODAY)\n\n[Full Changelog](https://github.com/CosmWasm/cw-plus/compare/\1...$UPCOMING_TAG)+" CHANGELOG.md
+fi
```
ueco-jb, line 73: Just an idea, you could try to add some validator here that sed actually worked (in case anything happens and for example changelog format slightly changes)

ueco-jb (approved):
lgtm

scripts/update_changelog.sh
maurolacy: Yes, that's an ugly line indeed. If I enable `shellcheck` here, which I tried, it complains about many issues with this script, and with abundant whining on this line in particular.

Will try to improve on it. Perhaps silencing some warnings.

PR #684, opened 2022-03-24
Fix cw20 ics20 packets

### LINKED ISSUE #662: ICS20, invalid packet data
Packet data sent:
```
{
   "amount":"125000",
   "denom":"ujunox",
   "receiver":"osmo14vhcdsyf83ngsrrqc92kmw8q9xakqjm0hq6xsa",
   "sender":"juno14vhcdsyf83ngsrrqc92kmw8q9xakqjm0ff2dpn",
   "v":2
}
```
Ack: [40906A62...8D1BB9E7](https://lcd.juno.giansalex.dev/cosmos/tx/v1beta1/txs/40906A62DCCEEB0114F0B4FFA4C924AB37602E2E473CC6DB0E8FC98A8D1BB9E7)
```
{"error":"cannot unmarshal ICS-20 transfer packet data"}
```

ibc-go/transfer doesn't allow unknown fields, i needed a [custom UnmarshalJSON](https://github.com/disperze/ibc-osmo/blob/206ddbf7f2e4d5213111b788a606e6b2e8ccbf7b/x/intergamm/types/codec.go#L31)

ethanfrey:
Closes #662 

We change the logic to always be the "v2" style, and no longer add a flag in the packet (which is rejected by the transfer module), but rather migrate current state to new format if possible (for 1 open channel, we can assign all current balance to inflight for this channel). 

If there are production versions that have multiple open channels and need migration, please comment here, and we could possibly support that migration in the future (but it is much trickier).

cc: @giansalex

TODO: add a test for the migration code

### DIFF
--- a/contracts/cw20-ics20/src/migrations.rs
+++ b/contracts/cw20-ics20/src/migrations.rs
@@ -17,11 +17,13 @@ pub mod v1 {
 
 // v2 format is anything older than 0.13.1 when we only updated the internal balances on success ack
 pub mod v2 {
-    use crate::state::{CHANNEL_INFO, CHANNEL_STATE};
+    use crate::amount::Amount;
+    use crate::state::{ChannelState, CHANNEL_INFO, CHANNEL_STATE};
     use crate::ContractError;
-    use cosmwasm_std::{Coin, DepsMut, Env, Order, StdResult};
+    use cosmwasm_std::{to_binary, Addr, DepsMut, Env, Order, StdResult, WasmQuery};
+    use cw20::{BalanceResponse, Cw20QueryMsg};
 
-    pub fn update_balances(deps: DepsMut, env: &Env) -> Result<(), ContractError> {
+    pub fn update_balances(mut deps: DepsMut, env: &Env) -> Result<(), ContractError> {
         let channels = CHANNEL_INFO
             .keys(deps.storage, None, None, Order::Ascending)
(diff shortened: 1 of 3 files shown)


contracts/cw20-ics20/src/migrations.rs
```diff
@@ -14,3 +14,43 @@ pub mod v1 {
...
+                let channel = &channels[0];
+                let addr = &env.contract.address;
+                let states = CHANNEL_STATE
+                    .prefix(channel)
+                    .range(deps.storage, None, None, Order::Ascending)
+                    .collect::<StdResult<Vec<_>>>()?;
+                for (denom, mut state) in states.into_iter() {
+                    // this checks if we have received some coins that are "in flight" and not yet accounted in the state
+                    let Coin { amount, .. } = deps.querier.query_balance(addr, &denom)?;
+                    let diff = state.outstanding - amount;
```
giansalex, line 40:
```suggestion
                    let diff = amount - state.outstanding;
```
should be cw-balance > state.outstanding ?

assuming cw-balance > state.outstanding, because state will be updated in ack (old version).

ethanfrey: Good catch! This is why I need tests (to panic here), or good reviewers

ethanfrey: still missing the handling for cw20 tokens...

PR #694, opened 2022-03-29
storage-plus: Implement u128 key

orkunkl:
I needed to implement a `(Uint128, &Addr)` composite keymap, since u128 is not implemented for `PrimaryKey` I was not able to.
Not sure if u128 is intentionally not implemented here, just pushing it here in case it is left over

### DIFF
--- a/packages/storage-plus/src/keys.rs
+++ b/packages/storage-plus/src/keys.rs
@@ -22,7 +22,7 @@ impl<'a> AsRef<[u8]> for Key<'a> {
             Key::Val16(v) => v,
             Key::Val32(v) => v,
             Key::Val64(v) => v,
-            Key::Val128(v) => v
+            Key::Val128(v) => v,
         }
     }
 }
--- a/packages/storage-plus/src/keys.rs
+++ b/packages/storage-plus/src/keys.rs
@@ -11,6 +11,7 @@ pub enum Key<'a> {
     Val16([u8; 2]),
     Val32([u8; 4]),
     Val64([u8; 8]),
+    Val128([u8; 16]),
 }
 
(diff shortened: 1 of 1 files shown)


ueco-jb (commented):
> since u128 is not implemented for PrimaryKey I was not able to.
Not sure if u128 is intentionally not implemented here

Yes it was. https://github.com/CosmWasm/cw-plus/pull/547#discussion_r749359057
The real question is - do you really need 128 bit key?
@maurolacy ?

orkunkl:
I building an optimized version of [Claim](https://github.com/CosmWasm/cw-plus/blob/main/packages/controllers/src/claim.rs and a sorted index that will save a sorted list of a defined number of claims.
I am using prefixed keys but saving the data to key as well.

```
pub total_claim_sort_index: Map<'a, (u128, &'a Addr), bool>
```

To achieve the sorted claim list, indexing the total claim amount with the address as appended key, bool is just a placeholder.

orkunkl:
Looks like doubling the size of every key will be very costly for this feature...

maurolacy:
> Looks like doubling the size of every key will be very costly for this feature...

Yes. We can add this, but better to avoid it.

My question would be, why you can't use `u64` for your key?

Converting the key to a string would work too. Or using a tuple of two `u64` values. Which is somewhat clumsy, but effective.

maurolacy (approved):
On second thought, I think this wouldn't hurt, as those bytes are in memory / stack in the first place, and do not hit storage.

Approving. Will create a follow-up issue to benchmark `Key` performance with and without u128 support, so we confirm this.

**Update**: #695.

PR #699, opened 2022-04-09
Upgrade all contracts and packages to cosmwasm-std beta8

the-frey:
Contracts using multi-test currently fail on cw-std beta8 because all addresses are now required to be normalized/lower case. This should fix that issue.

Specifically this is done by lower-casing the return value of the `next_address` function in multi-test.

### DIFF
--- a/packages/multi-test/Cargo.toml
+++ b/packages/multi-test/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cw-multi-test"
-version = "0.13.1"
+version = "0.13.2"
 authors = ["Ethan Frey <ethanfrey@users.noreply.github.com>"]
 edition = "2018"
 description = "Test helpers for multi-contract interactions"
@@ -18,8 +18,8 @@ staking = ["cosmwasm-std/staking"]
 backtrace = ["anyhow/backtrace"]
 
 [dependencies]
-cw-utils = { path = "../../packages/utils", version = "0.13.1" }
-cw-storage-plus = { path = "../../packages/storage-plus", version = "0.13.1"}
+cw-utils = { path = "../../packages/utils", version = "0.13.2" }
+cw-storage-plus = { path = "../../packages/storage-plus", version = "0.13.2"}
 cosmwasm-std = { version = "1.0.0-beta8", features = ["staking"] }
 cosmwasm-storage = { version = "1.0.0-beta8" }
 itertools = "0.10.1"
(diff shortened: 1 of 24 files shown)


the-frey:
Question: should this also update contracts to cosmwasm-std beta8 at the same time? That should then confirm the fix works for all contracts that are using multi-test.

maurolacy:
> Question: should this also update contracts to cosmwasm-std beta8 at the same time? That should then confirm the fix works for all contracts that are using multi-test.

Yes, please. Let's transform this into a "upgrade cosmwasm deps" PR, and this fix is part of it.

the-frey:
@maurolacy cool, will do.

packages/multi-test/src/wasm.rs
```diff
@@ -779,7 +779,8 @@ where
             .range_raw(storage, None, None, Order::Ascending)
             .count();
         // we make this longer so it is not rejected by tests
-        Addr::unchecked(format!("Contract #{}", count))
+        // it is lowercase to be compatible with beta8
```
webmaster128, line 782:
```suggestion
        // it is lowercase to be compatible with the MockApi implementation of cosmwasm-std >= 1.0.0-beta8
```

packages/multi-test/src/wasm.rs
```diff
@@ -779,7 +779,8 @@ where
...
+        Addr::unchecked(format!("contract{}", count))
```
webmaster128, line 783: Correct fix, thank you

packages/multi-test/Cargo.toml
```diff
@@ -20,8 +20,8 @@ backtrace = ["anyhow/backtrace"]
 [dependencies]
 cw-utils = { path = "../../packages/utils", version = "0.13.1" }
 cw-storage-plus = { path = "../../packages/storage-plus", version = "0.13.1"}
-cosmwasm-std = { version = "1.0.0-beta6", features = ["staking"] }
-cosmwasm-storage = { version = "1.0.0-beta6" }
+cosmwasm-std = { version = "1.0.0-beta8", features = ["staking"] }
+cosmwasm-storage = { version = "1.0.0-beta8" }
```
webmaster128, line 24: I guess we want to do this change for all contracts and packages (simply search for `1.0.0-beta6` in the whole project)

webmaster128 (commented):
Nice

packages/multi-test/Cargo.toml
the-frey: Yep. Have done this and made version 13.1 -> 13.2. Just about to push.

packages/multi-test/src/wasm.rs
ueco-jb: Does it supposed to be without space?

webmaster128:
> supposed to be without separation?

I think this is a matter of taste. It was `Contract #0`, `Contract #1`, … before and is now `contract0`, `contract1`, …. Having an address without spaces is a bit closer to reality. Since it is lower case letters followed by digits, it's still readable nicely.

webmaster128 (approved):

ueco-jb:
If you put it that way, I'm all in favor.

PR #700, opened 2022-04-12
Add existence checking to indexed map

shanev:
Had the need for existence checking so added `has()` to indexed map. Hope I did it correctly :)

### DIFF
--- a/packages/storage-plus/src/indexed_map.rs
+++ b/packages/storage-plus/src/indexed_map.rs
@@ -130,6 +130,12 @@ where
         self.primary.may_load(store, key)
     }
 
+    /// has returns true or false if any data is at this key, without parsing or interpreting the
+    /// contents.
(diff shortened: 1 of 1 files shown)


ueco-jb (approved):
lgtm
Thank you for your contribution.

Just nitpicking - I see you've copied from other files (which is fine), though still part of test is rather redundant.
For example, there's no point in creating references in tests, since we don't care about performance there. To be extra picky, `load it properly` section is tested in other cases as well.

Whole test case could basically look like this:
```rust
    #[test]
    fn existence() {
        let mut store = MockStorage::new();
        let map = build_map();
        let (pks, _) = save_data(&mut store, &map);

        assert!(map.has(&store, pks[0]));
        assert!(!map.has(&store, "6"));
    }
```

shanev:
@ueco-jb Simplified the test. Thanks for the feedback.

PR #702, opened 2022-04-13
Handle duplicate members in cw4-group create

codehans:
When a cw4-group contract is instantiated with a list of members where some are duplicated, the iterator overwrites any previous member weight, whilst still incrementing the total weight. This PR merges duplicate weights in the `create` function


ueco-jb:
@codehans will you pick up that PR and add some proper test?

ezekiiel:
fwiw - i do think this could be considered to be a non-backwards compatible change.

for example, before [our audit](https://github.com/oak-security/audit-reports/blob/master/DAO%20DAO/2022-06-22%20Audit%20Report%20-%20DAO%20DAO%20v1.0.pdf) (see the first issue there) we did not dedup the members list when instantiating a dao that uses cw4-groups. we [now do](https://github.com/DA0-DA0/dao-contracts/blob/693a5d6a36ada1505c5a52e8029b1b80ea557393/contracts/cw4-voting/src/contract.rs#L30-L38), but were this to land we'd probably stop doing that to be consistent with the new behavior.

codehans:
apologies @ueco-jb @ezekiiel missed this one. this isn't something we're actively using or particularly precious about. happy to leave as-is, especially as time is fairly scarce these days

uint:
This looks like a serious flaw. If you don't mind, I'll pick this up, rebase, add a test, and merge. @codehans do you mind if I push here?

uint:
> fwiw - i do think this could be considered to be a non-backwards compatible change.

@0xekez That's fair! I don't think we guarantee backwards-compatibility currently (`0.*` semantic versions). Is this going to cause problems for you?

codehans:
> This looks like a serious flaw. If you don't mind, I'll pick this up, rebase, add a test, and merge. @codehans do you mind if I push here?

go ahead!

uint:
@codehans Would you mind signing the CLA thing? I'll need that to merge later.

ethanfrey (approved):
lgtm with error on duplicates

PR #703, opened 2022-04-13
Make `HooksResponse` public.

ezekiiel:
This not being public makes testing contracts using this a bit of a
pain.

### DIFF
--- a/contracts/cw20-base/src/contract.rs
+++ b/contracts/cw20-base/src/contract.rs
@@ -225,6 +225,9 @@ pub fn execute(
             marketing,
         } => execute_update_marketing(deps, env, info, project, description, marketing),
         ExecuteMsg::UploadLogo(logo) => execute_upload_logo(deps, env, info, logo),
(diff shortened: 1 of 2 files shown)


ueco-jb (commented):
Looking at source code:
https://github.com/CosmWasm/cw-plus/blob/a2b3653ba146bec6a38660400ac8fae1ca742d1c/packages/controllers/src/hooks.rs#L14-L19
Would you mind doing a proper fix here? Put one `HooksResponse` in `packages/utils/hooks_response.rs`, then remove it both from `cw4` and `controllers`.

ezekiiel:
@ueco-jb I think the TODO comment there may be a little misleading. `cw4` does already use the `HooksResponse` defined in `controllers`. It gets around needing the type to be public because it returns `HOOKS.query_hooks` in it's query message.

I could still move the `HooksResponse` type to utils if you'd like though it seems to me that keeping it inside the hooks package may make the most sense here.

ueco-jb (approved):
LGTM

PR #705, opened 2022-04-15
No Minter Params Error added in cw20 Mint function (to differentiate between unauthorized via wrong sender)

brianlee1597:
The unauthorized error when mint was not configured in instantiate was a bit misleading


ueco-jb (approved):
lgtm

PR #706, opened 2022-04-15
Code to convert cw20 to bank denom and vice versa

mankenavenkatesh:
## Cw20 ↔ Bank denom conversion

This PR extends Cw20 base contract with functionality to convert Cw20 tokens into Bank denom  and vice versa.
### Advantages
- Cw20 tokens can be used as native tokens on the cosmos chain.
- Cw20 tokens can be transferred through IBC to other cosmos chains.
- Cw20 tokens  can be transferred to Ethereum if the bridge from Cosmos to Ethereum exists. etc

### Implementation
To achieve this, the CW20-base contract should be extended with two new functions 
- [x]  **bank_to_cw20**
    - User can invoke this function to convert bank denom to cw20 tokens. Upon invocation,
        - the CW20 contract should verify the user bank balance.
        - transfer the bank denom to CW20 contract address from user balance.
        - Mint equal amount of CW20 tokens for user address.        
- [x]   **cw20_to_bank**
    - User can invoke this function to convert cw20 tokens to bank denom. Upon invocation,
        - the CW20 contract should verify the user cw20 balance.
        - Burn cw20 tokens of the user
        - transfer equal amount of bank denom to user from CW20 contract address.

### DIFF
--- a/contracts/cw20-base/src/contract.rs
+++ b/contracts/cw20-base/src/contract.rs
@@ -1,7 +1,8 @@
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::entry_point;
 use cosmwasm_std::{
-    to_binary, Binary, Deps, DepsMut, Env, MessageInfo, Response, StdError, StdResult, Uint128, BankMsg, coins,
+    coins, to_binary, BankMsg, Binary, Deps, DepsMut, Env, MessageInfo, Response, StdError,
+    StdResult, Uint128,
 };
 
 use cw2::set_contract_version;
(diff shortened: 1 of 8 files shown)


contracts/cw20-base/src/contract.rs
```diff
@@ -228,6 +233,95 @@ pub fn execute(
...
+    TOKEN_INFO.update(deps.storage, |mut info| -> StdResult<_> {
+        info.total_supply = info.total_supply.checked_sub(amount)?;
+        Ok(info)
+    })?;
+
+    // transfer the bank denom to user address from CW20 contract address.
+    // Create and submit BankTransfer sub message. Here sender is the contract address.
+    let config = TOKEN_INFO.load(deps.storage)?;
+
+    let bank_transfer = BankMsg::Send {
```
ethanfrey, line 272: This assumes the contract starts with large (unbounded?) amount of such bank_denom.
That is not the usual case and I don't believe it fits in cw-plus.

I would encourage this to exist in cw-tokens repo, which is more open to community usage of cw20, not the basic standard.

ethanfrey (commented):
I would re-submit this as a PR on cw-tokens, but first I don't understand the context.

Osmosis is allowing contracts to *mint* tokens which will enable something like this. Failing that, the only other use I understand it eg. "wrapped ETH".

Originally token contract has 0 supply. As you send `bank_denom` to it, it will mint cw20 tokens. As you burn cw20 tokens, it will release the `bank_denom`.

With that explanation, this could work (as a cw-token) extension.

That could also be a separate contract that is just the "minter" of a cw20-base contract and has these two functions, where it can receive bank and tell cw20 to mint, or receive cw20 and burn it and send native.

Personally, I prefer the second approach as it makes a simple stand-alone contract that is interoperable with all cw20 tokens (many are not just the cw20-base implementation).

A good explanation in README is key here

uint:
Closing due to inactivity and lack of relevance to the `cw20` specification.

The `cw20-base` contract is meant to demonstrate how to implement the `cw20` specification only.

PR #707, opened 2022-04-16
fix(multi-test): contract mock name for 0.9.x

tk-o:
This PR makes the mock name lowercase so the mocked address could pass the normalized validation check in `cosmwasm-std@0.16.7`

Related to: https://github.com/CosmWasm/cw-plus/pull/699

FYI: @webmaster128 @ethanfrey

### DIFF
--- a/packages/multi-test/src/wasm.rs
+++ b/packages/multi-test/src/wasm.rs
@@ -741,7 +741,8 @@ where
             .range(storage, None, None, Order::Ascending)
             .count();
         // we make this longer so it is not rejected by tests
-        Addr::unchecked(format!("Contract #{}", count.to_string()))
+        // it is lowercase to be compatible with the MockApi implementation of cosmwasm-std >= 0.16.7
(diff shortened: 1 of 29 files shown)


packages/multi-test/src/wasm.rs
```diff
@@ -741,7 +741,8 @@ where
             .range(storage, None, None, Order::Ascending)
             .count();
         // we make this longer so it is not rejected by tests
-        Addr::unchecked(format!("Contract #{}", count.to_string()))
+        // it is lowercase to be compatible with the MockApi implementation of cosmwasm-std >= 0.16.7
+        Addr::unchecked(format!("contract #{}", count.to_string()))
```
webmaster128, line 745: Let's use the format we have in the main branch as well

```suggestion
        Addr::unchecked(format!("contract{}", count))
```

webmaster128 (commented):
LGTM

ueco-jb (approved):

webmaster128 (approved):

PR #712, opened 2022-04-26
Tracing contract

hashedone:


ethanfrey:
> [Getting rid of usize](https://github.com/CosmWasm/cw-plus/pull/712/commits/315cc41f98d147efed49ac9354bd5d38887ad7e0)

We need to make some very clear, obvious docs with this, as it has bit many people over time. It is very non-intuitive and the biggest source of float errors.

contracts/tracer/src/contract.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    match msg {
+        Log { depth } => to_binary(&query_log(deps, depth)?),
+    }
+}
+
+pub fn query_log(deps: Deps, depth: Option<u64>) -> StdResult<LogResponse> {
+    let mut log = LOG.load(deps.storage)?;
+    if let Some(depth) = depth {
+        log = log[log.len() - (depth as usize)..].into();
```
webmaster128, line 170: `usize` is 32 bit when compiled to Wasm. Woudn't it be safer and cleaner to use `depth: Option<u32>` such that the cast is safe and the code behaves the same in Wasm and unit tests?

hashedone: Actually true. I was not sure about usize size, I use `u64` by default, but good point.

hashedone:
> > [Getting rid of usize](https://github.com/CosmWasm/cw-plus/pull/712/commits/315cc41f98d147efed49ac9354bd5d38887ad7e0)
> 
> We need to make some very clear, obvious docs with this, as it has bit many people over time. It is very non-intuitive and the biggest source of float errors.

No here, but yes, I agree.

webmaster128:
> I thing that `usize` should be useable in messages.

I don't think so. I would strongly discourage the use of it whenever possible. Independent of serde and floats, this type has different size when compiled to Wasm (32 bit) and in unit tests (64 bit). Something that behaves differently in tests and production feels very dangerous.

hashedone:
@webmaster128 I kind of agree, but there are cases you just cannot avoid usizes - eg. when you work with stdlib lengths or even iterators - you have them one way on another. And you can insist, that those are behaving differently on UT and prod and technically you would be right - but do we ever unit test for numbers not fitting in `u32` but fitting `u64`? I don't recall single such case. If I would ever use `usize` in the message is because I would just it only in `usize` context, not as general purpose integer. And in such cases using there anything not fitting in `u32` for test would probably blow up it because of out of memory (as it mostly relates to data count, and those would need to be stored somewhere in test to make a test reasonable).

Long story short - I understand your reasoning why you don't like usize and in general I would avoid it too. However I see some cases when I think it would make sense to allow this in messages.

Anyway - it is not a problem of this PR, as I removed usize from interfaces anyway.

maurolacy (commented):
Some comments. Can you add an example / description of how this contract would be used?

contracts/tracer/src/msg.rs
```diff
@@ -0,0 +1,38 @@
...
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum QueryMsg {
+    Log { depth: Option<u32> },
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct LogResponse {
+    pub log: Vec<Vec<LogEntry>>,
```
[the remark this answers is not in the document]
hashedone: Because we want to have a view on change history. The sole purpose of this contract is to show how things changed step by step on more complicated sub calls sequences.

In other words - every element of a vector is a whole log from the beginning. Technically it is obvious for us, that it would just be extended one by one, but the purpose is to not assume we know it happens, and there are no strange ordering things, it is to show it using this contract.

hashedone:
@maurolacy example of this contract usage is basically put into tests. All those MT before I created them, I executed on uploaded contracts to observe how exactly state changes in different situations (I actually executed more messages, but I decided to put those in MTs as I think they cover important findings, and verification is clear enough). It serves two purposes.

First - it is a kind of "smart-contract level manual test" for all the behaviour described in https://github.com/CosmWasm/cosmwasm/blob/main/SEMANTICS.md - running complicated "tree-like" compositions of `forward` message, we can check how exactly contract state is changing after every execution one by one (and which one are missing in the log, so were rolled back as failed).

Second - to practically show what is described in the SEMANTICS.md file - to me at least, reading this file was always a puzzle - not because it was badly written, but because the things around when things are applied to the state, and when exactly calls are reverted are entangled. I created this contract so I was able to prepare arbitrarily completed sub-messages flows (including both multiple contracts calling to each other, and a single contract calling itself), and watch the exact changes state (also being able to fail at any point - either in execution or reply). I think it may be a good teaching tool to present this behaviour.

maurolacy:
Will this be merged? Can we close it?

uint:
Hey @hashedone. I'm trying to reorganize `cw-plus`. The idea is that this repo will soon have a clearer purpose - it's supposed to be purely for `cw-*` specs and their implementations.

That means the tracing contract is not a good fit here anymore.

It looks like it deserves a repo of its own though. If you do that, I'll make sure to include it in [`cw-awesome`](https://github.com/CosmWasm/cw-awesome). `cw-awesome` will probably soon be linked to from README files in our repos as a listing of important/useful resources.

I'm closing this PR, but let me know what's going to happen to `tracer`!

PR #713, opened 2022-04-28
Update item.rs: typo

rtviii:


packages/storage-plus/src/item.rs
```diff
@@ -10,7 +10,7 @@ use crate::helpers::{may_deserialize, must_deserialize};
 
 /// Item stores one typed item at the given key.
 /// This is an analog of Singleton.
-/// It functions just as Path but doesn't ue a Vec and thus has a const fn constructor.
+/// It functions the same way Path does but doesn't use a Vec and thus has a const fn constructor.
```
ueco-jb, line 13:
```suggestion
/// It functions the same way as Path does but doesn't use a Vec and thus has a const fn constructor.
```

ueco-jb (approved):
Add the `as` to keep the proper form and it's good to go.

LGTM

PR #715, opened 2022-05-03
Add code coverage to CI

### LINKED ISSUE #172: Add code coverage tooling to the CI
It would be great to see the code coverage on the contract code from the unit tests. To see if critical checks/branches are properly covered.

One popular tool seems to be [`tarpaulin`](https://github.com/xd009642/tarpaulin). But I am open to any other tool as well. (Note tarpaulin has a [caveat about signals](https://github.com/xd009642/tarpaulin#note-on-tests-using-signals) which is needed for any of the wasm/vm tests, but not when running contracts as native rust).

Deliverable:
* A command to get code coverage for each contract (bonus: also the packages)
* A way to visualize it locally (even if just text)
* Automated upload to a web service in CI (eg. https://coveralls.io or https://codecov.io)

It would nice to run this locally as well. You can look at #171 as a possible way for scripts to work in CI as well as locally.

maurolacy:
Closes #172.

Running this in all dev branches seems overkill. It also takes ~8 minutes to run.


webmaster128:
> Running this in all dev branches seems overkill. It also takes ~8 minutes to run.

I think it's worth it because if you have codecov set up, it tells you if your PR has enough diff coverage. I had this in https://github.com/CosmWasm/cosmwasm/pull/1286 where the tool made me add more tests to pass the codecov checks.

webmaster128 (commented):
Looks like a good start. I'd add a codecov config as otherwise you hardly have visibility of the results.

maurolacy:
> Looks like a good start. I'd add a codecov config as otherwise you hardly have visibility of the results.

Nice. I'll copy / adapt it from cosmwasm.

webmaster128:
Also have a look at https://github.com/CosmWasm/cosmwasm/blob/main/codecov.yml, especially the first line avoids annoying comments from the bot.

.circleci/config.yml
```diff
@@ -34,6 +38,17 @@ workflows:
...
+
+  coverage:
+    jobs:
+      - coverage:
+          filters:
+            branches:
+              only:
+                # Long living branches
+                - main
+                # 👇Add your branch here if code coverage matters to your work
```
ueco-jb, line 50: Let's stick to plain old ASCII, I already removed those emojis in some different PR. :)
And I agree with Simon, I'd remove filter completely. Or maybe even put `except` clause for `main` and `tags` - since coverage here is for devs, and not to fullfill some corpo requirements.

webmaster128: I'll add new emojis elsewhere, no worries

maurolacy: Thanks, forgot about this change. I'll remove the filter.

ueco-jb (approved):
LGTM

PR #719, opened 2022-05-14
Cw derive instantiation

hashedone:
Finally - the first contract *fully migrated* to the ng framework. The framework itself is not finished - the "glue" message to be generated, and many things to be done. But what is here is working POC (even almost MVP I would say) and I would really like to merge it - first to cw-derive), and then finally to main so I don't triply rebase all the time. There is cleanup needed for sure, some refactoring and renaming, but it is not yet published on crates and does not go into any conflicts with anything else, so I don't think it is harmful to merge it in this state.

Not only working tests but tested on real-life actual blockchain - everything as it should be. I see some complaints about missing schema and clippy, but those are minors, it is review ready, and I will go back to this on actual working day.


contracts/cw1-whitelist-ng/Cargo.toml
```diff
@@ -25,17 +24,20 @@ multitest = ["cw-multi-test", "anyhow"]
 cw-utils = { path = "../../packages/utils", version = "0.13.2" }
 cw1 = { path = "../../packages/cw1", version = "0.13.2" }
 cw2 = { path = "../../packages/cw2", version = "0.13.2" }
-cosmwasm-std = { version = "1.0.0-beta8", features = ["staking"] }
+cosmwasm-std = { version = "1", features = ["staking"] }
```
ethanfrey, line 27: 🎉

contracts/cw1-whitelist-ng/Cargo.toml
```diff
@@ -25,17 +24,20 @@ multitest = ["cw-multi-test", "anyhow"]
...
 cw-storage-plus = { path = "../../packages/storage-plus", version = "0.13.2" }
 schemars = "0.8.1"
 serde = { version = "1.0.103", default-features = false, features = ["derive"] }
 thiserror = { version = "1.0.23" }
 cw-multi-test = { path = "../../packages/multi-test", version = "0.13.2", optional = true }
 anyhow = { version = "1", optional = true }
+cw-derive = { version = "0.13.2", path = "../../packages/cw-derive" }
+serde-cw-value = { version = "0.7.0", git = "https://github.com/hashedone/serde-cw-value.git" }
```
ethanfrey, line 35: 👀 you don't need patch directive?

contracts/cw1-whitelist-ng/src/msg.rs
```diff
@@ -1,147 +1,123 @@
...
 use crate::error::ContractError;
-use crate::interfaces::*;
+pub use crate::interfaces::{cw1_msg, whitelist};
 use crate::state::Cw1WhitelistContract;
 
-#[derive(Serialize, Deserialize, JsonSchema)]
-pub struct InstantiateMsg {
-    pub admins: Vec<String>,
-    pub mutable: bool,
+#[derive(Serialize, Clone, PartialEq, JsonSchema, Debug)]
```
ethanfrey, line 16: This part was to be auto-generated, right?
Fine if you do that in a future PR (this is already a lot), but just want to clarify that this code (especially dispatch implementation) is not intended that people write in ng, right?

contracts/cw1-whitelist-ng/src/contract.rs
```diff
@@ -22,12 +16,12 @@ pub fn validate_admins(api: &dyn Api, admins: &[String]) -> StdResult<Vec<Addr>>
...
 
+#[cw_derive::contract(module=msg)]
 impl<T> Cw1WhitelistContract<T> {
+    #[msg(instantiate)]
     pub fn instantiate(
         &self,
-        deps: DepsMut,
-        _env: Env,
-        _info: MessageInfo,
+        (deps, _env, _info): (DepsMut, Env, MessageInfo),
```
ethanfrey, line 24: This structure is fine, but why not define something like `type MutContext = (DepsMut, Env, MessageInfo)` to make this easier to write?

ethanfrey: But then I guess that opens the "naming" issue.

More suggestion for future cleanup

contracts/cw1-whitelist-ng/src/query.rs
```diff
@@ -1,57 +1,47 @@
...
 
-use crate::msg::{AdminListResponse, Cw1QueryMsg, WhitelistQueryMsg};
+use crate::msg::{self, AdminListResponse};
 
 #[must_use]
-pub struct Cw1Querier<'a, C>
-where
-    C: CustomQuery,
-{
+pub struct Cw1Querier<'a> {
```
ethanfrey, line 10: Why choice to remove `C` here?
Not an objection, just curious if it was design choice, or too difficult to implement for now?
(As you did include generic `T` on messages)

ethanfrey: On the other hand, there is no need here for it, so no big deal...

packages/multi-test/Cargo.toml
```diff
@@ -20,8 +20,8 @@ backtrace = ["anyhow/backtrace"]
 [dependencies]
 cw-utils = { path = "../../packages/utils", version = "0.13.2" }
 cw-storage-plus = { path = "../../packages/storage-plus", version = "0.13.2"}
-cosmwasm-std = { version = "1.0.0-beta8", features = ["staking"] }
-cosmwasm-storage = { version = "1.0.0-beta8" }
+cosmwasm-std = { version = "1.0.0-rc.0", features = ["staking"] }
```
ethanfrey, line 23: How about "1" now 😄

ethanfrey (approved):
Good stuff.

Some questions / ideas for future ng development, but I approve merging this (huge) implementation, and then there can be smaller follow up PRs to polish off the interfaces

👍 🎉  glad to see this working

contracts/cw1-whitelist-ng/Cargo.toml
hashedone: I don't understand? I can use git normally, even if tag, or slug. I mean I need to correct the git repo as I gave ownership to CosmWasm today, but on the moment of pushing it it was correct ;)

contracts/cw1-whitelist-ng/src/msg.rs
hashedone: Absolutely, as top level comment claims:

> "This whole file except `AdminListResponse` shall be generated

Generating those file is the next thing I am working on, should be done tomorrow. However, I want to close up some chapters - I came back to completely not working ng, now it works properly on blokchain.

contracts/cw1-whitelist-ng/src/contract.rs
hashedone: This argument has to implement `From<(DepsMut, Env, MessageInfo)` - this is the only requirement on this. You can define as you want, as long as it is makeable from such a tuple.

I want to deliver types to be used here wrapping those types for each message type, but it is only QoL, need to first think how to structure them. But it can be easily any user defined wrapper type right now.

contracts/cw1-whitelist-ng/src/query.rs
hashedone: Because I inittially messed it up, and now I don't yet support custom query types - only custom message types (I initially supported them both but in the way they had to be the same type which was incorrect). As it is not critical for this very contract, I leave it as it is, but custom queries are definitely to be supported (it is not even hard - the same way as custom messages are supported, just realising it is distinct thing).

packages/multi-test/Cargo.toml
hashedone: You see, I missed it in this place.

contracts/cw1-whitelist-ng/src/msg.rs
ethanfrey: Missed this... just looking at diff.

Perfect 👌

contracts/cw1-whitelist-ng/Cargo.toml
ethanfrey: I thought this was something indirectly imported via serde, not directly here.
Nevermind me, tired and confused about package imports.

hashedone: Ah, no - we need it. And it wont be here finally, long story - I can talk about details in Malaga ;)

PR #720, opened 2022-05-16
Cw derive

hashedone:
So everything here was reviewed in small chunks, mostly by @ethanfrey. It is already a working contract. It is still working to do for v1.0.0, but I really don't want to work on a separate branch any longer - it makes rebasing hell. And it doesn't affect any code which was there before "ng", so it won't harm anything.


hashedone:
Closing in favor of https://github.com/CosmWasm/sylvia/pull/1

PR #721, opened 2022-05-16
Repo reclippization

hashedone:
After deprecation of a type in STD it was still used in some packages, which caused not compiling them (or rather - not passing clippization step) on rebuilding them. This PR solves all clippy issues in the repo (also some other minor onces), and updates `Cargo.lock` to make sure all packages are build with `1.0.0` of `cosmwasm-std`.


ethanfrey (approved):
lgtm

PR #722, opened 2022-05-16
Use deny_unknown_fields and combine all query responses in one file

ethanfrey:
This is for testing formats that play well with ts codeine tooling.

@pyramation this is for you


pyramation:
https://gist.github.com/pyramation/5eab090a565e95c5e143f419337b0e6d

How does this look?

ethanfrey:
> https://gist.github.com/pyramation/5eab090a565e95c5e143f419337b0e6d
> 
> How does this look?

Very nice. Cleaner without the additional properties

Strictly speaking `QueryResponse` doesn't need to be exposed. The main item is that it helps you find the return types without relying on naming convertions.

ethanfrey:
https://github.com/CosmWasm/cw-plus/pull/722/commits/d1b336ccf3bf0fdf7da8212fb0e7e4e75c4fde35 is an approach to combine it all into one file, which will make for much easier distribution

ethanfrey:
Not small, but not as big as I feared, especially if stored gzipped.

```
% ls -l schema 
-rw-r--r-- 1 ethanfrey staff 10625 May 16 22:20 api.json
-rw-r--r-- 1 ethanfrey staff  1327 May 16 22:20 api.json.gz
```

ethanfrey:
IIRC, another request was returning some information on the Events that are returned from the different messages that can be executed?

That would definitely be a different task, but good to gather a Wishlist.

Anything else I should have in mind

pyramation:
Ok, I just added a `packed` option, not sure if that's the best name, but I got https://github.com/CosmWasm/cw-plus/commit/d1b336ccf3bf0fdf7da8212fb0e7e4e75c4fde35 to work

https://github.com/pyramation/cosmwasm-typescript-gen/commit/7446f19b6ab9ff7abb627c5fa9280a32679baaca

I've added the CW4Group as a fixture so we can continue to improve in testing

pyramation:
Regarding certain messages that we want to remove, happy to adapt the library. I think some of the `{Message}Response` are autogenerated as a naming convention. 

let's definitely get a wish list going! We can also start making issues [in the repo ](https://github.com/pyramation/cosmwasm-typescript-gen/issues)

I think some of naming conventions I derived from looking at the original concept that @JakeHartnell helped me create — I may need to jam with one of you guys to better understand a few items, as I'm more on the JS/TS side of all this — but I'm sure we can improve it!

ueco-jb:
@ethanfrey I think this one can be closed, since follow up is in progress?

pyramation:
yea @ueco-jb I think you can this work is all being done here https://github.com/CosmWasm/cosmwasm/pull/1339 by @uint

PR #729, opened 2022-05-30
add execute msg to update minter

janitachalam:
Allows current minter to update minter. Cap cannot be updated.

### DIFF
--- a/contracts/cw20-base/src/contract.rs
+++ b/contracts/cw20-base/src/contract.rs
@@ -307,8 +307,17 @@ pub fn execute_mint(
         return Err(ContractError::InvalidZeroAmount {});
     }
 
-    let mut config = TOKEN_INFO.load(deps.storage)?;
-    if config.mint.is_none() || config.mint.as_ref().unwrap().minter != info.sender {
+    let mut config = TOKEN_INFO
+        .may_load(deps.storage)?
+        .ok_or(ContractError::Unauthorized {})?;
+
+    if config
+        .mint
+        .as_ref()
+        .ok_or(ContractError::Unauthorized {})?
+        .minter
+        != info.sender
+    {
         return Err(ContractError::Unauthorized {});
     }
 
@@ -386,16 +395,19 @@ pub fn execute_update_minter(
     info: MessageInfo,
     new_minter: String,
 ) -> Result<Response, ContractError> {
-    let mut config = TOKEN_INFO.load(deps.storage)?;
-    // Check that sender is authorized to update minter
-    if config.mint.is_none() || config.mint.as_ref().unwrap().minter != info.sender {
+    let mut config = TOKEN_INFO
+        .may_load(deps.storage)?
+        .ok_or(ContractError::Unauthorized {})?;
+
+    let mint = config.mint.as_ref().ok_or(ContractError::Unauthorized {})?;
+    if mint.minter != info.sender {
         return Err(ContractError::Unauthorized {});
     }
 
     let minter = deps.api.addr_validate(&new_minter)?;
     let minter_data = MinterData {
         minter,
-        cap: config.mint.unwrap().cap,
+        cap: mint.cap,
     };
     config.mint = Some(minter_data);
 
@@ -403,7 +415,7 @@ pub fn execute_update_minter(
 
     Ok(Response::default()
         .add_attribute("action", "update_minter")
-        .add_attribute("minter", new_minter))
+        .add_attribute("new_minter", new_minter))
 }
 
 pub fn execute_update_marketing(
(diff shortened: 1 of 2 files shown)


contracts/cw20-base/src/contract.rs
```diff
@@ -377,6 +380,32 @@ pub fn execute_send(
...
 
+pub fn execute_update_minter(
+    deps: DepsMut,
+    _env: Env,
+    info: MessageInfo,
+    new_minter: String,
+) -> Result<Response, ContractError> {
+    let mut config = TOKEN_INFO.load(deps.storage)?;
+    // Check that sender is authorized to update minter
+    if config.mint.is_none() || config.mint.as_ref().unwrap().minter != info.sender {
```
orkunkl, line 391: Instead of unwrapping, better to use `ok_or` like: https://github.com/CosmWasm/cw-plus/blob/main/contracts/cw20-base/src/contract.rs#L388-L396

contracts/cw20-base/src/contract.rs
```diff
@@ -377,6 +380,32 @@ pub fn execute_send(
...
+        minter,
+        cap: config.mint.unwrap().cap,
+    };
+    config.mint = Some(minter_data);
+
+    TOKEN_INFO.save(deps.storage, &config)?;
+
+    Ok(Response::default()
+        .add_attribute("action", "update_minter")
+        .add_attribute("minter", new_minter))
```
orkunkl, line 406:
```suggestion
        .add_attribute("new_minter", new_minter))
```
looks more explanatory.

contracts/cw20-base/src/contract.rs
```diff
@@ -871,6 +900,50 @@ mod tests {
...
+        let msg = ExecuteMsg::UpdateMinter {
+            new_minter: String::from("new_minter"),
+        };
+
+        let info = mock_info("not the minter", &[]);
+        let env = mock_env();
+        let err = execute(deps.as_mut(), env, info, msg).unwrap_err();
+        assert_eq!(err, ContractError::Unauthorized {});
+    }
+
```
orkunkl, line 946: Just for extra future security, you can write a test that checks `update_minter` updates only minter address not other fields in `MinterData`

orkunkl (approved):
Thanks!

JakeHartnell:
Any updates on this? We would really like to have it for DAO migrations.

contracts/cw20-base/src/contract.rs
```diff
@@ -377,6 +389,35 @@ pub fn execute_send(
...
+    let mut config = TOKEN_INFO
+        .may_load(deps.storage)?
+        .ok_or(ContractError::Unauthorized {})?;
+
+    let mint = config.mint.as_ref().ok_or(ContractError::Unauthorized {})?;
+    if mint.minter != info.sender {
+        return Err(ContractError::Unauthorized {});
+    }
+
+    let minter = deps.api.addr_validate(&new_minter)?;
```
maurolacy, line 407: Nit: You can move this validation before loading the config, to save some cycles in case of error.

packages/cw20/src/msg.rs
```diff
@@ -55,6 +55,8 @@ pub enum Cw20ExecuteMsg {
     /// Only with the "mintable" extension. If authorized, creates amount new tokens
     /// and adds to the recipient balance.
     Mint { recipient: String, amount: Uint128 },
+    /// Only with the "mintable" extension. The current minter may set a new minter.
+    UpdateMinter { new_minter: String },
```
maurolacy, line 59: This is breaking / changing the cw20 spec. In `cw20/README.md` it says:

```
## Mintable

This allows another contract to mint new tokens, possibly with a cap.
There is only one minter specified here, if you want more complex
access management, please use a multisig or other contract as the
minter address and handle updating the ACL there.
```

So, that could be a way of doing what you need, without touching the spec.

@ethanfrey, what do you think?

ethanfrey: I agree with this addition.
It is not spec breaking, it is a backwards compatible addition.

we still maintain that all contracts have 0 or 1 minter, and all existing queries and messages should work the same, right?

ethanfrey: If there is concern, we can just add it to the cw20-base contract and not the general cw20 interface, but I don't see a big issue there.

contracts/cw20-base/src/contract.rs
```diff
@@ -304,8 +307,17 @@ pub fn execute_mint(
         return Err(ContractError::InvalidZeroAmount {});
     }
 
-    let mut config = TOKEN_INFO.load(deps.storage)?;
-    if config.mint.is_none() || config.mint.as_ref().unwrap().minter != info.sender {
+    let mut config = TOKEN_INFO
```
ethanfrey, line 310: why may_load? it should always be present

contracts/cw20-base/src/contract.rs
```diff
@@ -304,8 +307,17 @@ pub fn execute_mint(
...
+        .may_load(deps.storage)?
+        .ok_or(ContractError::Unauthorized {})?;
+
+    if config
```
ethanfrey, line 314: this is nicer than previous code. thnaks

contracts/cw20-base/src/contract.rs
```diff
@@ -377,6 +389,35 @@ pub fn execute_send(
     Ok(res)
 }
 
+pub fn execute_update_minter(
```
ethanfrey, line 392: function looks good

ethanfrey (approved):
good stuff.
just some minor nits

contracts/cw20-base/src/contract.rs
maurolacy: It is just because of style, yes. If there are no more concerns, I think we can merge this as it is.

PR #732, opened 2022-06-02
Dump state multitest

ethanfrey:
Add new function to multitest to dump raw state of any contract by address.

This is similar to a function we have in wasmd and can be useful for some lower-level integration tests, in particular import/export and initialization.


maurolacy (commented):
Looks good! Thanks for this.

maurolacy (approved):
Tested it locally against `tgrade-valset` and it works fine.

PR #734, opened 2022-06-07
Use standard CosmosMsg

ethanfrey:
We had (temporarily) used an internal copy for trait bounds. Let's go back to the original one. (See TODO comment in the deleted file)


ueco-jb (approved):
lgtm

Issue #739, opened 2022-06-17
Add optional executor restriction to cw3-flex


ethanfrey:
In [`cw3-flex-multisig`](https://github.com/CosmWasm/cw-plus/tree/main/contracts/cw3-flex-multisig), [anyone can execute the proposal](https://github.com/CosmWasm/cw-plus/blob/main/contracts/cw3-flex-multisig/src/contract.rs#L189-L195) once the vote has passed. This is not always ideal.

I would add a backwards-compatible `executor: Option<Executor>` to both InstantiateMsg and Config. This could accept the following values:

* `None` - same behavior as now (and same struct)
* `Some(Member{})` -  any member in the voting group (even with 0 points) can execute the contract, same restriction as [those who can propose](https://github.com/CosmWasm/cw-plus/blob/main/contracts/cw3-flex-multisig/src/contract.rs#L91-L99)
* `Some(Only{address})` - only the given address can execute the contract

Use case for `Member` is eg. a swap from a multisig, where you want control of the timing (not someone using a sandwich attack on your trade).

Use case for `Only` is a separate approval / veto step. This is like the US President needing to sign a law once it passed congress. Or even the two houses. It can allow more complex governance structures.

PR #741, opened 2022-06-18
Add optional executor restriction to cw3-flex

ueco-jb:
closes https://github.com/CosmWasm/cw-plus/issues/739

Test are bloated, but all of them are... Some proper refactoring would be useful here.

### DIFF
--- a/contracts/cw3-flex-multisig/src/contract.rs
+++ b/contracts/cw3-flex-multisig/src/contract.rs
@@ -19,7 +19,7 @@ use cw_utils::{maybe_addr, Expiration, ThresholdResponse};
 
 use crate::error::ContractError;
 use crate::msg::{ExecuteMsg, InstantiateMsg, QueryMsg};
-use crate::state::{Config, CONFIG};
+use crate::state::{Config, Executor, CONFIG};
 
 // version info for migration info
 const CONTRACT_NAME: &str = "crates.io:cw3-flex-multisig";
@@ -46,6 +46,7 @@ pub fn instantiate(
         threshold: msg.threshold,
         max_voting_period: msg.max_voting_period,
         group_addr,
+        executor: msg.executor,
     };
     CONFIG.save(deps.storage, &cfg)?;
 
@@ -192,7 +193,26 @@ pub fn execute_execute(
     info: MessageInfo,
     proposal_id: u64,
 ) -> Result<Response, ContractError> {
-    // anyone can trigger this if the vote passed
+    let cfg = CONFIG.load(deps.storage)?;
+
+    // Executor can be set in 3 ways:
+    // - Member: any member of the voting group can execute
+    // - Only: only passed address is able to execute
+    // - None: Anyone can execute message
+    if let Some(executor) = cfg.executor {
+        match executor {
+            Executor::Member => {
+                cfg.group_addr
+                    .is_member(&deps.querier, &info.sender, None)?
+                    .ok_or(ContractError::Unauthorized {})?;
+            }
+            Executor::Only(addr) => {
+                if addr != info.sender {
+                    return Err(ContractError::Unauthorized {});
+                }
+            }
+        }
+    }
 
     let mut prop = PROPOSALS.load(deps.storage, proposal_id)?;
     // we allow execution even after the proposal "expiration" as long as all vote come in before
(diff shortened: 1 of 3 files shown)


contracts/cw3-flex-multisig/src/contract.rs
```diff
@@ -192,15 +193,33 @@ pub fn execute_execute(
...
         return Err(ContractError::WrongExecuteStatus {});
     }
 
+    // Executor can be set in 3 ways:
+    // - Member: any member of the voting group can execute
+    // - Only: only passed address is able to execute
+    // - None: Anyone can execute message
+    let cfg = CONFIG.load(deps.storage)?;
+    if let Some(executor) = cfg.executor {
+        match executor {
```
ethanfrey, line 209: Is this something that could be pulled out to a method on Executor?
Already thinking of reusing it for who can make proposals. (but default is Member)

contracts/cw3-flex-multisig/src/contract.rs
```diff
@@ -1191,6 +1226,128 @@ mod tests {
...
+            Addr::unchecked(Addr::unchecked(VOTER2)),
+            flex_addr,
+            &execution,
+            &[],
+        )
+        .unwrap();
+    }
+
+    #[test]
+    fn execute_with_executor_only() {
```
ethanfrey, line 1286: Nice tests here. And yes, some of that Suite goodness from poe-contracts would be great here

contracts/cw3-flex-multisig/src/state.rs
```diff
@@ -1,16 +1,29 @@
...
 use serde::{Deserialize, Serialize};
 
+use cosmwasm_std::Addr;
 use cw4::Cw4Contract;
 use cw_storage_plus::Item;
 use cw_utils::{Duration, Threshold};
 
+/// Defines who is able to execute proposals once passed
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
+pub enum Executor {
```
ethanfrey, line 11: I went to the issue to add a comment, then saw you had implemented it (thank you!)

My idea was to only use `Option<Executor>` in the InstantiateMsg and store `Executor` in Config.
This also means to add one more enum variant: `Everyone`.

~~When instantiating, if the message has `None`, we assign the default `Executor::Everyone`, but then the application of the default / meaning of "None" is made on init and not later~~ (see below)

contracts/cw3-flex-multisig/src/state.rs
```diff
@@ -1,16 +1,29 @@
...
+
 #[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug)]
 pub struct Config {
     pub threshold: Threshold,
     pub max_voting_period: Duration,
     // Total weight and voters are queried from this contract
     pub group_addr: Cw4Contract,
+    // who is able to execute passed proposals
+    // None means that anyone can execute
+    pub executor: Option<Executor>,
```
ethanfrey, line 26: Disregard the above comment. This is great as it is drop-in state compatible with previous version, not requiring an explicit migration (we should have some placeholder migrate function, but no state change needed). No need to make breaking changes for some aesthetic opinion of mine.

ethanfrey (approved):
Wow. Thanks for the quick reaction here.

Code looks good. I had an idea how to extend it a bit, left comments on the Executor variant and making the check a method.

The other idea (for a different PR) was to use the same pattern for who can submit a proposal, but have that default to "Member" rather than "Everyone" if the InstantiateMsg has `None` in the field.

contracts/cw3-flex-multisig/src/contract.rs
ueco-jb: I moved it under `Config` struct. I think it's more versatile this way.
What do you think?

contracts/cw3-flex-multisig/src/state.rs
ueco-jb: Yes, I considered that but as you pointed out - this way it's completely backward compatible.

contracts/cw3-flex-multisig/src/contract.rs
```diff
@@ -200,25 +200,8 @@ pub fn execute_execute(
...
-                    .ok_or(ContractError::Unauthorized {})?;
-            }
-            Executor::Only(addr) => {
-                if addr != info.sender {
-                    return Err(ContractError::Unauthorized {});
-                }
-            }
-        }
-    }
+    cfg.authorize(&deps.querier, &info.sender)?;
```
ethanfrey, line 204: nice

contracts/cw3-flex-multisig/src/state.rs
```diff
@@ -26,5 +28,29 @@ pub struct Config {
     pub executor: Option<Executor>,
 }
 
+impl Config {
```
ethanfrey, line 31: looks good

ethanfrey (approved):
thanks

PR #745, opened 2022-06-27
Add migrate method to cw20 base.

ezekiiel:
This adds a migrate method to cw20-base. Having this will be nice as it will allow cw20 users to upgrade to a token where the [minter can be upgraded](https://github.com/CosmWasm/cw-plus/pull/729).

We'd particularly like this for DAO DAO as it will make it easier for DAOs to upgrade their governance tokens.


contracts/cw20-base/src/contract.rs
```diff
@@ -1265,6 +1270,73 @@ mod tests {
...
+
+        #[test]
+        fn test_migrate() {
+            let mut app = App::default();
+
+            let cw20_id = app.store_code(cw20_contract());
+            let cw20_addr = app
+                .instantiate_contract(
+                    cw20_id,
+                    Addr::unchecked("meow"),
```
ueco-jb, line 1297: Sure, this is just a one of many tests. But this is still an open source code that will be used in multiple projectes - please replace those random strings with something proper.

ezekiiel: Sure, np.

PR #746, opened 2022-06-28
Fix specification about CW20 Enumerable Queries

lukepark327:
In this PR,
- the `Queries` about `Enumerable` specification, described in the wrong location, has been corrected (relocated correctly).


ueco-jb (approved):
LGTM
Thank you.

PR #749, opened 2022-07-06
Use into_iter() instead of iter().cloned().

ezekiiel:
This ought to be a little more efficient as we don't need to clone every element as we walk the list.

Just for fun, I wrote a [little benchmark](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=51a85eeefe13d255485a656de7a3e2ef) for this, then ran it with a couple different sizes using hyperfine. The `into_iter()` method was around 1.5x as fast.


ueco-jb (approved):
LGTM
Thank you for that, very good spot :)

PR #750, opened 2022-07-06
Add bytes variant of key using `Box<Vec<u8>>`

y-pakorn:
This makes us able to use dynamic size bytes as a key.

One use case is to index `AssetInfo` with utf-8 bytes and deserialize it back properly into the correct enum variant w/o usage of `addr_validate` as written in `Map` unit-test.

Another use case is to write a macro for auto implementing, `PrimaryKey`, `Prefixer`, and `KeyDeserialize` for "any" struct that impl serde `Serialize` and `Deserialize`.

This means that any struct/enum can be used as a `Map` key. Very useful for storing/indexing.

I'm not sure about the performance of this though, bench function might also be needed.

### DIFF
--- a/packages/storage-plus/src/keys.rs
+++ b/packages/storage-plus/src/keys.rs
@@ -12,6 +12,7 @@ pub enum Key<'a> {
     Val32([u8; 4]),
     Val64([u8; 8]),
     Val128([u8; 16]),
+    Bytes(Box<Vec<u8>>),
 }
 
 impl<'a> AsRef<[u8]> for Key<'a> {
@@ -23,6 +24,7 @@ impl<'a> AsRef<[u8]> for Key<'a> {
             Key::Val32(v) => v,
             Key::Val64(v) => v,
             Key::Val128(v) => v,
+            Key::Bytes(v) => v,
         }
     }
 }
--- a/packages/storage-plus/src/keys.rs
+++ b/packages/storage-plus/src/keys.rs
@@ -514,7 +514,7 @@ mod test {
 
     #[test]
     fn nested_bytes_key() {
-        #[derive(Debug, Clone)]
+        #[derive(Debug, Clone, PartialEq)]
         struct Test(String);
 
         impl PrimaryKey<'_> for Test {
@@ -564,5 +564,10 @@ mod test {
         assert_eq!(2, path.len());
         assert_eq!(b"TTT-Test", path[0].as_ref());
         assert_eq!(100u64.to_cw_bytes(), path[1].as_ref());
+
+        assert_eq!(
+            Test::from_vec(b"TTT-F".to_vec()).unwrap(),
+            Test("F".to_string())
+        );
     }
 }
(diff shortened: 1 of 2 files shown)


packages/storage-plus/src/keys.rs
```diff
@@ -12,6 +12,7 @@ pub enum Key<'a> {
     Val32([u8; 4]),
     Val64([u8; 8]),
     Val128([u8; 16]),
+    Bytes(Box<Vec<u8>>),
```
maurolacy, line 15: This is not needed, as you can use `Vec<u8>`, or even `&[u8]` as a keys directly. This is becase `PrimaryKey` is implemented for those types.

y-pakorn: My intention is to make it more convenient to impl `Key`. Instead of pre allocate `Vec<u8>` and use that as `Key`, we can create any custom temporary `Key` from an arbitrary function.

The primary example is serializing/deserializing the enum variant to/from its `Key` form. We can't just use `Key::Ref` because the bytes representation of the `AssetInfo` is created temporarily in the function. (Code below)

```rust
fn key(&self) -> Vec<Key> {
    vec![Key::Ref(
        match self {
            AssetInfo::Native(native) => format!("n-{}", native),
            AssetInfo::Cw20(cw20) => format!("c-{}", cw20),
        }
        .as_bytes(),
    )]
}
```

y-pakorn: Instead, in order to make it work properly, we need to wrap it into an additional struct, let's say `AssetInfoKey`, and then serialize/deserialize through it. Which imo is pretty redundant.

```rust
struct AssetInfoKey(Vec<u8>);
```

y-pakorn: This can be solve using `Key::Bytes` above, since `Box<Vec<u8>>` allocated here is dropped at the end of the scope anyway.

```rust
fn key(&self) -> Vec<Key> {
    vec![Key::Bytes(Box::new(
        match self {
            AssetInfo::Native(native) => format!("n-{}", native),
            AssetInfo::Cw20(cw20) => format!("c-{}", cw20),
        }
        .as_bytes()
        .to_vec(),
    ))]
}
```

maurolacy: > My intention is to make it more convenient to impl `Key`. Instead of pre allocate `Vec<u8>` and use that as `Key`, we can create any custom temporary `Key` from an arbitrary function.

Yes, but the decision not to do that is deliberate. For performance and efficiency reasons, we want the `key` method to return a reference. 

We only hesitantly introduced the `Key` enum and its `Val` variants, because the internal representation of integer types does not match their lexicographical sorting order (basically, because wasm32 ints repr is little-endian).
So, `Val` is there on the condition of the values to be small. Up to 16 bytes max.

Adding a generic `Bytes` will defeat that, so, we will not do it.

y-pakorn: Thanks for clarifying! Then these changes will not be needed!

PR #751, opened 2022-07-07
Add an optional proposal deposit to cw3-flex-multisig.

### LINKED ISSUE #742: Backport from DAO DAO: optionally charge to make proposal
`cw3-flex-multisig` may require a payment in order to create a proposal. This will often not be set on multisig-like implementations, and set on token-weighted voting to stop people from staking one token and spamming proposals.

Requirements:
* The proposal start_time will not be set (not be active) until the required funds are sent
* If the proposal passes, the deposit will be returned

DAO DAO implementation (you can use this, or do something better, but this is a reference):

* https://github.com/DA0-DA0/dao-contracts/pull/11 (to require deposit)
* https://github.com/DA0-DA0/dao-contracts/pull/35 (to refund even on failed vote)

Some notes:

* Don't use [two fields in Instantiate](https://github.com/DA0-DA0/dao-contracts/pull/11/files#diff-2a31c9bce208313e13c9ce0f229156f1f92090938ceb17950cacab5fa33e6dc7R20-R21) but rather [`deposit: Option<Cw20Coin>`](https://github.com/CosmWasm/cw-plus/blob/main/packages/cw20/src/coin.rs#L7-L11) and store [`Cw20CoinVerified`](Cw20CoinVerified). This is also more compatible with current instantiation
* Note: UpdateConfig needs to update this field as well (as in the original PR)
* They use [`TransferFrom` that needs a previous authorization](https://github.com/DA0-DA0/dao-contracts/pull/11/files#diff-1da256e88678c7e356c71cc74bdc1bc0411aa5f9c04115869079cbe5c5ef6a98R168-R188) to do this. That is the least API breaking change and I would stick with it here. Other times I would say use Cw20 Receive interface, but that would be a lot breaking and have two different paths for with deposit and without deposit.

ezekiiel:
Resolves #742 

This adds an optional proposal deposit to cw3-flex-multisig. In the process it refactors shared proposal code from the fixed and flex multisig into the cw3 package.

Both native and cw20 deposits are accepted.


contracts/cw3-flex-multisig/src/contract.rs
```diff
@@ -238,7 +270,19 @@ pub fn execute_close(
...
+    let response = match prop.deposit {
+        Some(deposit) => {
+            if deposit.refund_failed_proposals {
+                Response::new().add_message(deposit.get_return_deposit_message(&prop.proposer)?)
+            } else {
+                Response::new()
+            }
+        }
+        None => Response::new(),
+    };
```
ueco-jb, line 283:
```suggestion
    let mut response = Response::new();
    if let Some(deposit) = prop.deposit {
        if deposit.refund_failed_proposals {
            response = response.add_message(deposit.get_return_deposit_message(&prop.proposer)?)
        }
    }
```

packages/cw20/Cargo.toml
```diff
@@ -17,3 +17,5 @@ serde = { version = "1.0.103", default-features = false, features = ["derive"] }
 
 [dev-dependencies]
 cosmwasm-schema = { version = "1.0.0" }
+cw20-base = { version = "0.13.4", path = "../../contracts/cw20-base" }
+cw-multi-test = { version = "0.13.4", path = "../multi-test" }
```
ueco-jb, line 21: `cw20-base` creates a circular dependency.
Both of those are not used anywhere.
```suggestion
```

packages/cw3/src/deposit.rs
```diff
@@ -0,0 +1,139 @@
...
+#[derive(Error, Debug, PartialEq)]
+pub enum DepositError {
+    #[error("Invalid zero deposit. Set the deposit to None to have no deposit.")]
+    ZeroDeposit {},
+    #[error("Invalid cw20")]
+    InvalidCw20 {},
+    #[error("{0}")]
+    Payment(#[from] PaymentError),
+    #[error("Invalid native deposit amount")]
+    InvalidDeposit {},
```
ueco-jb, line 44:
```suggestion
    #[error("Invalid zero deposit. Set the deposit to None to have no deposit.")]
    ZeroDeposit {},
    
    #[error("Invalid cw20")]
    InvalidCw20 {},
    
    #[error("{0}")]
    Payment(#[from] PaymentError),
    
    #[error("Invalid native deposit amount")]
    InvalidDeposit {},
```

ueco-jb (commented):
I haven't reviewed everything yet, but those are comments so far.
Really nice work. :+1:

contracts/cw3-flex-multisig/src/contract.rs
ezekiiel: Much more clean. :) Thanks.

packages/cw20/Cargo.toml
ezekiiel: Ah - good catch. Thanks. Leftover from a scratched testing approach.

maurolacy:
Can you rebase from main? And we release this in 0.16.

uint:
@0xekez Could you update this to `main`? Or shall I pick this up? I promise this time we won't let this PR sit and get stale ;)

uint (approved):

PR #752, opened 2022-07-13
Fix broken links, minor typo

mikedotexe:
I've been running through a few repos that I want to share with my team and correcting small items.
Pretty self-explanatory. I think when the NFTs moved we forgot a couple links so I'm trying to lean in a little.


ueco-jb (approved):
LGTM
Thank you for fixing those links.

PR #758, opened 2022-07-20
Implement `Index` trait restructuring and update according struct and function (#757)

### LINKED ISSUE #757: Restructure `Index` trait to allow for more extensive `Index` struct implementation.
This should not affect/or break anything. Trait change allows for the implemented `Index` struct can access both `old_data` and `data` and act accordingly.

For the ideas, In some cases, we will not need to update the indexes every update, only handle the first save and remove. This can be called `ConstantMultiIndexes`.

Or we can update the indexes whether the condition of the data updated is met. This can be called `ConditionalMultiIndexes`.

```rust
pub trait Index<T>
where
{
    fn update(
        &self,
        store: &mut dyn Storage,
        pk: &[u8],
        old_data: Option<&T>,
        data: Option<&T>,
    ) -> StdResult<()>;

    fn save(&self, store: &mut dyn Storage, pk: &[u8], data: &T) -> StdResult<()> {
        self.update(store, pk, None, Some(data))
    }

    fn remove(&self, store: &mut dyn Storage, pk: &[u8], old_data: &T) -> StdResult<()> {
        self.update(store, pk, Some(old_data), None)
    }
}
```

y-pakorn:
This should close #757.

### DIFF
--- a/packages/storage-plus/src/indexed_map.rs
+++ b/packages/storage-plus/src/indexed_map.rs
@@ -84,19 +84,16 @@ where
     ) -> StdResult<()> {
         // this is the key *relative* to the primary map namespace
         let pk = key.joined_key();
-        if let Some(old) = old_data {
-            for index in self.idx.get_indexes() {
-                index.remove(store, &pk, old)?;
-            }
+
+        for index in self.idx.get_indexes() {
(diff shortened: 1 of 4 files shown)


maurolacy:
We won't merge this (see main issue).

Issue #764, opened 2022-07-25
Make cw1155 can add token's url at the first mint


ALPAC-4:
There is no way to add url of cw1155 token despite it has [queryMsg](https://github.com/CosmWasm/cw-plus/blob/e360fed40b6148399337770324912bcdc518fa22/contracts/cw1155-base/src/contract.rs#L463-L466) for it.

ethanfrey:
Huh, that contract is something of the orphan.

Since the CosmWasm core team has little to no understanding of the use cases of NFTs, we moved the nft code over to https://github.com/CosmWasm/cw-nfts

There is an open PR to add cw1155 there open for a long time: https://github.com/CosmWasm/cw-nfts/pull/27

Can you please follow up on that repo and see if they can update / maintain this (or volunteer to do so yourself)? There were too many demos, we trimmed down to the ones we could really maintain and let the community manage others.

PR #765, opened 2022-07-25
Make cw1155 can add token's url at the first mint (#764)

ALPAC-4:
#764 

cw1155 need to be able to add token url when the execute of the `mint` and `batch_mint` messages.


ueco-jb (commented):
Of course tests are required.

uint:
@ethanfrey This PR seems to be asking to modify a spec/standard. In my mind, decisions on what goes into a spec have a huge impact on the whole ecosystem that I don't understand. I'm in no way qualified to call the shots here. Can I delegate these PRs to you?

PR #768, opened 2022-07-28
Release 0.14-2

maurolacy:
A follow-up to #767 to fix publishing to crates.io issues.


ueco-jb (approved):
LGTM
Was that about order of releasing crates?

maurolacy:
Indeed. Cyclic deps between the new storage-macro package and storage-plus.

PR #774, opened 2022-08-08
Fix: `Prefix::keys` return errors

### LINKED ISSUE #766: `Prefix::keys` method never fails
Keys are just skipped if deserializating of its key or value fails.
That's because de_fn returns Result which implements Iterator as well and, being iterated, returns nothing in case of error value
https://github.com/CosmWasm/cw-plus/blob/b2a15611d663109d1794e2827f1cf25038e5b9a6/packages/storage-plus/src/prefix.rs#L188
Is it desired behavior?

maurolacy:
Closes #766.

### DIFF
--- a/packages/storage-plus/src/map.rs
+++ b/packages/storage-plus/src/map.rs
@@ -267,6 +267,8 @@ mod test {
     use std::ops::Deref;
 
     use cosmwasm_std::testing::MockStorage;
+    use cosmwasm_std::to_binary;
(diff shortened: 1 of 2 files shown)


ueco-jb (approved):
Can you add some simplest testcase?

maurolacy:
Was thinking about that, but it may be nontrivial to make deserialisation fail. Will see what I can do.

PR #781, opened 2022-08-16
Add primary key to `MultiIndex` index fn params

### LINKED ISSUE #670: Change `MultiIndex` index function signature to include the pk
It would be good that the `MultiIndex` index function signature, instead of being `fn(&T) -> IK`, was `fn(&PK, &T) -> IK` instead.

That way, the pk (or parts of it) can be used for indexing. And so the index keys don't have / need to come from the values alone.

maurolacy:
Closes #670. Adds the (raw) primary key to the `MultiIndex` index fn params.

It is then the index function's responsibility to deserialise the pk, if it wants to use all or parts of it for building the index key.

TODO:

- ~~Add example / tests for this indexing mechanism~~. (done)
- ~~Add MIGRATING entry / notice~~. (done, for upcoming v0.14.0)

### DIFF
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -2,6 +2,30 @@
 
 This guide lists API changes between *cw-plus* major releases.
 
+## v0.13.x -> v0.14.0
+
+### Breaking Issues / PRs
+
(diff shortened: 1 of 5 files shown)


hashedone (approved):
lgtm

Issue #753, opened 2022-07-16
Some multitest bindings for staking are missing such as `BondedDenom`


shanev:
This causes `deps.querier.query_bonded_denom()` to fail in mulitest.

ethanfrey:
@shanev can you provide more details here, like a quick code sample to demo?

It can either be:
1. Building fully functional mocks for Staking (which is useful, but I would push to 0.16 not this release)
2. Export a few more missing types from multitest to allow a 3rd party to implement it (I would do this right away)

I know there was some other bug that someone couldn't build a multi-test mock for Distribution due to missing exports (it wasn't pub). We should check and address all those, maybe trying to build sample/failing mocks there in another module to show we have all types.

ueco-jb:
> Building fully functional mocks for Staking (which is useful, but I would push to 0.16 not this release)

Some time ago I started tinkering around that.
https://github.com/CosmWasm/cw-plus/pull/782/files
It's kind of tricky to properly utilize tokens from Bank module storage, "temporary" I created own `STAKING` storage which acts as staked balances.

ethanfrey:
Okay, then we just verify all types are exported. And close this, leaving some "future issue" for proper multi-test staking support

maurolacy:
> Okay, then we just verify all types are exported. And close this, leaving some "future issue" for proper multi-test staking support

Regarding this, not exactly sure what needs to be done.

`BondedDenom` is published as part of `StakingQuery` in cosmwasm-std

https://github.com/CosmWasm/cosmwasm/blob/421f75cde11ff8842fd279ea3ec2cbcca65ef399/packages/std/src/query/staking.rs#L11-L34

along with the other staking queries.

The same with the staking messages

https://github.com/CosmWasm/cosmwasm/blob/421f75cde11ff8842fd279ea3ec2cbcca65ef399/packages/std/src/results/cosmos_msg.rs#L76-L90

What's required is implement the `Module` trait for staking in multitask, adding support for the staking messages and queries. But that's out of scope here. So, unless I'm missing something related to exporting these types, I would say, this is already done.

ethanfrey:
@ueco-jb has started work on better testing in https://github.com/CosmWasm/cw-plus/pull/782

I assume this is what is desired (when finished)?

PR #782, opened 2022-08-18
[multi-test] Add staking and distribution module

ueco-jb:
Closes https://github.com/CosmWasm/cw-plus/issues/753

Implementation of Staking and Distribution modules in multitest package, which allows now to test staking and delegation.

### DIFF
--- a/packages/multi-test/src/staking.rs
+++ b/packages/multi-test/src/staking.rs
@@ -254,17 +254,23 @@ impl StakeKeeper {
         reward - commission
     }
 
-    /// Updates the staking reward for the given validator. This mutates the validator info,
-    /// but does not save it.
-    /// Always call this to update rewards before changing a validator stake.
+    /// Updates the staking reward for the given validator and their stakers
+    /// It saves the validator info and it's stakers, so make sure not to overwrite that.
+    /// Always call this to update rewards before changing anything that influences future rewards.
     fn update_rewards(
         api: &dyn Api,
         staking_storage: &mut dyn Storage,
         block: &BlockInfo,
-        staking_info: &StakingInfo,
-        validator_info: &mut ValidatorInfo,
-        validator: &Validator,
+        validator: &Addr,
     ) -> AnyResult<()> {
+        let staking_info = Self::get_staking_info(staking_storage)?;
+
+        let mut validator_info = VALIDATOR_INFO
+            .may_load(staking_storage, validator)?
+            .ok_or_else(|| anyhow!("validator not found"))?;
+
+        let validator_obj = VALIDATOR_MAP.load(staking_storage, validator)?;
+
         if validator_info.last_rewards_calculation >= block.time {
             return Ok(());
         }
@@ -273,7 +279,7 @@ impl StakeKeeper {
             block.time,
             validator_info.last_rewards_calculation,
             staking_info.apr,
-            validator.commission,
+            validator_obj.commission,
             validator_info.stake,
         );
 
@@ -281,7 +287,10 @@ impl StakeKeeper {
         if !new_rewards.is_zero() {
             validator_info.last_rewards_calculation = block.time;
 
-            let validator_addr = api.addr_validate(&validator.address)?;
+            // save updated validator
+            VALIDATOR_INFO.save(staking_storage, validator, &validator_info)?;
+
+            let validator_addr = api.addr_validate(&validator_obj.address)?;
             // update all delegators
             for staker in validator_info.stakers.iter() {
                 STAKES.update(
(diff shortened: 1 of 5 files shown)


packages/multi-test/src/staking.rs
```diff
@@ -15,12 +29,240 @@ pub enum StakingSudo {
...
+        msg: StakingMsg,
+    ) -> AnyResult<AppResponse> {
+        let mut staking_storage = prefixed(storage, NAMESPACE_STAKING);
+        match msg {
+            StakingMsg::Delegate { validator, amount } => {
+                let events = vec![Event::new("delegate")
+                    .add_attribute("recipient", &validator)
+                    .add_attribute("sender", &sender)
+                    .add_attribute("amount", format!("{}{}", amount.amount, amount.denom))];
+                self.add_stake(&mut staking_storage, sender, vec![amount])?;
```
ethanfrey, line 126: TODO: we need to deduct from the account that was validating (from the bank module).
I am pretty sure this is possible, I just need some time to do this (and then I will push it)

packages/multi-test/src/staking.rs
```diff
@@ -15,12 +29,240 @@ pub enum StakingSudo {
...
+                Ok(AppResponse { events, data: None })
+            }
+            StakingMsg::Undelegate { validator, amount } => {
+                let events = vec![Event::new("undelegate")
+                    .add_attribute("from", &validator)
+                    .add_attribute("to", &sender)
+                    .add_attribute("amount", format!("{}{}", amount.amount, amount.denom))];
+                self.remove_stake(&mut staking_storage, sender, vec![amount])?;
```
ethanfrey, line 134: TODO: need to do the bank send back to this account as above

packages/multi-test/src/staking.rs
```diff
@@ -15,12 +29,240 @@ pub enum StakingSudo {
...
+                // for now validator is ignored, as I want to support only one validator
+                let delegator = api.addr_validate(&delegator)?;
+                let stakes = match self.get_stakes(&staking_storage, &delegator) {
+                    Ok(stakes) => stakes[0].clone(),
+                    Err(_) => {
+                        let response = DelegationResponse { delegation: None };
+                        return Ok(to_binary(&response)?);
+                    }
+                };
+                // set fixed reward ratio 1:10 per delegated amoutn
```
ethanfrey, line 175: This should definitely be configurable somehow (via sudo or via custom "setup" calls).
Basically, nothing is free, you set it up defining how much rewards, then contract withdraws.
Can simulate contract withdrawing twice in one block, with 1 day between, with.1 year between that way

packages/multi-test/src/staking.rs
```diff
@@ -15,12 +29,240 @@ pub enum StakingSudo {
...
+        _block: &BlockInfo,
+        sender: Addr,
+        msg: DistributionMsg,
+    ) -> AnyResult<AppResponse> {
+        let staking_storage = prefixed(storage, NAMESPACE_STAKING);
+        match msg {
+            // For now it ignores validator as I want to support only one
+            DistributionMsg::WithdrawDelegatorReward { validator } => {
+                let stakes = self.get_stakes(&staking_storage, &sender)?[0].clone();
+                // set fixed reward ratio 1:10 per delegated amoutn
```
ethanfrey, line 233: As above in the staking query, need a good way to configure how many rewards.
Maybe we do it in this module and the staking module calls this internally to service this request?

ethanfrey (commented):
Nice start.
The two big TODO are obvious missing points you mentioned to me.
I designed this to be possible. Probably need 30 minutes in the morning to push such a commit that will compile and do that

chipshort:
This now allows adding multiple validators and the rewards are calculated based on a configurable APR, the validator's commission and the elapsed time, so you can just change `block.time` and send the withdrawal message. I also implemented the unbonding period (tough only when undelegating, not when redelegating). To trigger the actual payout when unbonding, a sudo message has to be sent to process the unbonding queue.
Not sure if this is a good solution, but I couldn't come up with a better way. 

Tarpaulin seems to be a bit silly. It marks a lot of covered lines as uncovered.

JakeHartnell:
Can't wait for this!

uint (approved):
The code looks fine! I can't easily verify this is precisely how staking/distribution works, so I'm taking that on faith.

ethanfrey:
@ueco-jb can you add "Closes #753 " to the description, and maybe update it (I think it isn't WIP anymore)

ueco-jb:
@chipshort In reward formula remember to subtract the validator commision
`reward * (1 - commision)`

chipshort:
> @chipshort In reward formula remember to subtract the validator commision `reward * (1 - commision)`

That should already be the case in `calculate_rewards`, albeit I calculated it in a slightly different way:
```rust
let commission = reward * validator_commission;

reward - commission
```

PR #783, opened 2022-08-19
Multisig status logic

### LINKED ISSUE #677: Redundant logic in `ThresholdResponse` multisigs
The logic for determining if a proposal is passed for a multisig with a `ThresholdResponse` includes an option for considering a expired proposal as passed:

https://github.com/CosmWasm/cw-plus/blob/5c871af54bf67df3d78a4127d15460c1486418c4/contracts/cw3-fixed-multisig/src/state.rs#L78-L81

That condition will never be reached though as the proposal's status is only updated on a vote. It is impossible to vote on an expired proposal so there will never be an `update_status` call which will flip such a proposal to `Status::Passed`.

It is only possible to execute a passed proposal which has been set to passed via a call to `update_status`:

https://github.com/CosmWasm/cw-plus/blob/5c871af54bf67df3d78a4127d15460c1486418c4/contracts/cw3-fixed-multisig/src/contract.rs#L189-L191

Open to thoughts here. Seems to me that logic should either be removed or the status check should call `prop.current_status()` instead of using the one in storage.

maurolacy:
Closes #677. This bug only affects fixed-multisig, as it was already fixed in flex-multisig.

Fixes a related / similar bug in both multisigs, in which a `Passed` (due to expiration) proposal could still be `Closed` after expiration.

### DIFF
--- a/contracts/cw3-fixed-multisig/src/contract.rs
+++ b/contracts/cw3-fixed-multisig/src/contract.rs
@@ -213,8 +213,12 @@ pub fn execute_close(
     let mut prop = PROPOSALS.load(deps.storage, proposal_id)?;
     if [Status::Executed, Status::Rejected, Status::Passed]
         .iter()
-        .any(|x| *x == prop.status)
+        .any(|&s| s == prop.status)
     {
+        return Err(ContractError::WrongCloseStatus {});
+    }
+    // Avoid closing of Passed due to expiration proposals
+    if prop.current_status(&env.block) == Status::Passed {
         return Err(ContractError::WrongCloseStatus {});
     }
     if !prop.expires.is_expired(&env.block) {
--- a/contracts/cw3-fixed-multisig/src/contract.rs
+++ b/contracts/cw3-fixed-multisig/src/contract.rs
@@ -177,7 +177,7 @@ pub fn execute_vote(
 
 pub fn execute_execute(
     deps: DepsMut,
-    _env: Env,
+    env: Env,
     info: MessageInfo,
     proposal_id: u64,
 ) -> Result<Response, ContractError> {
@@ -186,7 +186,7 @@ pub fn execute_execute(
     let mut prop = PROPOSALS.load(deps.storage, proposal_id)?;
     // we allow execution even after the proposal "expiration" as long as all vote come in before
     // that point. If it was approved on time, it can be executed any time.
-    if prop.status != Status::Passed {
+    if prop.current_status(&env.block) != Status::Passed {
         return Err(ContractError::WrongExecuteStatus {});
     }
 
--- a/contracts/cw3-fixed-multisig/src/contract.rs
+++ b/contracts/cw3-fixed-multisig/src/contract.rs
@@ -212,10 +212,7 @@ pub fn execute_close(
     // anyone can trigger this if the vote passed
 
     let mut prop = PROPOSALS.load(deps.storage, proposal_id)?;
-    if [Status::Executed, Status::Rejected, Status::Passed]
-        .iter()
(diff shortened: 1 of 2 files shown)


contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -186,7 +186,7 @@ pub fn execute_execute(
     let mut prop = PROPOSALS.load(deps.storage, proposal_id)?;
     // we allow execution even after the proposal "expiration" as long as all vote come in before
     // that point. If it was approved on time, it can be executed any time.
-    if prop.status != Status::Passed {
+    if prop.current_status(&env.block) != Status::Passed {
         return Err(ContractError::WrongExecuteStatus {});
     }
```
ezekiiel, line 191: Might suggest doing:

```
prop.update_status(&env.block);
if prop.status != Status::Passed {
    return Err(ContractError::WrongExecuteStatus {});
}
```

See this as slightly less error prone as checking `prop.status` below will return correct values (if future code ever does this).

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -213,10 +213,14 @@ pub fn execute_close(
...
         .iter()
-        .any(|x| *x == prop.status)
+        .any(|&s| s == prop.status)
     {
         return Err(ContractError::WrongCloseStatus {});
     }
+    // Avoid closing of Passed due to expiration proposals
+    if prop.current_status(&env.block) == Status::Passed {
+        return Err(ContractError::WrongCloseStatus {});
+    }
```
ezekiiel, line 223: Simpler logic:

```rust
    prop.update_status(&env.block);
    if prop.status != Status::Rejected {
        return Err(ContractError::WrongCloseStatus {});
    }
```

maurolacy: This doesn't preserve the current logic of expired vs. non-expired error messages. That's why I prefer the other variant.

contracts/cw3-fixed-multisig/src/contract.rs
```diff
@@ -213,10 +214,14 @@ pub fn execute_close(
     let mut prop = PROPOSALS.load(deps.storage, proposal_id)?;
     if [Status::Executed, Status::Rejected, Status::Passed]
         .iter()
-        .any(|x| *x == prop.status)
+        .any(|&s| s == prop.status)
```
hashedone, line 217: Nitpicking, but it looks like:
```rust
if  [Status::Executed, Status::Rejected, Status::Passed].contains(&prop.status)
```

contracts/cw3-flex-multisig/src/contract.rs
```diff
@@ -226,10 +227,14 @@ pub fn execute_close(
     let mut prop = PROPOSALS.load(deps.storage, proposal_id)?;
     if [Status::Executed, Status::Rejected, Status::Passed]
         .iter()
-        .any(|x| *x == prop.status)
+        .any(|&s| s == prop.status)
```
hashedone, line 230: Same here

hashedone (approved):
LGTM with the very minor suggestion (the "problem" was not even introduced in this PR, but change happened around).

contracts/cw3-fixed-multisig/src/contract.rs
hashedone: Ignore me - just realized you changed it this way in follow-up.

maurolacy: This is a valid suggestion. Applied it in both multisigs.

PR #784, opened 2022-08-19
Multisig status logic follow-up

maurolacy:
#783 follow-up.

Changes the voting logic so that not only open proposals, but also approved and rejected proposals, can still be voted on before they expire.

This has no impact on the final proposal's status, but it allows voters to better reflect the tallies and results of a proposal's voting, for accounting, statistics and historical information.


hashedone (approved):
lgtm

Issue #776, opened 2022-08-05
Add stack and queue implementations to storage-plus


itsHaseebSaeed:
I was hoping cosmwasm devs can add something like this; if not, I'd love to add such a package.

Description in mind: An "append store" is a storage wrapper that guarantees constant-cost appending to and popping from a list of items in storage. This is achieved by storing each item in a separate storage entry. A unique key is reserved for storing the length of the collection so far.

itsHaseebSaeed:
That makes much more sense, tbh. Let me know what the team thinks. I'm very excited to contribute. Thanks

chipshort:
Hi @itsHaseebSaeed
Is there any work done on the queue already? I need it for something and would like to implement it. If you want, I can also take over the stack PR you started.

itsHaseebSaeed:
> Hi @itsHaseebSaeed Is there any work done on the queue already? I need it for something and would like to implement it. If you want, I can also take over the stack PR you started.

The stack is almost ready but needs a few modifications plus the need to write documentation. Planning to complete it this weekend. Maybe check back on Sunday morning. In the meanwhile, if you see any errors feel free to comment.

uint:
We just talked on Slack with @chipshort. Chris suggested having one double-ended queue instead of a separate queue and stack. Seems like a good idea - we have no reason for two separate implementations here AFAIK. @webmaster128 @ethanfrey what do you think?

I'm also thinking if we skip having any sort of "stack of stacks" or "StackMap" functionality now, we could:
1. Implement a simple `Deque` type
2. Take time figuring out if we can somehow compose the map types we already have, like `Map<Deque>` and so on.  @itsHaseebSaeed are you looking to use this in some contract of yours?

I'd rather move carefully with these complex storage types. Once we provide an implementation, it will be a bit hard to go back and change it.

itsHaseebSaeed:
> 

Sure I would love to work on Deque; I planned to propose stack and then deque. It all would be a little easier If I could have a list of DO and DON'T.  I DM-ed you on discord. Let's talk there and make this happen. Thanks.

PR #786, opened 2022-08-23
Added Stack storage.

itsHaseebSaeed:
As discussed [here](https://github.com/CosmWasm/cw-plus/issues/776). Added  stack implementation to storage-plus


packages/storage-plus/src/stack.rs
```diff
@@ -0,0 +1,652 @@
...
+    }
+    /// Remove an element from the collection at the specified position.
+    ///
+    /// Removing the last element has a constant cost.
+    /// The cost of removing from the middle/start will depend on the proximity to tail of the list.
+    /// All elements above the specified position will be shifted in storage.
+    ///
+    /// Removing an element from the start (head) of the collection
+    /// has the worst runtime and gas cost.
+    pub fn remove(&self, storage: &mut dyn Storage, pos: u32) -> StdResult<T> {
```
webmaster128, line 149: Removing elements in the middle does not belong to a stack. The only way to remove elements should be pop.

packages/storage-plus/src/stack.rs
```diff
@@ -0,0 +1,652 @@
...
+    }
+    /// Pushes an item to Stack
+    pub fn push(&self, storage: &mut dyn Storage, item: &T) -> StdResult<()> {
+        let len = self.get_len(storage)?;
+        self.set_at_unchecked(storage, len, item)?;
+        self.set_len(storage, len + 1);
+        Ok(())
+    }
+    /// Pops an item from Stack
+    pub fn pop(&self, storage: &mut dyn Storage) -> StdResult<T> {
```
webmaster128, line 132: I think this should return `Option<T>` and nor error in case there is no element left

packages/storage-plus/src/stack.rs
```diff
@@ -0,0 +1,652 @@
...
+
+        let mut may_len = self.length.lock().unwrap();
+        *may_len = Some(len);
+    }
+    /// Clear the collection
+    pub fn clear(&self, storage: &mut dyn Storage) {
+        self.set_len(storage, 0);
+    }
+    /// Replaces data at a position within bounds
+    pub fn set_at(&self, storage: &mut dyn Storage, pos: u32, item: &T) -> StdResult<()> {
```
webmaster128, line 113: I don't think this belongs in a stack

packages/storage-plus/src/stack.rs
```diff
@@ -0,0 +1,652 @@
...
+    /// Pops an item from Stack
+    pub fn pop(&self, storage: &mut dyn Storage) -> StdResult<T> {
+        if let Some(len) = self.get_len(storage)?.checked_sub(1) {
+            let item = self.get_at_unchecked(storage, len);
+            self.set_len(storage, len);
+            item
+        } else {
+            Err(StdError::generic_err("Can not pop from empty Stack"))
+        }
+    }
```
webmaster128, line 140: I think pop also should remove the element from storage

packages/storage-plus/src/stack.rs
```diff
@@ -0,0 +1,652 @@
...
+    item_type: PhantomData<T>,
+}
+
+impl<'a, T: Serialize + DeserializeOwned> Stack<'a, T> {
+    /// constructor
+    pub const fn new(prefix: &'a str) -> Self {
+        Self {
+            namespace: prefix.as_bytes(),
+            prefix: None,
+            length: Mutex::new(None),
```
webmaster128, line 38: What do you need that mutex for? Should not be needed because CosmWasm is a single threaded environment

packages/storage-plus/src/stack.rs
```diff
@@ -0,0 +1,652 @@
...
+    fn set_len(&self, storage: &mut dyn Storage, len: u32) {
+        let len_key = [self.as_slice(), LEN_KEY].concat();
+        storage.set(&len_key, &len.to_be_bytes());
+
+        let mut may_len = self.length.lock().unwrap();
+        *may_len = Some(len);
+    }
+    /// Clear the collection
+    pub fn clear(&self, storage: &mut dyn Storage) {
+        self.set_len(storage, 0);
```
webmaster128, line 110: This should also clear the elements in storage

webmaster128 (commented):
Okay, good start, thanks. I think it's still a bit too complicated becasue the type has too many features. Ideally it just does this: https://en.wikipedia.org/wiki/Stack_(abstract_data_type)#/media/File:Lifo_stack.svg, maybe plus len/is_empty/clear methods. But no change or access of data on the middle. That would be an array or linked list implementation, which is a different story.

Can you also check how the `Map` type works in this repo? It can be used in const places and is just a configuration how to access the contract storage. I think we should do the same here too.

uint:
Nice job including all the tests.

I don't fully understand the use case for `Stack::add_suffix`. It looks like you want `Stack` to be both a stack implementation **and** a manager/factory of `Stack`s, all at once?

itsHaseebSaeed:
> Nice job, including all the tests.
> 
> I don't fully understand the use case for `Stack::add_suffix`. It looks like you want `Stack` to be both a stack implementation **and** a manager/factory of `Stack`s, all at once?

Suffix does what Key() does in Map. Creates new Path

itsHaseebSaeed:
> I'm wondering if it wouldn't make more sense to
> 
> 1. implement a `Stack` that only allows you to have one stack at a fixed prefix, and then
> 2. think about some sort of `StackMap<K, T>` type to aid in managing access to multiple `Stack`s in a similar way `Map` helps in managing access to multiple pieces of data.
> 
> That way the new types have clearer purposes and the use case of "one value defined as a constant to manage multiple stacks" is still achieved.
> 
> With the current `Stack` implementation there might be a name collision issue too if someone actually tries to add new items to both a "root" `Stack` and a "derivative" `Stack`.


That could be done as well. Right now, you can have or not have a suffix, it's your choice. I'll happy to do that as well. Just making sure that everyone is on board with the idea, and then I'll do a coding sprint in the coming days. Just keep on adding any other suggestions and ideas as well.

packages/storage-plus/src/stack.rs
```diff
@@ -0,0 +1,649 @@
...
+            prefix: Some(prefix),
+            length: None,
+            item_type: self.item_type,
+        }
+    }
+}
+
+impl<'a, T: Serialize + DeserializeOwned> Stack<'a, T> {
+    /// gets the length from storage, and otherwise sets it to 0
+    pub fn get_len(&self, storage: &dyn Storage) -> StdResult<u32> {
```
ueco-jb, line 60: nit: Sticking to std's convention:
```suggestion
    pub fn len(&self, storage: &dyn Storage) -> StdResult<u32> {
```

packages/storage-plus/src/stack.rs
```diff
@@ -0,0 +1,649 @@
...
+        Some(Ok(item))
+    }
+
+    // This needs to be implemented correctly for `ExactSizeIterator` to work.
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let len = (self.end - self.start) as usize;
+        (len, Some(len))
+    }
+
+    // I implement `nth` manually because it is used in the standard library whenever
```
ueco-jb, line 280: Please rewrite those comments - refrain from writing in first-person.

```suggestion
    // `nth` is implemented manually, because it is used in the standard library whenever
```

ueco-jb (commented):
I really like the tests.
Please update README with examples and CHANGELOG.
There are plenty of small stuff to go through - please make sure spacing between methods is one line. Currently sometimes there is none.

uint:
Hey @itsHaseebSaeed! Are you still working on this?

itsHaseebSaeed:
> Hey @itsHaseebSaeed! Are you still working on this?

@uint funnily enough, I just added this to my weekly to-do list just today. It's will be done max by Friday then I'll start working on deque store

uint:
Hey, @itsHaseebSaeed! We currently have both a need for this functionality and the resources to develop this in-house. Additionally, we might have some ideas around how to make a "map of stacks" functionality better fit our current system. Rather than spam you with change requests, we'd like to take over the implementation. Thank you for your contribution!

uint:
Closing in favor of #807

PR #788, opened 2022-08-29
Make cw4-group queries public

shanev:
This allows this contract to be extended, including the queries.

Note: Should the version be incremented?

### DIFF
--- a/contracts/cw4-group/src/contract.rs
+++ b/contracts/cw4-group/src/contract.rs
@@ -165,12 +165,12 @@ pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
     }
 }
 
-fn query_total_weight(deps: Deps) -> StdResult<TotalWeightResponse> {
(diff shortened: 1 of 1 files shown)


ueco-jb (approved):
I guess there's no harm in that.

> Should the version be incremented?

Version of what, the contract? It's updated when we're releasing new version of whole `cw-plus`.

webmaster128 (commented):
Can you show how you use this? Making all those internals public is not great IMO. It will make it hard to impossible to do contract internal cleanups like fixing the naming inconsistency between `fn query_member` and `fn list_members`.

Can you go through `query` instead?

webmaster128 (commented):
Okay, let's go with the flow. Can you add the missing `query_` prefix to `list_members`, then merge?

ismellike:
This has been resolved
https://github.com/CosmWasm/cw-plus/pull/802

uint:
Closing, see above comment

PR #792, opened 2022-09-05
Validate allowance expiration

### LINKED ISSUE #628: cw20 allowance expiration can be set to a block height or timestamp in the past
Expirations are not validated when allowances are created,
https://github.com/CosmWasm/cw-plus/blob/fa0c5ac7b3a1575b436c2040c371afd7e68e9c05/contracts/cw20-base/src/allowances.rs#L28

only when they are used https://github.com/CosmWasm/cw-plus/blob/fa0c5ac7b3a1575b436c2040c371afd7e68e9c05/contracts/cw20-base/src/allowances.rs#L95

This is not a huge problem, as the expiration can be updated to a future block height or timestamp, or removed entirely. Should we add validation that the block height or timestamp is in the future when creating a new allowance? Happy to make a PR for this if there is interest.

NB this was found in an audit of Mars protocol in a contract that was forked from the cw-plus cw20-base contract.

chipshort:
Closes #628 

Adds a check for valid expiration when increasing or decreasing allowance.
Also had to change some of the tests because they were creating expired allowances to test the expiration checks of the other message handlers.

### DIFF
--- a/contracts/cw20-base/src/allowances.rs
+++ b/contracts/cw20-base/src/allowances.rs
@@ -24,7 +24,7 @@ pub fn execute_increase_allowance(
         let mut val = allow.unwrap_or_default();
         if let Some(exp) = expires {
             if exp.is_expired(&env.block) {
-                return Err(ContractError::Expired { });
+                return Err(ContractError::Expired {});
             }
             val.expires = exp;
         }
@@ -45,7 +45,7 @@ pub fn execute_increase_allowance(
 
 pub fn execute_decrease_allowance(
     deps: DepsMut,
-    _env: Env,
+    env: Env,
     info: MessageInfo,
     spender: String,
     amount: Uint128,
@@ -71,6 +71,9 @@ pub fn execute_decrease_allowance(
             .checked_sub(amount)
             .map_err(StdError::overflow)?;
         if let Some(exp) = expires {
+            if exp.is_expired(&env.block) {
+                return Err(ContractError::Expired {});
+            }
             allowance.expires = exp;
         }
         ALLOWANCES.save(deps.storage, key, &allowance)?;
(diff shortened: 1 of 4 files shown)


contracts/cw20-base/src/allowances.rs
```diff
@@ -20,9 +20,12 @@ pub fn execute_increase_allowance(
         return Err(ContractError::CannotSetOwnAccount {});
     }
 
-    let update_fn = |allow: Option<AllowanceResponse>| -> StdResult<_> {
+    let update_fn = |allow: Option<AllowanceResponse>| -> Result<_, _> {
         let mut val = allow.unwrap_or_default();
         if let Some(exp) = expires {
+            if exp.is_expired(&env.block) {
```
uint, line 26: You found this helper instead of writing the check yourself. Nice!

contracts/cw20-base/src/allowances.rs
```diff
@@ -20,9 +20,12 @@ pub fn execute_increase_allowance(
...
+                return Err(ContractError::Expired {});
```
uint, line 27: More a suggestion than a blocker. I think what we want is more a validation error, just to be clearer to the user/consumer. The problem isn't that an existing allowance is expired, the problem is the user is trying to create/modify an allowance so that it expires in the past.

I'd add a new error variant here:
https://github.com/CosmWasm/cw-plus/blob/main/contracts/cw20-base/src/error.rs

uint (approved):
Looks good in general, just one suggestion.

Nice tests!

uint:
For some reason CircleCI is not picking up this branch. Let's figure that out...

chipshort:
Closing this, as it is a part of #793 and the ci picked that one up.

PR #793, opened 2022-09-05
Validate allowance expiration

chipshort:
Closes https://github.com/CosmWasm/cw-plus/issues/628

Adds a check for valid expiration when increasing or decreasing allowance.
Also had to change some of the tests because they were creating expired allowances to test the expiration checks of the other message handlers.


uint (approved):

uint:
Thanks!

PR #798, opened 2022-09-13
CI: unified .json schema artifacts for contracts

### LINKED ISSUE #795: Adapt build_and_upload_schemas CI job to new schema format
With the changes in 1.1.0, the full API is in one file. We can show this by getting rid of all the .tar.gz logic in

```yml
      - run:
          name: Build and run schema generator for packages
          command: |
            for S in ./packages/*/examples/schema.rs
            do
              P=$(dirname $S)/..
              echo "Generating schema for $P ..."
              (cd $P && cargo schema --locked && tar -zcf ~/project/schemas/$(basename $(pwd))_schema.tar.gz ./schema)
            done
      - run:
          name: Build and run schema generator for contracts
          command: |
            for C in ./contracts/*/
            do
              echo "Generating schema for $C ..."
              (cd $C && cargo schema --locked && tar -zcf ~/project/schemas/$(basename $(pwd))_schema.tar.gz ./schema)
            done
```

uint:
Closes #795 

Let's test this!


uint:
@webmaster128 I guess to test, I'd push a tag here? We should be able to delete both the tag and the GH release later.

.circleci/config.yml
```diff
@@ -784,7 +784,7 @@ jobs:
             for C in ./contracts/*/
             do
               echo "Generating schema for $C ..."
-              (cd $C && cargo schema --locked && tar -zcf ~/project/schemas/$(basename $(pwd))_schema.tar.gz ./schema)
+              (cd $C && cargo schema --locked && mv ./schema/$(basename $(pwd)).json ~/project/schemas/)
```
webmaster128, line 787:
```suggestion
              (cd $C && cargo schema --locked && mv ./schema/*.json ~/project/schemas/)
```

We don't need to care how it is named here, right?

uint: We don't. Just thought in case something goes wrong, I'd rather risk uploading nothing than random things.

webmaster128:
> I guess to test, I'd push a tag here? We should be able to delete both the tag and the GH release later.

The other option is to comment out the "Publish schemas on GitHub" part and remove the filter for the whole job. That should at least run all commands and you see the "Show data" part.

uint:
> The other option is to comment out the "Publish schemas on GitHub" part and remove the filter for the whole job.

That worked out well. Thanks!

webmaster128 (approved):

PR #802, opened 2022-09-17
SnapshotItem total, public query methods, and safe math

ismellike:
cw4-group should be able to make use of the SnapshotMap's weights by also storing a snapshot of the total.
Here's a relevant example: https://docs.daodao.zone/docs/voting-config

Query methods should be public to allow importing.

Math should use checked arithmetic for safety.


ueco-jb (approved):
:+1:  for introducing safe math here

Rather minor thing, but you didn't test `at_height` variant. Please add some.

ueco-jb:
Nice, extensive tests :+1: Merging

PR #803, opened 2022-09-19
Add clear and is_empty methods to Map

manu0466:
This add two new utility methods to `Map` to make it more similar with the Rust `BTreeMap`
1. clear: Removes all the elements inside `Map`
2. is_empty: Returns `true` if the `Map` don't contains any element

To implement the two methods I rely on the `iterator` feature to get the keys of the elements inside the Map since I haven't find a better way to know the elements currently present inside the `Map`.

### DIFF
--- a/packages/storage-plus/src/map.rs
+++ b/packages/storage-plus/src/map.rs
@@ -109,6 +109,36 @@ where
             from_slice(&result).map(Some)
         }
     }
+
+    /// Clears the map, removing all elements.
+    #[cfg(feature = "iterator")]
+    pub fn clear(&self, store: &mut dyn Storage) {
+        const TAKE: usize = 10;
+        let mut cleared = false;
+
+        while !cleared {
+            let paths = self
+                .no_prefix_raw()
+                .keys_raw(store, None, None, cosmwasm_std::Order::Ascending)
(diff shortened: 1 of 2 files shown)


packages/storage-plus/src/map.rs
```diff
@@ -260,6 +260,41 @@ where
...
+{
+    /// Clears the map, removing all elements.
+    pub fn clear(&self, store: &mut dyn Storage) {
+        const TAKE: usize = 10;
+        let prefix = self.no_prefix_raw();
+        let mut cleared = false;
+
+        while !cleared {
+            let paths = prefix
+                .keys_raw(store, None, None, cosmwasm_std::Order::Ascending)
```
ueco-jb, line 277: Just nitpicking, but IMO it's better to keep it in one place in this case.

```suggestion
        let mut cleared = false;

        while !cleared {
            let paths = self
                .no_prefix_raw()
                .keys_raw(store, None, None, cosmwasm_std::Order::Ascending)
```

manu0466:
@ueco-jb Thanks for the feedbacks! 
I have applied your suggestions and added the implementation of `clear` and `is_empty` to `IndexedMap`.

ueco-jb (approved):
LGTM
Thank you for your contribution.

PR #807, opened 2022-09-26
Add Deque

### LINKED ISSUE #776: Add stack and queue implementations to storage-plus

chipshort:
closes #776

Implements a deque, similar to the std lib `VecDeque`.
This allows having both queue and stack semantics with a single data structure.
I used wrapping math, such that, in case the indexes ever reach `u32::MAX` or you always push to the front, it will still work (as long as it's not filled completely).
No random access at the moment, but can add that, if desired.

The implementation is inspired by #786

### DIFF
--- a/packages/storage-plus/src/deque.rs
+++ b/packages/storage-plus/src/deque.rs
@@ -10,6 +10,9 @@ const TAIL_KEY: &[u8] = b"t";
 const HEAD_KEY: &[u8] = b"h";
 
 /// A deque stores multiple items at the given key. It provides efficient FIFO and LIFO access.
+///
+/// It has a maximum capacity of `u32::MAX - 1`. Make sure to never exceed that number when using this type.
+/// If you do, the methods won't work as intended anymore.
 pub struct Deque<'a, T> {
     // prefix of the deque items
     namespace: &'a [u8],
--- a/packages/storage-plus/src/deque.rs
+++ b/packages/storage-plus/src/deque.rs
@@ -490,7 +490,7 @@ mod tests {
     fn readme_works() -> StdResult<()> {
         let mut store = MockStorage::new();
 
-        // read methods return Option<T>, so None if the deque is empty
+        // read methods return a wrapped Option<T>, so None if the deque is empty
         let empty = DATA.front(&store)?;
         assert_eq!(None, empty);
 
--- a/packages/storage-plus/src/deque.rs
+++ b/packages/storage-plus/src/deque.rs
@@ -348,6 +348,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(feature = "iterator")]
     fn iterator() {
         let deque: Deque<u32> = Deque::new("test");
         let mut store = MockStorage::new();
@@ -373,6 +374,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(feature = "iterator")]
     fn reverse_iterator() {
         let deque: Deque<u32> = Deque::new("test");
         let mut store = MockStorage::new();
@@ -431,6 +433,13 @@ mod tests {
             Some(1),
             "popping should work, even when wrapping"
         );
+    }
+
+    #[test]
+    #[cfg(feature = "iterator")]
+    fn wrapping_iterator() {
+        let deque: Deque<u32> = Deque::new("test");
+        let mut store = MockStorage::new();
 
         deque.set_head(&mut store, u32::MAX);
         deque.set_tail(&mut store, u32::MAX);
@@ -476,6 +485,7 @@ mod tests {
     const DATA: Deque<Data> = Deque::new("data");
 
     #[test]
+    #[cfg(feature = "iterator")]
     fn readme_works() -> StdResult<()> {
         let mut store = MockStorage::new();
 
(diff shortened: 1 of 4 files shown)


packages/storage-plus/src/deque.rs
```diff
@@ -0,0 +1,534 @@
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.start == self.end {
+            return None;
+        }
+
+        let item = self
+            .deque
+            .get_at_unchecked(self.storage, self.start)
+            .transpose()?;
```
uint, line 211: So if a piece of data that should be there is missing, we fail quietly by ending iteration? I wonder if we should produce an error instead. A queue with missing data in the middle sounds bad.

uint (commented):
In general, I really appreciate this PR. It's thorough, some important edge cases are covered, convention is followed, there's plenty of documentation, tests, and there's some clever stuff in there. Nice.

packages/storage-plus/src/deque.rs
```diff
@@ -0,0 +1,534 @@
...
+    }
+
+    /// Returns the first element of the deque without removing it
+    pub fn back(&self, storage: &dyn Storage) -> StdResult<Option<T>> {
+        let pos = self.tail(storage)?.wrapping_sub(1);
+        self.get_at_unchecked(storage, pos)
+    }
+
+    /// Gets the length of the deque.
+    #[allow(clippy::len_without_is_empty)]
```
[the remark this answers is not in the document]
chipshort: Yes, clippy was complaining because of the function signature.

packages/storage-plus/src/deque.rs
```diff
@@ -0,0 +1,534 @@
...
+        // update tail
+        self.set_tail(storage, pos.wrapping_add(1));
+
+        Ok(())
+    }
+
+    /// Adds the given value to the front of the deque
+    pub fn push_front(&self, storage: &mut dyn Storage, value: &T) -> StdResult<()> {
+        // need to subtract first, because head potentially points to existing element
+        let pos = self.head(storage)?.wrapping_sub(1);
```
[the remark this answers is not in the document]
chipshort: Yes, the limit is `u32::MAX`. We could guard against this when pushing by checking whether head and tail are equal and throwing an error. We will have to do an additional read from storage (of head or tail) for that though.
Which one do you prefer? Documenting or checking?

packages/storage-plus/src/deque.rs
chipshort: Good catch, I will produce an error instead

packages/storage-plus/src/deque.rs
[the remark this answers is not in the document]
uint: It looks like `std::collections::VecDeque` will probably panic when grown beyond `usize`. I think we're fine just documenting.

uint (approved):
Nice! Feel free to merge once it's up to date with `main` and tests pass.

webmaster128:
Very nice. Looks like a JavaScript Array now.

Can you explain the reasoning for renaming `Deque` to `VecDeque`? It looks like a [deque](https://en.wikipedia.org/wiki/Double-ended_queue) but it is not based in a vector (in contrast to `std::collections::VecDeque`).

uint:
> Can you explain the reasoning for renaming Deque to VecDeque? It looks like a [deque](https://en.wikipedia.org/wiki/Double-ended_queue) but it is not based in a vector (in contrast to std::collections::VecDeque).

I suggested it. Fair point. I guess I got swayed by the fact it implements indices and random access. It isn't resizable though. Would you prefer something like `ArrDeque`, or just `Deque`?

webmaster128:
> Would you prefer something like ArrDeque, or just Deque?

I vote for just `Deque` – nice and simple.

I think here is no need to add implementation details in the name itself. Rust tends to do that because it matters at the system language level and it allows them to communicate their version is not a linked list, which has certain overhead.

packages/storage-plus/src/deque.rs
```diff
@@ -0,0 +1,636 @@
...
+    T: Serialize + DeserializeOwned,
+{
+    deque: &'a VecDeque<'a, T>,
+    storage: &'a dyn Storage,
+    start: u32,
+    end: u32,
+}
+
+#[cfg(feature = "iterator")]
+impl<'a, T> Iterator for VecDequeIter<'a, T>
```
webmaster128, line 223: The range/iterator implementation is genius.

Do I understand correctly that this even works without the native iterator implementation? I think you can remove `#[cfg(feature = "iterator")]` since you don't rely on storage ranges.

chipshort: Yes, it only needs `storage.get`. Actually, I was wondering about that iterator feature. Are there CosmWasm chains that don't support `storage.range`?

uint: IIRC yes, there are chains that don't support it due to how their storage works. I'm sure Simon or Ethan could elaborate.

uint:
> So, should I make a follow-up PR renaming to Deque and with the iterator feature checks removed?

That would be great!

packages/storage-plus/src/deque.rs
webmaster128: There are underlying storage engines that don't have iterators. The default storage has it but e.g. Secret Network has an encrypted KV storage which you cannot iterate. Also Patricia Merkle Tries (Ethereum) do not support iterating.

PR #814, opened 2022-09-28
Auto-generate CHANGELOG

uint:


uint:
Idea: rather than appending to the committed file (problematic due to how tags aren't technically related to any branch, error-prone), upload a partial CHANGELOG to the description of a release.

PR #815, opened 2022-09-29
feat: add necessary event attributes for cw20 contracts

vinhphuctadang:
[Update]
# What I've done:
- I added some necessary attributes (balances/logo) to events upon instantiate, mint, burn, transfer, send, transfer_from, update_log, send_from. Those are values that offchain services need to understand about the contract

# Motivation
- For offchain services, we are indexing based on events, and the events should describe what happened and the result.
E.g: 
  - User A transfers B `n` token -> we should know how much A and B have after the transfer tx. So we need necessary attribute lkike sender_balance and receiver_balance.
- Why don't we use smart query to the contract?
  - Anyone can create their own cw20 contract, query might cost a lot more resource (cpu, mem)
  - Too many query to the node might make node choke for some time:
Assume we have 1k accounts in the contract, and each query cost ~1+ms, I attack the test node with many smart query and see node stuck, can't produce some blocks

```
query token info time: 172.3335ms
query balance time: 1.026708ms
query balance time: 485.417µs
query balance time: 900.459µs
```

-> The solution is emitting enough info for indexer services, data can be stored into db, scalable too


contracts/cw20-base/Cargo.toml
```diff
@@ -28,6 +28,7 @@ schemars = "0.8.1"
 semver = "1"
 serde = { version = "1.0.103", default-features = false, features = ["derive"] }
 thiserror = { version = "1.0.23" }
+serde_json = "1.0.85"
```
uint, line 31: We specifically avoid using `serde_json` in our code and use [`serde_json_wasm`](https://github.com/CosmWasm/serde-json-wasm) instead. This is for efficiency and determinism as far as I recall.

contracts/cw20-base/src/contract.rs
```diff
@@ -153,7 +167,13 @@ pub fn instantiate(
         MARKETING_INFO.save(deps.storage, &data)?;
     }
 
-    Ok(Response::default())
+    let serialized_balances = serde_json::to_string(initial_balances).unwrap();
```
uint, line 170: I don't feel the benefit of doing this beats the cost.

uint (commented):
Hi! Thank you for the time spent working on this.

> add necessary event attributes for cw20 contracts

Necessary for what?

contracts/cw20-base/src/contract.rs
```diff
@@ -248,25 +268,29 @@ pub fn execute_transfer(
...
     )?;
 
     let res = Response::new()
         .add_attribute("action", "transfer")
-        .add_attribute("from", info.sender)
+        .add_attribute("from", sender_addr.as_str())
         .add_attribute("to", recipient)
-        .add_attribute("amount", amount);
+        .add_attribute("amount", amount)
+        .add_attribute("sender_balance", sender_balance)
```
uint, line 292: I think in general we tend to have events describe the action taken, not the outcome. That would get out of hand quickly.

contracts/cw20-base/Cargo.toml
vinhphuctadang: Thanks, I will update

contracts/cw20-base/src/contract.rs
```diff
@@ -128,14 +131,25 @@ pub fn instantiate(
...
+                Logo::Url(url) => {
+                    logo_string.push_str(url.as_str());
+                    Some(LogoInfo::Url(url))
+                },
+                Logo::Embedded(binary_type) => {
+                    let b64_logo = match binary_type {
+                        EmbeddedLogo::Svg(binary) => binary.to_base64(),
+                        EmbeddedLogo::Png(binary) => binary.to_base64(),
+                    };
+                    logo_string.push_str((logo_type + ":" + b64_logo.as_str()).as_str());
```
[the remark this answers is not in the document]
vinhphuctadang: 5KB limited so I think it's fine;

contracts/cw20-base/src/contract.rs
vinhphuctadang: Thanks for all reviews
With the outcome, we will index it easier in offchain services; 
Assume that we have only from, to, amount, we won't know what's the target balance each account, we have to query the chain each time event happens, but it is not good

contracts/cw20-base/src/contract.rs
vinhphuctadang: Will update this code

contracts/cw20-base/src/contract.rs
uint: Why is it bad? Queries don't cost gas.

contracts/cw20-base/src/contract.rs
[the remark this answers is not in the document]
uint: Doesn't this basically mean those 5KB are stored on the chain twice? Once in contract storage, second time in the events log.

So you have an added gas cost and extra storage used. If you just perform a contract query after the tx, you have none of that.

contracts/cw20-base/src/contract.rs
uint: If we model this after the `send` messages in the Cosmos SDK bank module, there's minimal information there too:

https://docs.cosmos.network/master/modules/bank/section_04.html

I keep looking up the topic and it keeps occurring to me that events are not meant to provide full information for clients; they're meant to help block explorers and other tools index txs. [See here](https://docs.cosmos.network/master/core/events.html).

vinhphuctadang: > Why is it bad? Queries don't cost gas.
I agree that queries won't cost gas; but for some off-chain indexer/FE, querying a lot may cause nodes to be choked. So emit event on contract is a good choice for off-chain indexers, scaling off-chain services for querying is much easier than adding more nodes I think

vinhphuctadang: > https://docs.cosmos.network/master/modules/bank/section_04.html
The links are dead somehow xD

Anyway I understand that we can query the node, which is also good for user end; for some applications like what I'm working on, the outcome is better because we will know exactly what the result is after txs happened

contracts/cw20-base/src/contract.rs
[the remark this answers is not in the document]
vinhphuctadang: Ah yes you're right, it doubled gas cost, I removed in event to save some gas consumption;
I think only allowing url is better since it's more efficient, few bytes compare to large logo binary files, just an idea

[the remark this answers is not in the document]
uint: Well. What problem are we solving here? I would assume it's not a huge problem if an offchain service has to query for the logo of a currency here and there, and I would assume an offchain service would cache logos for various currencies.

There are events with `"action", "upload_logo"` that the offchain service can track to know when to update its cache.

vinhphuctadang:
> Hey @vinhphuctadang, could you describe what you're working on? It will be easier for me if I understand what problems you're trying to solve.
> 
> If your software is supposed to work with **any** cw20 contracts, I worry relying on this particular implementation of the `cw20` protocol isn't a good idea. This isn't the only `cw20` implementation in existence.

Yes, I'm indexing those events in an off-chain service serving chain explorer. And atm querying the node is not a good idea for the service since it might consume much resources for executing contract queries. Events are included in block result and can be read while index.

Can we discuss more via Discord, I think it will be more realtime. Mine: Phuc Ta#0615

uint:
> Can we discuss more via Discord, I think it will be more realtime. Mine: Phuc Ta#0615

I added you on Discord. My handle there is stratcat.

> Yes, I'm indexing those events in an off-chain service serving chain explorer. And atm querying the node is not a good idea for the service since it might consume much resources for executing contract queries. Events are included in block result and can be read while index.

Does your service have a caching layer between accepting user requests and fetching data from a blockchain node?

uint:
Alright, so while these might be good improvements, we can't really merge them.

* There are hundreds of `cw20` contracts already deployed in the wild.
* We cannot apply these changes retrospectively to transactions that already happened.
* We cannot expect old contracts to migrate to the new format. Migrations are a tricky chore.
* By merging this, we'd be suggesting to future block explorers and wallets they can rely on this new format of events. They cannot. The de facto standard is already set and used.

We're going to document events emitted in the specs soon, and we're going to make it official the specs are frozen and immutable.

The one thing that can be done if there's enough justification would be to create a new standard, separate from `cw20`, but for that to be useful you'd need to convince the ecosystem it's significant enough to migrate to.

Closing this PR as I don't see any way forward.

PR #817, opened 2022-09-29
Remove storage-plus dependency from storage-macro

### LINKED ISSUE #808: Remove circular dependency between `storage-macro` and `storage-plus`
Documentation can live next to the `storage-plus` reexport. Tests can live somewhere in `storage-plus`. That way there's no need for the `macro` crate to rely on `storage-plus`, or for contract devs to manage `storage-macro` as an explicit dependency in their projects. This would be more idiomatic.

chipshort:
The tests that were depending on `storage-plus` are now in `storage-plus`. I also changed the CI to actually run them.

I added the docs from readme to the actual macro. I'm not sure how docs.rs works in this regard though, since the reexport is only available with the `macro` feature.

closes #808

### DIFF
--- a/packages/storage-macro/Cargo.toml
+++ b/packages/storage-macro/Cargo.toml
@@ -16,6 +16,5 @@ proc-macro = true
 syn = { version = "1.0.96", features = ["full"] }
 
 [dev-dependencies]
-cw-storage-plus = { version = "<=0.15.1, >=0.14.0", path = "../storage-plus" }
 cosmwasm-std = { version = "1.1.0", default-features = false }
 serde = { version = "1.0.103", default-features = false, features = ["derive"] }
--- a/packages/storage-macro/src/lib.rs
+++ b/packages/storage-macro/src/lib.rs
@@ -5,6 +5,25 @@ use syn::{
     parse_macro_input, ItemStruct,
 };
 
+/// Auto generate an `IndexList` impl for your indexes struct.
+///
+/// # Example
+///
+/// ```rust
+/// #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+/// struct TestStruct {
+///     id: u64,
+///     id2: u32,
+///     addr: Addr,
+/// }
+///
+/// #[index_list(TestStruct)] // <- Add this line right here.
+/// struct TestIndexes<'a> {
+///     id: MultiIndex<'a, u32, TestStruct, u64>,
+///     addr: UniqueIndex<'a, Addr, TestStruct>,
+/// }
+/// ```
+///
 #[proc_macro_attribute]
 pub fn index_list(attr: TokenStream, item: TokenStream) -> TokenStream {
     let input = parse_macro_input!(item as ItemStruct);
--- a/packages/storage-macro/src/lib.rs
+++ b/packages/storage-macro/src/lib.rs
@@ -5,25 +5,6 @@ use syn::{
     parse_macro_input, ItemStruct,
 };
 
-/// Auto generate an `IndexList` impl for your indexes struct.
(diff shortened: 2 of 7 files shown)


uint:
> I added the docs from readme to the actual macro. I'm not sure how docs.rs works in this regard though, since the reexport is only available with the `macro` feature.

Good catch. You should be able to test the docs locally by running `cargo doc --features ...`

You could also add some metadata to `Cargo.toml` to make the docs on docs.rs include those features:
https://docs.rs/about/metadata

chipshort:
Yes, locally it was working with `--all-features` or `--features macro`, so should I just add this?
```toml
[package.metadata.docs.rs]
all-features = true
```

uint:
Yep, that's fine

packages/storage-macro/src/lib.rs
```diff
@@ -5,6 +5,25 @@ use syn::{
...
+///     addr: Addr,
+/// }
+///
+/// #[index_list(TestStruct)] // <- Add this line right here.
+/// struct TestIndexes<'a> {
+///     id: MultiIndex<'a, u32, TestStruct, u64>,
+///     addr: UniqueIndex<'a, Addr, TestStruct>,
+/// }
+/// ```
+///
```
uint, line 26: Ah, so normally macro docs would live in the `storage-plus` crate, above the reexport. [Here's an example.](https://github.com/CosmWasm/cosmwasm/blob/main/packages/schema/src/lib.rs#L12-L36)

Does this example pass when tested?

packages/storage-macro/Cargo.toml
```diff
@@ -16,6 +16,5 @@ proc-macro = true
 syn = { version = "1.0.96", features = ["full"] }
 
 [dev-dependencies]
-cw-storage-plus = { version = "<=0.15.1, >=0.14.0", path = "../storage-plus" }
```
uint, line 19: Just FYI: the way the version requirements are specified here is a hack to make a cyclic dependency publishable to crates.io. Good thing we don't actually need it!

packages/storage-macro/src/lib.rs
chipshort: Good point, it does not. I'll move it to the reexport

uint (approved):
LGTM. Thanks for making sure those tests are actually run in CI

packages/storage-plus/src/lib.rs
```diff
@@ -42,9 +42,34 @@ pub use prefix::{range_with_prefix, Prefix};
...
+/// #[index_list(TestStruct)] // <- Add this line right here.
+/// struct TestIndexes<'a> {
+///     id: MultiIndex<'a, u32, TestStruct, u64>,
+///     addr: UniqueIndex<'a, Addr, TestStruct>,
+/// }
+/// ```
+///
+pub use cw_storage_macro::index_list;
+#[cfg(all(feature = "iterator", feature = "macro"))]
 pub use cw_storage_macro::*;
```
uint, line 75: Ah, I think we can get rid of this!

chipshort: I added it in case we forget to reexport new items that might get added to the macro crate in the future, but I guess we'd have to bump at least the version anyway in that case.

uint: Yeah. This could also cause us to reexport something we didn't want to by accident. And we want to document everything we reexport from the macro crate anyway.

packages/storage-macro/Cargo.toml
maurolacy: Good that this is not needed, yes. Thanks for working on simplifying this.

PR #820, opened 2022-10-01
cw20-ics20 'Amount' address validation and tests

M-Daeva:
cw20-ics20
1. Added address verification for Amount::Cw20
2. Added tests for Amount::from_parts

### DIFF
--- a/contracts/cw20-ics20/src/amount.rs
+++ b/contracts/cw20-ics20/src/amount.rs
@@ -1,6 +1,6 @@
 use crate::error::ContractError;
 use cosmwasm_schema::cw_serde;
-use cosmwasm_std::{testing::MockApi, Api, Coin, Uint128};
+use cosmwasm_std::{Addr, CanonicalAddr, Coin, StdError, StdResult, Uint128};
 use cw20::Cw20CoinVerified;
 use std::convert::TryInto;
 
@@ -14,7 +14,7 @@ impl Amount {
     pub fn from_parts(denom: String, amount: Uint128) -> Self {
         if denom.starts_with("cw20:") {
             Amount::Cw20(Cw20CoinVerified {
-                address: MockApi::default()
+                address: LocalApi::default()
                     .addr_validate(denom.get(5..).unwrap())
                     .unwrap(),
                 amount,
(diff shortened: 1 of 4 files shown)


uint:
Hi! Thank you for the submission.

How is this `LocalApi` different from `cosmwasm_std::testing::MockApi`? At a glance it seems to do the exact same thing.

uint:
I also don't think the logic we use in our mocks is supposed to be used in a production setting. I suspect this won't validate an address properly at all.

contracts/cw20-ics20/src/amount.rs
```diff
@@ -1,30 +1,32 @@
...
-use cw20::Cw20Coin;
+use cosmwasm_std::{Addr, CanonicalAddr, Coin, StdError, StdResult, Uint128};
+use cw20::Cw20CoinVerified;
 use std::convert::TryInto;
 
 #[cw_serde]
 pub enum Amount {
     Native(Coin),
-    // FIXME? USe Cw20CoinVerified, and validate cw20 addresses
-    Cw20(Cw20Coin),
```
ethanfrey, line 11: I wouldn't do this. You can see the problem with trying to parse a string and suddenly needing `Deps` to do the verification.

I would rather keep Amount as is, and then make an `VerifiedAmount` type, with

```rust
impl Amount {
  pub fn verify(self, deps: Deps) -> StdResult<VerifiedAmount> { /* ... */ }
}
```

Moving parsing and verifying into two steps.

contracts/cw20-ics20/src/amount.rs
```diff
@@ -64,3 +66,158 @@ impl Amount {
...
+            Amount::from_parts(cw20_denom.into(), amount),
+            Amount::Cw20(Cw20CoinVerified {
+                address: cw20_addr,
+                amount
+            })
+        );
+    }
+
+    #[test]
+    #[should_panic]
```
ethanfrey, line 216: Definitely panic is also bad, which is why changing API and returning Result on a verify method is cleaner

uint:
Closing due to inactivity

PR #826, opened 2022-10-12
Generic query for cw3 unification

hashedone:
Small cw3 message update to be usable for external chains


uint (approved):
Good catch, thanks!

uint:
Note: we're having an issue with coverage, possibly because tarpaulin version isn't pinned. I'll deal with it once we have a solution.

PR #835, opened 2022-10-19
Remove the AGPL license

### LINKED ISSUE #833: Simplify licenses
Looks like AGPL is not actually used for anything anymore. Let's get rid of it and simplify the license story.

uint:
Closes #833


uint:
[`cosmwasm` has this Jehan Tremback fellow in its `NOTICE` file.](https://github.com/CosmWasm/cosmwasm/blob/main/NOTICE#L1)

Should we include him here too?

webmaster128 (approved):
LGTM but want to hear Ethan's thoughts as he started the current licensing schema

NOTICE
```diff
@@ -1,4 +1,15 @@
 CosmWasm-Plus: A collection of production-quality CosmWasm contracts
 
-Different crates in this repo may be either Apache-2.0 or AGPL-3.0 licensed.
-Please refer to the NOTICE file in each crate (and the Cargo.toml metadata)
\ No newline at end of file
+Copyright 2021 Confio GmbH
```
ueco-jb, line 3: Is it?

ethanfrey:
> LGTM but want to hear Ethan's thoughts as he started the current licensing schema

I requested this change. We were going to include a mix of things (tools + more product focused ones). In the end it is examples and tools and everything is already Apache. Just the warning that "something may be AGPL"

Let's clean it up as very package here is already Apache

NOTICE
uint: What do you mean? I think the idea with these attributions is to let people know whose intellectual property this is, but I never dived deep into that.

ueco-jb: I was wondering if project being under CosmWasm org isn't somehow under it as well from copyright standpoint - in previous license there copyright wasn't mentioned at all..
But as an intellectual property attribute, sure - I see no issues.

uint: Ah, I see! I figured Confio is the only legal entity we have, and in the [`cosmwasm` repo we state the company as the copyright holder](https://github.com/CosmWasm/cosmwasm/blob/main/NOTICE#L4).

@webmaster128 what do you think?

webmaster128: The whole point here is to list the copyright holders, which then at the same time grant a license to use the software. Strictly speaking every freelancer who's not employed could be listed individually. The list might not be exhaustive, but its a good start. More elements might be added as in the cosmwasm example.

The only change I'd do for now is change the year to a list ("2021,2022 Confio GmbH").

webmaster128 (approved):
🐎

Issue #842, opened 2022-11-03
adding supported interfaces value to cw2


eledra89:
Following some of the discussion on Twitter: https://twitter.com/EledraNguyen/status/1587640345701281792

### Motivation
There should be an easy way to detect what interfaces a cosmwasm contract implements (similar to EIP-165). There are several situations both on-chain and off-chain for such feature:

- An indexer wants to categorize different types of contract.
- A multichain application (wallet, marketplace, etc.) that already support ERC-165 from EVM chain wants to render different UX for different type of contracts.
- A contract wants different ways to interact with other contracts depending on their types.

### Implementation
We propose to add an optional map containing supported interfaces in the `ContractVersion` struct for these values.
The supported interface value should be a string of pre-defined format: <Crate Registry Domain>:<Crate Name>

For example: ["crates.io:cw721","crates.io:cw2"]

### Notes
There is no way to enforce the string format or forcing the contract to have actual implementation of the declared interfaces. Thus, this variable is entirely optional for both the implementor as well as the caller.

ethanfrey:
Interesting idea.
it will have to be optional, as it will not exist for all currently deployed contracts but does seem like a nice hint.

We also discussed adding source code link, repeatable compilation Metadata, etc in a wasmd issue. This falls into that general category for me.

@uint I would love to hear you opinions here as well

eledra89:
Yes, it is optional. We have opened a pull request on it.

For source code link, binaries, etc. I think the block explorer should be the place to submit those information (like Etherscan did). I think they are just informative data while the supported interface value is quite helpful in interacting with other program.

We have built a small utility on our explorer for general information of a wasm contract here: https://github.com/aura-nw/contract-source-code-verifier.

uint:
> @uint I would love to hear you opinions here as well

If we want contracts to self-describe their interfaces and we're trying to make things modular with [Sylvia](https://github.com/CosmWasm/sylvia/tree/main/sylvia), it makes total sense IMO to be able to reason which common interfaces (cw20 etc) they implement.

The cool thing is Sylvia might be able to autogenerate these and it'd fit right into its model.

eledra89:
So what is the general consensus on this ? @uint @ethanfrey 

Should we modify the current cw2 interface or we make a new cw2-support-interface crate ?

uint:
> Should we modify the current cw2 interface or we make a new cw2-support-interface crate ?

[...]

PR #844, opened 2022-11-03
Cw2 add supported interface

eledra89:
Implement #842

### DIFF
--- a/packages/cw2/src/lib.rs
+++ b/packages/cw2/src/lib.rs
@@ -33,6 +33,14 @@ pub struct ContractVersion {
     /// the only code that needs to understand the version parsing is code that knows how to
     /// migrate from the given contract (and is tied to it's implementation somehow)
     pub version: String,
+    /// supported_interface is an optional parameter returning a vector of string represents interfaces 
+    /// that the contract support The string value is the interface crate names in Rust crate Registry. 
+    /// This parameter is inspired by the EIP-165 from Ethereum.
+    /// Each string value should follow a common standard such as <Registry Domain>:<Crate Name>  
+    /// e.g ["crates.io:cw721","crates.io:cw2"]
+    /// NOTE: this is just a hint for the caller to adapt on how to interact with this contract.
+    /// There is no guarantee that the contract actually implement these interfaces.
+    pub supported_interface: Option<Vec<String>>,
 }
 
 /// get_contract_version can be use in migrate to read the previous version of this contract
@@ -46,10 +54,12 @@ pub fn set_contract_version<T: Into<String>, U: Into<String>>(
     store: &mut dyn Storage,
     name: T,
     version: U,
+    supported_interface: Option<Vec<String>>,
 ) -> StdResult<()> {
     let val = ContractVersion {
         contract: name.into(),
         version: version.into(),
+        supported_interface: supported_interface.into(),
     };
     CONTRACT.save(store, &val)
 }
@@ -78,6 +88,7 @@ where
 mod tests {
     use super::*;
     use cosmwasm_std::testing::MockStorage;
+    use std::vec::Vec;
 
     #[test]
     fn get_and_set_work() {
@@ -89,12 +100,27 @@ mod tests {
         // set and get
         let contract_name = "crate:cw20-base";
         let contract_version = "0.2.0";
-        set_contract_version(&mut store, contract_name, contract_version).unwrap();
+        set_contract_version(&mut store, contract_name, contract_version,None).unwrap();
 
         let loaded = get_contract_version(&store).unwrap();
         let expected = ContractVersion {
             contract: contract_name.to_string(),
             version: contract_version.to_string(),
+            supported_interface: None,
+        };
+        assert_eq!(expected, loaded);
+
+        // set and get with supported_interface
+        let contract_name = "crate:cw20-base";
+        let contract_version = "0.2.0";
+        let supported_interface = Some(Vec::from(["crates.io:cw2".to_string(),"crates.io:cw721".to_string()]));
+        set_contract_version(&mut store, contract_name, contract_version,supported_interface.clone()).unwrap();
+
+        let loaded = get_contract_version(&store).unwrap();
+        let expected = ContractVersion {
+            contract: contract_name.to_string(),
+            version: contract_version.to_string(),
+            supported_interface: supported_interface.clone(),
         };
         assert_eq!(expected, loaded);
     }


packages/cw2-supported-interface/README.md
```diff
@@ -0,0 +1,67 @@
...
+    pub version: String,
+    /// supported_interface is an optional parameter returning a vector of string represents interfaces
+    /// that the contract support The string value is the interface crate names in Rust crate Registry.
+    /// This parameter is inspired by the EIP-165 from Ethereum.
+    /// Each string value should follow a common standard such as <Registry Domain>:<Crate Name>
+    /// e.g ["crates.io:cw721","crates.io:cw2"]
+    /// NOTE: this is just a hint for the caller to adapt on how to interact with this contract.
+    /// There is no guarantee that the contract actually implement these interfaces.
+    pub supported_interface: Option<Vec<String>>,
+}
```
uint, line 56: If we're going "separate spec", it'd probably be better if this only had the `supported_interface` field. (Maybe we don't even need the encompassing struct, or it can be a transparent newtype.) This doesn't need to be optional then. Contracts can then include both `cw2` and this new spec to provide the full data.

packages/cw2-supported-interface/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+
+For more information on this specification, please check out the
+[README](https://github.com/CosmWasm/cw-plus/blob/main/packages/cw2/README.md).
+*/
+
+use cosmwasm_schema::cw_serde;
+use cosmwasm_std::{CustomQuery, QuerierWrapper, QueryRequest, StdResult, Storage, WasmQuery};
+use cw_storage_plus::Item;
+
+pub const CONTRACT: Item<ContractVersion> = Item::new("contract_info");
```
uint, line 24: If we're going "separate spec", let's make this a different key from `cw2` so they can both be used together.

packages/cw2-supported-interface/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "cw2-supported-interface"
```
uint, line 2: I think we can name it `cw22` or something along those lines, to follow the current convention.

packages/cw2-supported-interface/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+    };
+    CONTRACT.save(store, &val)
+}
+
+/// This will make a raw_query to another contract to determine the current version it
+/// claims to be. This should not be trusted, but could be used as a quick filter
+/// if the other contract exists and claims to be a cw20-base contract for example.
+/// (Note: you usually want to require *interfaces* not *implementations* of the
+/// contracts you compose with, so be careful of overuse)
+pub fn query_contract_info<T, CQ>(
```
uint, line 72: We could have two queries:
- query all available interfaces
- query if a specified interface is supported

uint (commented):
Looks useful. Nice!

I left some comments. In general if we're doing a separate spec, I'd prefer to make it independent from `cw2` rather than a superset.

peara:
Hi, since this is a new package, I think we could add more functionalities into it.
For example, instead of a simple Vec, we could use a Map

peara:
Hi @eledra89  @uint @ThienLK1. Since this is a new package, I think we could add more functionalities.
For example, instead of a Vec stored in a Item, we could use Map which map an **interface_name** to a **version**:
`const SUPPORTED_INTERFACES: Map<&str, &str> = Map::new("supported_interfaces");`
With that, we can support smart query like `check_supported_interface(interface, version)` which would make using this much more convenient.

eledra89:
also cc @shanev per our discussion on Twitter. Can't cc Larry here though

ThienLK1:
> Hi @eledra89 @uint @ThienLK1. Since this is a new package, I think we could add more functionalities. For example, instead of a Vec stored in a Item, we could use Map which map an **interface_name** to a **version**: `const SUPPORTED_INTERFACES: Map<&str, &str> = Map::new("supported_interfaces");` With that, we can support smart query like `check_supported_interface(interface, version)` which would make using this much more convenient.

@peara  I just merge a new pull request, add more function to define and check version of supported interfaces, please take a look!

packages/cw22/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "cw22"
+version = "1.0.0"
+authors = ["ThienLK1@aura-nw"]
+edition = "2021"
+description = "Definition and types for the CosmWasm-2 interface"
```
peara, line 6: please update name

packages/cw22/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+homepage = "https://cosmwasm.com"
+
+[dependencies]
+cosmwasm-schema = "1.0.0"
+cosmwasm-std = { version = "1.0.0", default-features = false }
+cw-storage-plus = "0.16.0"
+schemars = "0.8.1"
+serde = { version = "1.0.0", default-features = false, features = ["derive"] }
+thiserror = { version = "1.0.23" }
+version-compare = "0.1"
```
peara, line 18: is this package needed?

packages/cw22/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+    /// e.g  "v0.16.0"
+    pub version: String,
+}
+
+/// set_contract_supported_interface should be used in instantiate to store the original version
+pub fn set_contract_supported_interface(
+    store: &mut dyn Storage,
+    mut supported_interfaces: Vec<ContractSupportedInterface>,
+) -> Result<(), StdError> {
+    while let Some(supported_interface) = supported_interfaces.pop() {
```
peara, line 40: can we use `for supported_interface in supported_interfaces {}` here?

packages/cw22/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+    interface: String,
+    query: String,
+) -> StdResult<VersionResponse> {
+    let req = VersionReq::parse(&query).unwrap();
+    let supported_version_rs = SUPPORTED_INTERFACES
+        .may_load(store, interface)?
+        .unwrap_or_default();
+    let supported_version = Version::parse(&supported_version_rs);
+    match supported_version {
+        Ok(ver) => Ok(VersionResponse {
```
peara, line 76: since this is a query, I think just return true or false would be enough.

packages/cw22/src/lib.rs
ThienLK1: yes, the code will be similar.

packages/cw22/src/lib.rs
ThienLK1: it returns more information to the querier. Example which specifies version interface is supported, or not?

packages/cw22/Cargo.toml
ThienLK1: unused packages have been removed

eledra89:
@uint @ethanfrey , we added cw22: supported interface with additional support for version checking.

ethanfrey:
Nice. I took a quick look over the comments and agree with the direction.

Using a separate independent interface, as well as using a Map for easier checks.

I want to say that one nice part of cw2 is that it has a well-defined storage structure, so you can do "raw queries" against it if needed. Making the storage layout part of this spec also sounds like a good idea. Sure, we can do smart queries, but that depends on a higher-level API and is less efficient. A well defined storage layer allows efficient raw queries with no more code on the queried contract.

We also have helpers for this, a calling contract can do something like:

`let version = SUPPORTED_INTERFACES.query(contract_addr, "crates.io:cw20")?;`

I will go through briefly on the code, but this is my high-level input.

packages/cw22/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+**Required**
+
+All CW22-compliant contracts must store the following data:
+
+* key: `contract_supported_interface`
+* data: Json-serialized `ContractSupportedInterface`
+
+```rust
+pub struct ContractSupportedInterface {
```
ethanfrey, line 19: Nit: this is different than implementation, please update

packages/cw22/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+mod query;
+
+use cosmwasm_schema::cw_serde;
+use cosmwasm_std::StdError;
+use cosmwasm_std::{StdResult, Storage};
+use cw_storage_plus::Map;
+use query::VersionResponse;
+use semver::{Version, VersionReq};
+
+pub const SUPPORTED_INTERFACES: Map<String, String> = Map::new("supported_interfaces");
```
ethanfrey, line 24: Please use `Map<&str, String>`, it is more efficient and easier to call

packages/cw22/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+    /// semantic version on release tags of the interface package following SemVer guideline.
+    /// e.g  "0.16.0"
+    pub version: String,
+}
+
+/// set_contract_supported_interface should be used in instantiate to store the original version
+/// of supported interfaces. It should also be used after every migration.
+pub fn set_contract_supported_interface(
+    store: &mut dyn Storage,
+    mut supported_interfaces: Vec<ContractSupportedInterface>,
```
ethanfrey, line 45: If we just use &str in the Map, this cheaply becomes: `supported_interfaces: &[ContractSupportedInterface]`

packages/cw22/src/lib.rs
ethanfrey: Argument change would force this for loop here, which is better anyway

packages/cw22/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+) -> Result<(), StdError> {
+    while let Some(supported_interface) = supported_interfaces.pop() {
+        let id = supported_interface.supported_interface;
```
ethanfrey, line 48: Why not just use a shorter local variable than "supported_interface" and inline this.  The following should be clear enough (but change x to item or so if you prefer)

```rust
for x in supported_interfaces {
  SUPPORTED_INTERFACES.save(store, &x.supported_interface, &x.version)?;
}
```

packages/cw22/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+        let version = supported_interface.version;
+        SUPPORTED_INTERFACES.save(store, id, &version)?;
+    }
+    Ok(())
+}
+
+/// query_supported_interface_version show the version of an interface supported by the contract
+pub fn query_supported_interface_version(
+    store: &dyn Storage,
+    interface: String,
```
ethanfrey, line 58: Again make this take &str. Avoid ownership in all arguments unless needed internally.

packages/cw22/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+) -> StdResult<ContractSupportedInterface> {
```
ethanfrey, line 59: I would return `StdResult<Option<String>>` here. Unless there is a read error, it returns `None` for not registered, and the version if registered.

packages/cw22/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+    let version = SUPPORTED_INTERFACES
+        .may_load(store, interface.clone())?
+        .unwrap_or_default();
```
ethanfrey, line 62: We *do* want the difference between None and Some here (not just "")

packages/cw22/src/lib.rs
ethanfrey: I think this function can be removed.

`query_supported_interface_version` already returns the version (or missing), be DRY.
Also, we will most likely be getting this though `Map.query(...)` as we remotely query a contract (don't compare with our own contracts version). Let's make the version comparison helper separate.

I would just add two functions that don't touch storage and show the semver usage, like:

// most comparisons just require a minimum - make this easy
`minimum_version(version: &str, required: &str) -> bool {}` - maybe make `Result<(), Error>`?

// allow more powerful query strings - ">=1.2.3,<2"
`require_version(version: &str, request: &str) -> bool {}` - or Result as above

packages/cw22/src/lib.rs
ethanfrey: Actually, since we assume all stored versions are semver later on in queries, enforce it here.
Parse the values as Version and error if they are invalid - much better to error on instantiate (fail fast in tests) than later when first querying in production.

packages/cw22/src/query.rs
```diff
@@ -0,0 +1,8 @@
+use cosmwasm_schema::cw_serde;
+
+#[cw_serde]
+pub struct VersionResponse {
```
ethanfrey, line 4: Let's remove this. I don't define smart queries for this, they require more wiring.

Let's be like cw2 (and cw4) and focus on spec'ing the storage layout for efficient queries. The contract should be able to easily write its own storage. Other contracts will be querying it's storage (Map.query) and testing the resulting value.

packages/cw22/src/lib.rs
ethanfrey: If you go for the Result type, please define a custom Cw22Error type in the spec.

ethanfrey (changes requested):
A number of comments on the code.

I like the direction, but please clean this up so it is up to standards to merge

eledra89:
oh wow this is super detail @ethanfrey, we are kind of new to Rust & CosmWasm so this is very helpful in the long term !

packages/cw22/src/lib.rs
peara: @ethanfrey for the `minimum_version`, you mean a helper function like this?
```
pub fn minimum_version(version: &str, required: &str) -> bool {
  let req = VersionReq::parse(">={required}").unwrap();
  let version = Version::parse(version);
  req.matches(&version)
}
```

shanev:
Would love to get this merged in @eledra89 if you get a chance to resolve the PR comments.

hashedone (approved):
The idea is great, and the solution is pretty good. Still, I would strongly advise migrating from `String` to `Cow` - I know we do `String` in most places in cw, but we already tried to use `Cow` where possible in Sylvia. We should go for that in interfaces, particularly those which we want to be implemented by most of the contracts to reduce extra gas cost on those small things (I get it is only for instantiation/migration, but it is very much a free change).

@jawoznia please take a look at that - we probably will auto-support this in Sylvia in the future.

I give approval as the `Cow` thing is not super-critical, but if you can please align. If not, please at least create an issue out of my comment.

peara:
@hashedone I have updated with `Cow`.
I'd like to add some examples of `QueryRaw` like in cw4 but haven't been able to finished.
I also want to add this to cw20-base as well.
So maybe in a follow up PR, I will add this to cw20-base and add some QueryRaw examples as test there.

hashedone (approved):
LGTM with minor docs comment.

peara:
> Sorry for the annoying nitpicking here. We struggle with documentation quality, and we don't want to introduce any more dept in this area.
@hashedone No problem, I actually forgot that, would have hate it myself.

PR #845, opened 2022-11-23
Standardize spec events

### LINKED ISSUE #823: Standardize protocol events
As far as I understand, our `README.md` files like [this one](https://github.com/CosmWasm/cw-plus/blob/main/packages/cw20/README.md) are protocol specs. Events emitted don't seem to be part of the protocol spec, and yet [client software like a cw20 wallet or block explorer might rely on the format of emitted events to index transactions](https://docs.cosmos.network/main/core/events).

It might be tempting to change the events emitted in a concrete implementation like [`cw20-base`](https://github.com/CosmWasm/cw-plus/tree/main/contracts/cw20-base) to suit clients, but the whole point of having a specification is to assume `cw20-base` isn't the only implementation. IMO any changes to the protocol should be well communicated and start with the specification, not our "minimal examples".

uint:
Closes #823

Just setting the event attributes in stone so that the APIs are clear

### DIFF
--- a/packages/cw1/README.md
+++ b/packages/cw1/README.md
@@ -1,33 +1,30 @@
 # CW1 Spec: Proxy Contracts
 
-CW1 is a specification for proxy contracts based on CosmWasm.
-It is a very simple, but flexible interface designed for the case
-where one contract is meant to hold assets (or rights) on behalf of
-other contracts.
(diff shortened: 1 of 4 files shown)


ethanfrey (approved):
Good addition. Thanks

PR #849, opened 2022-12-05
Secure Admin controller

grod220:
In a recent security audit for Mars, we received the feedback we should use two-step admin transfers and that we should consider using the Admin package within cw-controllers. There were a few issues:
- The current struct does not have support for safe role transfers
- The .set() method moves Deps and makes it harder for users to continue using Deps below calling set()

Went ahead and rebuilt this from the ground up. Hoping to share it with the wider community. 

flowchart

⭐ New features ⭐ 
- I found the complexity of state to be best modeled via a finite state machine. Especially given the sensitivity of role changes, this prevents unexpected states will occur.
- First class support for two-step admin transfers: Proposed --> Accept
- Support for abolishing the role (burning the keys)
- Uses storage as a param instead of Deps so consumers don't need to sweat "moved" errors.
- A ton of tests

### DIFF
--- a/packages/controllers/src/secure_admin.rs
+++ b/packages/controllers/src/secure_admin.rs
@@ -70,10 +70,6 @@ impl AdminSetNoneProposed {
         })
     }
 
-    pub fn clear_admin(self) -> AdminState {
-        AdminState::A(AdminUninitialized)
-    }
-
     pub fn abolish_admin_role(self) -> AdminState {
         AdminState::D(AdminRoleAbolished)
     }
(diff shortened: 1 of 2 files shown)


ethanfrey:
Thanks for the addition and the great description.

We had similar request from dao dao but we decided to not change the existing contracts, and stabilize what was de facto 1.0 release.

I do think this would be a wonderful addition as a new controller. Not modifying existing packages, but adding one.

Call it SecureAdmin or another such name. Adding the controller should be one PR that doesn't affect any existing contract. After that, I would be open for another PR that adds one more contract to show this workflow, like a cw4-safe-group contract or such.

I am off this week, but if you make this a pure addition that doesn't change existing APIs, I would be happy to review it

grod220:
Think that makes sense. Especially given the sensitivities of admin role management, it's probably best to gradually integrate. Cool to hear that dao dao has similar needs, perhaps they can take a look too.

grod220:
Improvements:
- Removed ClearAdmin action. On second thought, it doesn't make sense and no admin should want to return to the uninitialized state where anyone can take the admin role. That state should only be accessed via the instantiate func. Also updated the flowchart.
- State change events now consume self. The reduces the ability to do multiple state changes on the same AdminState obj.
- Created helper from() for SecureAdminUpdate

packages/controllers/src/secure_admin.rs
```diff
@@ -0,0 +1,1091 @@
...
+
+type AdminResult<T> = Result<T, SecureAdminError>;
+
+/// The finite states that are possible
+#[cw_serde]
+enum AdminState {
+    A(AdminUninitialized),
+    B(AdminSetNoneProposed),
+    C(AdminSetWithProposed),
+    D(AdminRoleAbolished),
```
0xekez, line 42: any particular reason to give these single character names? imo, it makes the code a little hard to follow elsewhere because you've got to map back in your mind.

0xekez (commented):
thanks for putting this together. i've also thought for a while that moving this into a package would be good.

to be 100% honest, this feels a little too complex for me to feel comfortable using it in our (dao dao) contracts. the state machine architecture is cool, but i feel it may be overcomplexifying something which can be implemented pretty simply.

for reference, we do the whole admin nominate, withdraw, accept flow in about [seventy lines of code](https://github.com/DA0-DA0/dao-contracts/blob/74bd3881fdd86829e5e8b132b9952dd64f2d0737/contracts/dao-core/src/contract.rs#L201-L274) using a non-state machine, old-fashioned if statement approach. imo, in this case the simple approach is more readable which makes me more comfortable using it.

that said, i do like the api you have here for `SecureAdmin`. i wonder if we could keep that API, but use a simpler non-state-machine approach under the hood?

larry0x:
I'm working on a similar package called `cw-ownable`. The name comes from a [similar Ethereum package](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/Ownable.sol).

It will work this way (@0xekez's idea). We will define two proc-macros: `#[cw_ownable_execute]` and `#[cw_ownable_query]`:

```rust
// apply the `cw_ownable_execute` macro to an enum...
#[cw_ownable_execute]
#[cw_serde]
enum ExecuteMsg {
    Foo {},
    Bar {},
}

// ...transforms the enum to:
#[cw_serde]
enum ExecuteMsg {
    TransferOwnership {
        new_owner: String,
        /// An optional deadline for the new owner to accept ownership
        expiry: Option<cw_utils::Expiration>,
    },
    AcceptOwnership {},
    Foo {},
    Bar {},
}
```

```rust
// apply the `cw_ownable_query` macro to an enum...
#[cw_ownable_query]
#[cw_serde]
#[derive(QueryResponses)]
enum QueryMsg {
    #[returns(FizzResponse)]
    Fizz {},
    #[returns(BuzzResponse)]
    Buzz {},
}

// ...transforms the enum to:
#[cw_serde]
enum QueryMsg {
    #[returns(OwnerRespose)]
    Owner {},
    #[returns(FizzResponse)]
    Fizz {},
    #[returns(BuzzResponse)]
    Buzz {},
}

struct OwnerResponse {
    pub owner: String,
    pub pending_transfer: Option<OwnershipTransfer>,
}

struct OwnershipTransfer {
    pub new_owner: String,
    pub expiry: Option<Expiration>,
}
```

The package will also ship with the following functions, error type, and storage constants, which can be used to hook things up in `contract.rs`:

```rust
fn transfer_ownership(
    deps: DepsMut,
    info: MessageInfo,
    new_owner: String,
    expiry: Option<Expiration>,
) -> Result<Response, OwnableError> {
    // ...
}

fn accept_ownership(
    deps: DepsMut,
    env: Env,
    info: MessageInfo,
) -> Result<Response, OwnableError> {
    // ...
}

fn query_owner(deps: Deps) -> StdResult<OwnerResponse> {
    // ...
}
```

```rust
#[derive(Debug, thiserror::Error)]
enum OwnableError {
    #[error("sender is not the current contract owner")]
    NotOwner,

    #[error("sender is not the proposed new contract owner")]
    NotPendingOwner,

    #[error("proposed ownership transfer has expired")]
    OwnershipTransferExpired,
}
```

```rust
const OWNER: Item<Addr> = Item::new("owner");
const OWNERSHIP_TRANSFER: Item<OwnershipTransfer> = Item::new("ownership_transfer");
```

larry0x:
Also I think it's helpful to distinguish the use of the words "admin" vs "owner":

- "Admin" is the account who can migrate the contract
- "Owner" is the account who can invoke certain privileged execution functions

Often these two concepts are mixed up leading to confusions.

grod220:
Commit Improvements:
- Simplified SecureAdminExecUpdate and SecureAdminUpdate into one enum.
- Moved initialization logic out of the update function to force consumers to have to use this in instantiate.

packages/controllers/src/secure_admin.rs
grod220: I found AdminUninitialized(AdminUninitialized) to be a bit verbose when my IDE gives me the preview:
Screenshot 2022-12-07 at 8 57 58 AM


But perhaps that's better having it like that 🤷

grod220:
Hey folks, bumping this PR. Hopefully we can advance this further. Few things:

- Oak security (@christianvari) completed their audited of this package. No issues arose. 
- The naming is definitely one to think about. Re: "Admin", I agree we should change this so as to not confuse the role that can do migrations. Though, is “Owner” the right word? I almost think “Manager” or “Controller” or even “Governor” may be a better word. I kinda feel like we should use a term that best means “an entity with elevated privileges”. Doesn’t “Owner” imply something a bit different (or even more) than that?
- As an exercise, I implemented the same API but using a state struct (vs state machine) under the hood: [https://github.com/grod220/cw-plus/blob/admin-imperative-style/packages/controllers/src/secure_admin.rs](https://github.com/grod220/cw-plus/blob/admin-imperative-style/packages/controllers/src/secure_admin.rs). Surprisingly, it’s only 20 lines shorter in length. Given the safety reasons given above, I still prefer the original to this.
- We should add abolish & initialized state to the SecureAdminResponse query.
- @ethanfrey would you be able to review and weigh in here?

bitcoinfinancier (approved):

grod220:
Due to lack of interest, going to close this PR. [Mars-Owner](https://github.com/mars-protocol/owner) builds on this with additional features like an "emergency owner" role and simpler state machine design. Given it's such a universal need for cosmwasm contracts, if there is community interest, happy to re-open this and port that code over here.

PR #850, opened 2022-12-05
Standardised all derives to #[cw_serde]

nseguias:
Changed #[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema, Debug, Default)] for the new version #[cw_serde] in a few structs.


nseguias:
derive(Default)] needed, never mind!

PR #852, opened 2022-12-15
Remove unused schemars dependencies

webmaster128:
This is wrapped in cosmwasm-schema now such that we don't need it anymore in those packages.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -311,7 +311,6 @@ dependencies = [
  "cosmwasm-schema",
  "cosmwasm-std",
  "cw-storage-plus",
- "schemars",
  "serde",
 ]
 
@@ -386,7 +385,6 @@ dependencies = [
  "cw20",
  "cw20-base",
  "cw3",
- "schemars",
  "serde",
  "thiserror",
 ]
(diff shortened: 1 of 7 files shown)


ueco-jb (approved):

webmaster128:
oO macros are evil:

```
   Compiling cw2 v1.0.0 (/root/project/packages/cw2)
   Compiling cw-utils v0.16.0
error[E0433]: failed to resolve: use of undeclared crate or module `schemars`
  --> packages/cw2/src/lib.rs:26:1
   |
26 | #[cw_serde]
   | ^^^^^^^^^^^ use of undeclared crate or module `schemars`
   |
   = note: this error originates in the attribute macro `cw_serde` (in Nightly builds, run with -Z macro-backtrace for more info)
```

ueco-jb:
Ay, I stumbled upon this one once already.
Even `cargo machete` shows this dep as unused.
I'm starting to consider if macro is a good idea...

webmaster128:
Yeah, I think we can avoid that. If `cw_serde` does not assume other dependencies are installed and only uses stuff from cosmwasm_schema, we should be good. Then cosmwasm_schema could re-export the relevant symbols from schemars.

PR #853, opened 2022-12-16
Update dependencies

apollo-sturdy:
Updates the following dependencies:

cw-utils 0.16.0 -> 1.0.1
cw-storage-plus 0.16.0 -> 1.0.1
cw-multi-test 0.16.0 -> 0.16.1

I chose here to use the lowest (but >=1 where available) compatible versions of each so that we can run with `-Z minimal-versions`.

Please let me know if I should also bump the patch version of each package in this repo or if you have CI or something else that handles this. Also if I should bump the patch version of cw1 which was not modified in this PR.

### DIFF
--- a/contracts/cw1-subkeys/Cargo.toml
+++ b/contracts/cw1-subkeys/Cargo.toml
@@ -20,12 +20,12 @@ test-utils = []
 
 [dependencies]
 cosmwasm-schema = { version = "1.1.0" }
-cw-utils = "0.16.0"
+cw-utils = "1.0.1"
 cw1 = { path = "../../packages/cw1", version = "1.0.0" }
 cw2 = { path = "../../packages/cw2", version = "1.0.0" }
 cw1-whitelist = { path = "../cw1-whitelist", version = "1.0.0", features = ["library"] }
 cosmwasm-std = { version = "1.1.0", features = ["staking"] }
-cw-storage-plus = "0.16.0"
+cw-storage-plus = "1"
 schemars = "0.8.1"
 serde = { version = "1.0.103", default-features = false, features = ["derive"] }
 thiserror = "1.0.23"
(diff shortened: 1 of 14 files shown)


contracts/cw1-subkeys/Cargo.toml
```diff
@@ -20,12 +20,12 @@ test-utils = []
...
 [dependencies]
 cosmwasm-schema = { version = "1.1.0" }
-cw-utils = "0.16.0"
+cw-utils = "1.0.1"
 cw1 = { path = "../../packages/cw1", version = "1.0.0" }
 cw2 = { path = "../../packages/cw2", version = "1.0.0" }
 cw1-whitelist = { path = "../cw1-whitelist", version = "1.0.0", features = ["library"] }
 cosmwasm-std = { version = "1.1.0", features = ["staking"] }
-cw-storage-plus = "0.16.0"
+cw-storage-plus = "1"
```
webmaster128, line 28: Let's use 1.0.1 here as well since 1.0.0 is known to be broken for Wasm

webmaster128:
Thanks you for this PR. Good stuff.

contracts/cw1-subkeys/Cargo.toml
apollo-sturdy: Hmm strange. For some reason the first time I tried building with `-Z minimal-versions` it worked, but now that I tried again it does indeed fail because cw-storage-plus 1.0.0 is selected. I must have missed something the first time. Will change to 1.0.1. Could be a good idea to also yank 1.0.0.

webmaster128: When you use `-Z minimal-versions` you need to remove the lockfile fist. Otherwise the version from the lockfile is used.

> Could be a good idea to also yank 1.0.0.

This would not have helped here. A version that is in a lockfile already will still be used when yanked.

apollo-sturdy:
@webmaster128 Bumped cw-storage-plus 1.0.1. Please le me know if I should bump cw-plus package versions or if you will handle that.

webmaster128:
> Please let me know if I should bump cw-plus package versions or if you will handle that.

@ueco-jb will take care of release and CHANGELOG updates

Issue #830, opened 2022-10-17
Remove `cw1155` and `cw1155-base`


uint:
These are unmaintained.

AshutoshModanwal21:
Hi team, 
it was nice while working with CW packages, 
In my project requirement, there is need of this package & its contract, for developing NFT with fixed quantity of replicas, 
but as we see this package has been removed from the Cosmwasm, 
so please anyone suggest me any alternatives of this package to achieve the same functionalities was holded by cw115-base.

ueco-jb:
Hey @AshutoshModanwal21, I would recommend you checking out some of Stargaze repositories. It's a leading chain in cosmos that handles NFTs.
https://github.com/public-awesome/cw-nfts
https://github.com/public-awesome/cw-ics721

PR #856, opened 2023-01-07
Remove non-existent profile

mandrean:
`cw1155-base` was removed in https://github.com/CosmWasm/cw-plus/issues/830, but the release profile still exists. I caught the bug while running e2e tests in https://github.com/mandrean/cw-optimizoor.


ueco-jb (approved):

ueco-jb:
Thank you, good catch.

Issue #857, opened 2023-02-08
CW2: add a `cw2::VersionError` error type


larry0x:
A major use case of cw2 is when migrating a contract, check the stored version info to make sure we're not migrating a wrong contract, or migrating from a wrong version. Therefore, it makes sense to add a `VersionError` type which is to be emitted if this check fails:

```rust
#[derive(thiserror::Error, Debug)]
enum VersionError {
    #[error("contract version info not found")]
    NotFound,

    #[error("wrong contract: expecting `{expected}`, found `{found}`")]
    WrongContract {
        expected: String,
        found: String,
    },

    #[error("wrong contract version: expecting `{expected}`, found `{found}`")]
    WrongVersion {
        expected: String,
        found: String,
    },
}
```

```rust
// run this as part of the migration process
fn assert_version(
    storage: &dyn Storage,
    expected_contract: &str,
    expected_version: &str,
) -> Result<(), VersionError> {
    let Some(ContractVersion { contract, version }) = CONTRACT.may_load(storage)? else {
        return Err(VersionError::NotFound);
    };

    if contract != expected_contract {
        return Err(VersionError::WrongContract {
            expected: expected_contract.into(),
            found: contract,
        });
    }

    if version != expected_version {
        return Err(VersionError::WrongVersion {
            expected: expected_version.into(),
            found: version,
        });
    }

    Ok(())
}
```

PR #858, opened 2023-02-08
CW2: add a method to assert contract version

larry0x:
Closes: #857


larry0x:
I'd really like to use the let-else syntax, which imo is much nicer:

```diff
- let ContractVersion { contract, version } = match CONTRACT.may_load(storage)? {
-     Some(contract) => contract,
-     None => return Err(VersionError::NotFound),
- };
+ let Some(ContractVersion { contract, version }) = CONTRACT.may_load(storage)? else {
+     return Err(VersionError::NotFound);
+ };
```

Unfortunately it's only available after Rust 1.65, while CI is still on older versions.

packages/cw2/src/lib.rs
```diff
@@ -98,4 +146,44 @@ mod tests {
         };
         assert_eq!(expected, loaded);
     }
+
+    #[test]
+    fn assert_work() {
+        let mut store = MockStorage::new();
+
+        const EXPECTED_CONTRACT: &str = "crate:mars-red-bank";
```
ueco-jb, line 154: Shoo with that branding ser!

ueco-jb (approved):
LGTM
Thank you, a very nice addon

PR #859, opened 2023-03-03
Update workspace-optimizer version in README to 0.12.11

mikedotexe:
I came to build a cw1 contract and noticed we can update the `workspace-optimizer` version to the latest here:
https://hub.docker.com/r/cosmwasm/workspace-optimizer/tags

I've also run this locally myself and confirm it works as expected.


mikedotexe:
Looks as if CI is failing for one test with the result:

```
Caused by:
  package `base64ct v1.6.0` cannot be built because it requires rustc 1.60 or newer, while the currently active rustc version is 1.59.0
```

Seems to indicate a need to update the rustc version there

ueco-jb (approved):
LGTM, thank you.

PR #868, opened 2023-04-17
Added optional memo to ics20 packets

nicolaslara:
IBC has introduced an optional [memo field](https://github.com/cosmos/ibc/tree/main/spec/app/ics-020-fungible-token-transfer#using-the-memo-field) on the ICS20 transfer packet. The memo enables features such as Osmosis cross chain swaps, and forwarding tokens using the [packet forward middleware](https://github.com/strangelove-ventures/packet-forward-middleware). 

The change in this PR allows users to specify the memo to use when transferring CW20s. Note that since the memo is optional (both in the messages passed to the contract and the serialization of the Ics20Packet) this change will only affect the case when the memo is included in TransferMsg by the user. In short, this change should be backwards compatible.

### DIFF
--- a/contracts/cw20-ics20/src/ibc.rs
+++ b/contracts/cw20-ics20/src/ibc.rs
@@ -33,6 +33,9 @@ pub struct Ics20Packet {
     pub receiver: String,
     /// the sender address
     pub sender: String,
+    /// optional memo for the IBC transfer
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub memo: Option<String>,
 }
 
 impl Ics20Packet {
@@ -42,7 +45,12 @@ impl Ics20Packet {
             amount,
             sender: sender.to_string(),
             receiver: receiver.to_string(),
+            memo: None
         }
+    }
+
+    pub fn with_memo(self, memo: Option<String>) -> Self {
+        Ics20Packet { memo, ..self }
     }
 
     pub fn validate(&self) -> Result<(), ContractError> {
--- a/contracts/cw20-ics20/src/msg.rs
+++ b/contracts/cw20-ics20/src/msg.rs
@@ -51,6 +51,8 @@ pub struct TransferMsg {
     pub remote_address: String,
     /// How long the packet lives in seconds. If not specified, use default_timeout
     pub timeout: Option<u64>,
+    /// An optional memo to add to the IBC transfer
+    pub memo: Option<String>,
 }
 
 #[cw_serde]
--- a/contracts/cw20-ics20/src/contract.rs
+++ b/contracts/cw20-ics20/src/contract.rs
@@ -135,7 +135,7 @@ pub fn execute_transfer(
         amount.denom(),
         sender.as_ref(),
         &msg.remote_address,
-    );
+    ).with_memo(msg.memo);
     packet.validate()?;
 
     // Update the balance now (optimistically) like ibctransfer modules.


contracts/cw20-ics20/src/ibc.rs
```diff
@@ -33,6 +33,9 @@ pub struct Ics20Packet {
     pub receiver: String,
     /// the sender address
     pub sender: String,
+    /// optional memo for the IBC transfer
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub memo: Option<String>,
```
ueco-jb, line 38: This serde decorator is redundant. This field is `Some(_)` if filled or skipped if empty by default.

nicolaslara: That makes sense. I just wanted to be extra sure, but I can remove it.

nicolaslara: Actually, I don't think that's right. Removing it would make https://github.com/CosmWasm/cw-plus/pull/868/files#diff-0e169309202ffd9411af7c77249aeba5cec5e78912f3560e8ef8bb5c3ad565dbR700 fail.

ueco-jb: Oh, I think you're right. I was thinking about deserialization, while during serialization serde transforms null into some value - None in this case.

nicolaslara:
Tested this on mainnet as follows:

 * [Stored](https://www.mintscan.io/juno/txs/90A607FC00A85098D46E0425C6F83714DDF539D1487C239AFCDEB656B6DA6750) the compiled version of this contract
 * [Instantiated](https://www.mintscan.io/juno/txs/F7335F3E7EF20E3E54AF9AEE8B9EAADF14BAC7BC4BBD31EC8519BC68469CA640) it 
 * Created a channel between that contract and osmosis: `osmosis/transfer:channel-762 <> juno/wasm.juno1nnm4adhg6yz7a0s8ys0zwmm92reh66swqpfyp30m7dv3lawqtyuqh2utcu:channel-230`
 * Setup relayer for those channels
 * Sent some Neta from the Neta contract on Juno to Osmosis with a memo:
   * [Send tx](https://www.mintscan.io/juno/txs/E55CB7AF1EE717F2B7FB9F596D5C102008A551A9192AE4109F07165F3143177B)
   * [Receive tx](https://www.mintscan.io/osmosis/txs/702EE0564144A27AC16ED0257B73ED17665575A92AE863D5F1254AAAC941C1C3) (note the ICS20 memo "this is a test")
* Tried sending back to juno with a memo, which failed as expected since the receive side doesn't support the memo:
   * [Send tx](https://www.mintscan.io/osmosis/txs/2450D0F3228CD5569C6F5B08013A25E8C6C3F95C5C0425FD672E643F6E330EB1)
* Properly sent it back without the memo:
  * [Send tx](https://www.mintscan.io/osmosis/txs/62ABF85C9565D35B9554E14E8ECA8BDCA1FC6AEF44DDADCE7F8BF258A3D4F6AE) 
  * [Receive tx](https://www.mintscan.io/juno/txs/4964C7B1AEEC7DBD1E8AF69E6864A4114B1ED56718BE2A68BE6445859313487A)
  
 (If you notice the long time time between txs it's because it took me a while to discover allow lists on cw20_ics20 and copy pasting a CLI command made it so I kept sending to an osmosis addr instead of juno 😅)

ueco-jb (approved):
Okay, can you sign that license which is prompted on CI?

nicolaslara:
> Okay, can you sign that license which is prompted on CI?

I already did. Also clicked on the "recheck" link: 
Added_optional_memo_to_ics20_packets__868

nicolaslara:
Signed!

contracts/cw20-ics20/src/msg.rs
```diff
@@ -51,6 +51,8 @@ pub struct TransferMsg {
     pub remote_address: String,
     /// How long the packet lives in seconds. If not specified, use default_timeout
     pub timeout: Option<u64>,
+    /// An optional memo to add to the IBC transfer
+    pub memo: Option<String>,
```
webmaster128, line 55: This is a breaking change -.-

webmaster128:
Okay, so this breaks existing Rust code when constructing or destructing Ics20Packet or TransferMsg. Those are public symbols. However, I am not sure if this contract is embedded as a library somewhere.

webmaster128:
To be on the safe side, let's create a 1.x release branch befor this merge. Then we can have breaking changes on main that people can use. There will be a 2.0 release series of the contracts in summer anyways.

PR #880, opened 2023-09-01
Set the resolver version to value 2.

DariuszDepta:
Due to issue described here:

https://github.com/rust-lang/cargo/issues/9956

we have to set the resolver to version "2".


chipshort (approved):
LGTM

PR #888, opened 2023-11-23
Fix `Cargo.lock`

chipshort:
Had to update `cw-utils` again using `cargo update`, otherwise `cargo publish` was failing.


webmaster128 (approved):
This seems to be the circular dependency issue that was discussed. Here we now have two versions of cw2: the one maintained here and the dependency of cw-utils.

PR #892, opened 2024-01-11
Add cw orch

Kayanski:
This PR aims at adding cw-orch support to cw-plus.

This is NOT an enhancement to the contracts included in this repository, but rather an enhancement to the dev experience of using this repository.

I open this PR on behalf of [AbstractSDK](https://github.com/AbstractSDK/) to kick-start a discussion about inclusion of cw-orch into sample contracts and interfaces.

Including cw-orch doesn't interfere with on-chain contracts because every feature and inclusions are shielded behind the non-default interface feature.

Including cw-orch allows users to opt into using a powerful platform for tests/scripting/maintenance.

## Byproduct
Updated to cosmwasm-std 1.5.1


Kayanski:
We decided to integrate cw-plus directly inside cw-orch here: https://github.com/AbstractSDK/cw-orchestrator/tree/380bd236802499ac80a73c19f7f30e901082d755/packages/integrations/cw-plus

PR #894, opened 2024-02-05
chore: Move 'controllers' and 'cw2' to 'cw-minus'

kulikthebird:
Two packages are removed from this repository: **cw2** and **controllers**.  Their code is moved to cw-minus by this PR: https://github.com/CosmWasm/cw-minus/pull/26. 

Code coverage check was temporarily disabled for the purpose of moving the code. The PR itself does not modify the actual code coverage of any crate.


kulikthebird:
> Nice start. We should update the README, removing cw2 and cw-controllers there and adding a note about them being moved to cw-minus. Also, the scripts/publish.sh needs updating

Good point, I've just pushed some new commits that address that.

chipshort (approved):
LGTM

PR #895, opened 2024-02-05
docs: update token info response

etienne-napoleone:
The correct response for token info is `decimals`, not `decimal`.

See https://github.com/CosmWasm/cw-plus/blob/main/packages/cw20/src/query.rs#L59-L65


chipshort (approved):
Thanks (and sorry for the long wait)

PR #901, opened 2024-02-29
Update `cw1` to the v2 framework

uint:


uint:
It appears we can no longer split the update to v2 between cw1 and other packages due to dependency hell. Closing.

PR #905, opened 2024-03-27
Delete `publish.sh`

uint:
The `publish.sh` script is outdated again. Rather than maintain it, we should just use `cargo release` or something similar. I've updated the wiki for this repo accordingly.

`cargo release` automatically determines how to publish the entire workspace, taking into account the dependency graph between the packages. It's also more robust on failure to publish something.


DariuszDepta (approved):
LGTM
`cargo release` is promising, let's check how it works with this repo.
By default it works in **dry-run** mode, so it's safe to check before publishing.

uint:
> `cargo release` is promising, let's check how it works with this repo.

I used it for the last release and it worked beautifully!

PR #907, opened 2024-06-14
Upgrade optimizer and rename cargo config file

chipshort:
New Rust versions produce this warning:
> warning: `/code/contracts/cw4-stake/.cargo/config` is deprecated in favor of `config.toml`
note: if you need to support cargo 1.38 or earlier, you can symlink `config` to `config.toml`

`config.toml` is supported since cargo 1.39 (which is over 4 years old), so I think we can just rename them instead of symlinking.


uint (approved):
Sounds good!

PR #910, opened 2024-08-31
Replace serde/jsonschema derives with `cosmwasm_schema::cw_serde`

Buckram123:
This PR aims to reduce derive boilerplate with `cw_serde` macro. Originally I was missing `Debug` derive for cw20 base MigrateMsg and noticed that in this repo sometimes `cw_serde` macro used, sometimes `derive(Serialize, Deserialze...)`

### DIFF
--- a/packages/cw3/src/msg.rs
+++ b/packages/cw3/src/msg.rs
@@ -1,5 +1,4 @@
 use schemars::JsonSchema;
-use serde::{Deserialize, Serialize};
 use std::fmt;
 
 use cosmwasm_schema::cw_serde;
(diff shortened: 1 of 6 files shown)


packages/cw3/src/msg.rs
```diff
@@ -31,8 +30,8 @@ where
     },
 }
 
-#[derive(Serialize, Deserialize, Clone, Copy, PartialEq, Eq, JsonSchema, Debug)]
-#[serde(rename_all = "lowercase")]
+#[cw_serde]
+#[derive(Copy, Eq)]
```
Buckram123, line 34: cw_serde have `rename_all = "snake_case"`. It shouldn't have any difference from lowercase for those variants

PR #912, opened 2024-11-06
Install latest cosmwasm-check on CI

jawoznia:
I found out locally that `cosmwasm-check` fails on `cw1-whitelist` binary with:

```
target/wasm32-unknown-unknown/release/cw1_whitelist.wasm: failure
Error during static Wasm validation: Wasm bytecode could not be deserialized. Deserialization error: "reference-types not enabled: zero byte expected (at offset 0x139a)"
```

This PR is for consideration to detect such cases by using the latest version of `cosmwasm-check` and to showcase the issue.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -455,7 +455,7 @@ jobs:
       - run:
           name: Install cosmwasm-check
           # Uses --debug for compilation speed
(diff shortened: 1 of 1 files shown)


jawoznia:
The error is due to Rust `1.82` enabled reference types by default, while `cw-plus` uses an older version. Closing this issue.

